PR #51, opened 2024-05-25
Now users can submit forms by hitting enter

### LINKED ISSUE #47: Credential login form should submit when hitting the "enter" key
- form should submit when hitting the "enter" key
- the inputs smh are broken that I don't see the typical password suggestion by Chrome

career-tokens:
Fixes #47

### DIFF
--- a/packages/stack/src/components/credential-sign-in.tsx
+++ b/packages/stack/src/components/credential-sign-in.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import FormWarningText from "./form-warning";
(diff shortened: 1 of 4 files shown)


N2D4:
Since this will capture all Enter presses on the page, this will cause weird behaviour if there are other things on the same page as the SignIn component. I believe the better way to do this is to have an `<input type="submit" hidden>` inside of a `<form />` element, so we can use the default browser behavior. This should probably also be done on all the forms in stack-server.

@fomalhautb I believe you already built something like this? Can you combine it with this PR?

career-tokens:
Okay!

fomalhautb:
Moving to #57

PR #63, opened 2024-06-02
Image uploader base64

varunguleriaCodes:
Hi @fomalhautb , @N2D4  added a avatar upload functionality, added a size limit of 1mb.

Created a new field uploadedProfileImage which is of string type, I am storing the image as base64 string. Once user uploads a pic, he can see it but have to save it in order to change there avatar.

Issue -> https://github.com/stack-auth/stack/issues/48


N2D4:
Hey there, thanks for the contribution!

I'll take a closer look in a bit, but from glancing over it I have some comments:

- It's not enough to enforce the size limit on the client, we also need it on the server. We also need to convert the image to JPEG, because users may upload files in many different formats. Optimally, the server takes a data URL with length < 1,000,000, and converts it into a JPEG with some compression ratio, and if it's still less than 1MB in size, we store it in the DB. The `sharp` library is good for this task.

- If the user uploads an image > 500KB in size, instead of throwing an error, we should compress it on the client by converting it to JPEG with some compression ratio and scaling it down to 256 pixels on the larger dimension. Again, the `sharp` library can do this.

- Postgres' `text` type is limited in length, we probably need to use `bytea` to store the binary JPEG data, which maps to the Prisma `Bytes` type.

Do those make sense to you? Let me know what you think.

varunguleriaCodes:
Hey @N2D4 makes sense will implement these changes

varunguleriaCodes:
Hi @N2D4, @fomalhautb , a query regarding this, what I can understand:

Case 1. Image size < 500 kb send it as base 64 to backend convert this image to jpeg and save in database.
Case 2. Image size >= 500 kb compress it onto client with sharp as jpeg -> send as base 64 to backend then convert this image to jpeg and save in database

right ?

fomalhautb:
> Hi @N2D4, @fomalhautb , a query regarding this, what I can understand:
> 
> Case 1. Image size < 500 kb send it as base 64 to backend convert this image to jpeg and save in database. Case 2. Image size >= 500 kb compress it onto client with sharp as jpeg -> send as base 64 to backend then convert this image to jpeg and save in database
> 
> right ?

I guess it is better to:
- The client always convert all uploaded images (no matter size or format) into a valid jpeg (<500k, 256x256) using sharp
- The backend only accepts jpeg images that is valid and just throw an `KnownError` if it is not valid (the backend doesn't do any conversion or compression)

Vishvsalvi:
Is this still open? Would like to work

varunguleriaCodes:
Hey @Vishvsalvi still working on this, it is in progress

PR #79, opened 2024-06-16
Fixed issues related to magic link login functionality

RahulMishra0722:
1)if you look closely when i typed my email it was invisible i had to double click to highlight this happend because the background and text color were both black 

Screenshot 2024-06-15 at 8 10 46 PM
**After fix** 

Screenshot 2024-06-16 at 5 11 10 PM

2) @ Look Screenshot below the toggle with options Magic link and password were invisible same issue as before the text and the background color were the same white in the case 
Screenshot 2024-06-15 at 8 10 38 PM

**After fix ** 
 **In This Change When You Click On Any Of The Given Option It Becomes Coral and Magic Link is Set To Coral As Default**
Screenshot 2024-06-16 at 6 50 57 PM


RahulMishra0722:
i have signed this
Screenshot 2024-06-16 at 6 47 30 PM

PR #90, opened 2024-06-23
doc update for teams: remove user

roocell:
remove user from team had non-existent API call


roocell:
oh jeez - forgot to remove the removeUserFromTeam text

roocell:
forgot to remove the removeUserFromTeam text - redoing it

Issue #109, opened 2024-06-29
Projects should not be allowed to have same name 


wolfgunblood:
I can create new projects with same name 

- Its not ideal to have same name for different projects

- It can cause sorting and project management issues

![photo_2024-06-29_17-27-35](https://github.com/stack-auth/stack/assets/53962833/8cf8e6d6-773b-4d24-85be-8bb55af9ba5c)

PriyeshPandey2000:
@fomalhautb Would love to work on this issue .

P-yiush07:
@wolfgunblood heyy, can i work on this ?

fomalhautb:
The project name is just a display name, not an unique identifier. So it is fine to have two projects with the same name. We can add a warning when people creating a project with the same name though, but I think it should still be allowed.

P-yiush07:
> The project name is just a display name, not an unique identifier. So it is fine to have two projects with the same name. We can add a warning when people creating a project with the same name though, but I think it should still be allowed.

If it is so then can i implement this thing ?

N2D4:
@fomalhautb changed "Project Name" to "Project Display Name". I think this is enough to clarify intent — if it's still a problem, we can think about fixing this later.

PR #110, opened 2024-06-29
Projects should not be allowed to have same name

imran1khan:
project will always is going to have new name
issue #109


N2D4:
Closing this because it's no longer needed, see discussion in #109

PR #131, opened 2024-07-11
fix docs generation on windows

helios2003:
Fixes the error in the documentation generation for Windows.

Error message:

This was the error generated on running ``pnpm generate-docs``.
image

So, I fixed the path in in `apps/backend/scripts/generate-docs.ts`.


helios2003:
@N2D4,  @fomalhautb, can you PTAL at this?

fomalhautb:
I don't know if this works on windows, but this will break MacOS and Linux environments. which means it will also break production. Please make sure that it works on all the platforms

PR #170, opened 2024-08-04
Fix: Improve error handling for Server API

### LINKED ISSUE #141: Improve Error Handling in API Handlers
Currently, there is no error handling in the API handlers. When something goes wrong, the server returns an **Internal Server Error** without providing meaningful feedback to the user.

### Few Examples

1. **Adding an Existing User to a Team**

    -  Action: Attempt to add a user who is already a member of a team
    -  Result: Unique constraint fails
    -  Screenshot:
Screenshot from 2024-07-21 18-54-33

-  **Deleting a Non-Existent User from a Team**

    -  Action: Attempt to delete a user who is not in the team
    -  Result: Record does not exist
    -  Screenshot:
image

- **Getting User Teams with an Invalid User ID**

    -  Action: Attempt to fetch teams for an invalid user ID
    -  Result: Invalid uuid format
    -  Screenshot:
image

### Expected Results
The API should handle errors gracefully and return meaningful, user-friendly error messages, such as:
- For unique constraint violations: "User already exists in the team."
- For non-existent user deletions: "User not found in the team."
- For invalid user ID requests: "Invalid user ID."

### Additional Notes
These are just a few examples of the endpoints affected. There are many more endpoints that currently lack proper error handling.

**I would love to work on this issue. Please assign it to me if possible.**

kfahad5607:
**Fixes #141 by checking for entities before deleting/modifying them to avoid prisma query error.**

#### Route handler changes
- Checking if team membership exists in **Remove a user from a team**
- Checking if team and user exist in **Grant a team permission to a user**
- Checking if team-permission for user exists in **Revoke a team permission from a user**
- Checking if team exists in **Delete a team**
- Checking if user exists in **Update user**
- Checking if team membership exists when `selected_team_id` is passed in request body in **Update user**
- Checking if user exists in **Delete user**

#### Other changes
- Updated `selectedTeamIdSchema` to validate string as `uuid`
- Added new helper function `ensureUserTeamPermissionExist` in `request-checks` to use in **Revoke a team permission from a user**
- Added new error `UserTeamPermissionNotFound` to use in `ensureUserTeamPermissionExist`


apps/backend/src/lib/request-checks.tsx
```diff
@@ -105,6 +105,67 @@ export async function ensureUserHasTeamPermission(
   }
 }
 
+export async function ensureUserTeamPermissionExist(
```
fomalhautb: I think this function is a little bit duplicated to `ensureUserHasTeamPermission`. Maybe you can combine them.

apps/backend/src/app/api/v1/users/crud.tsx
```diff
@@ -158,6 +167,11 @@ export const usersCrudHandlers = createLazyProxy(() => createCrudHandlers(usersC
         }
       }
 
+      // If the selected_team_id is present and we reach here that means user does exist. Hence, we have this check.
+      if (!data.selected_team_id) {
+        await ensureUserExist(tx, { projectId: auth.project.id, userId: params.user_id });
```
fomalhautb: We can move this check to the beginning and remove the condition so it is easier to follow

fomalhautb (changes requested):
Looks pretty clean! Also can you sign the CLA?

apps/backend/src/lib/request-checks.tsx
kfahad5607: I am combining these functions, but logically, they signify different types of errors.

1. `ensureUserHasTeamPermission`
   - This function throws a `TeamPermissionRequired` error, which indicates an authorization error. Users might misconstrue this error as a permission issue in the case of the **Revoke a team permission from a user** API request.

2. `ensureUserTeamPermissionExist`
   - This function throws a `UserTeamPermissionNotFound` error, indicating that the permission the user is trying to delete does not exist. This is more appropriate for the **Revoke a team permission from a user** API request.

kfahad5607:
I have made the requested changes.

- Deleted the `ensureUserTeamPermissionExist` function and `UserTeamPermissionNotFound` error.
- Updated the `ensureUserHasTeamPermission` function to accept both **System Permission** and **Custom Permission** via `options.permissionId`.
- Checking for an existing membership between the team and user, instead of performing individual checks for the user and team in **Grant a team permission to a user**.

PS: I have also signed the CLA.

N2D4:
> [CLA assistant check](https://cla-assistant.io/stack-auth/stack?pullRequest=170)

FYI, looks like the bot hasn't picked up the CLA, but I can confirm that you signed it, so all good!

apps/backend/src/lib/request-checks.tsx
fomalhautb: yeah, I added `errorType` parameter to decide which type of error to throw in the `ensureUserHasTeamPermission` function.

fomalhautb:
Thanks for the contribution!

PR #187, opened 2024-08-15
Implement Discord oauth provider

NunoD87:
This PR adds Discord as a possible oauth provider for Stack, as discussed [here](https://discord.com/channels/1215503771199864912/1215504045410754620/1273292256597639220) on [Stack's Discord Server](https://discord.stack-auth.com/).

All the Discord information for this implementation was gathered from [Discord's Official Documentation](https://discord.com/developers/docs/intro):
- [API Reference](https://discord.com/developers/docs/reference)
- [OAuth2](https://discord.com/developers/docs/topics/oauth2)
- [Users API Reference](https://discord.com/developers/docs/resources/user)
- [User Object](https://discord.com/developers/docs/resources/user#user-object)

Brading information (SVG + background color) was also gathered from an official source [here](https://discord.com/branding).

Documentation was updated to contain a setup guide link to obtaining the OAuth credentials, following the previously added providers in the [`Going to Production`](https://docs.stack-auth.com/getting-started/production) section.

This implementation follows the same implementation for Spotify's oauth provider, and it should be very straightforward to get this running.

PR #200, opened 2024-08-22
Add Gitlab OAuth Provider Integration

Iamsidar07:
This PR introduces the integration of Gitlab as an OAuth Provider.


fomalhautb:
There are some more things you need to do to add a new provider, for example the frontend buttons. Also notice that GitLab should not be a new shared/proxied provider, but only as a standard provider. Here is an example of how to add discord

Iamsidar07:
> There are some more things you need to do to add a new provider, for example the frontend buttons. Also notice that GitLab should not be a new shared/proxied provider, but only as a standard provider. Here is an example of how to add discord. ALso please make sure all the typechecks, lints, and tests pass

Can you share an example of how to add discord?

fomalhautb:
> > There are some more things you need to do to add a new provider, for example the frontend buttons. Also notice that GitLab should not be a new shared/proxied provider, but only as a standard provider. Here is an example of how to add discord. ALso please make sure all the typechecks, lints, and tests pass
> 
> Can you share an example of how to add discord? Do I have to add oauth in both backend and dashboard?

Sorry, forgot to link it, https://github.com/stack-auth/stack/pull/187

Iamsidar07:
> > > There are some more things you need to do to add a new provider, for example the frontend buttons. Also notice that GitLab should not be a new shared/proxied provider, but only as a standard provider. Here is an example of how to add discord. ALso please make sure all the typechecks, lints, and tests pass
> > 
> > 
> > Can you share an example of how to add discord? Do I have to add oauth in both backend and dashboard?
> 
> Sorry, forgot to link it, #187

Implemented in new PR #201, Please can you check it now? And thanks!

PR #203, opened 2024-08-23
Fix incorrect redirect on account settings when using a custom handler

### LINKED ISSUE #202: Incorrect redirect on account settings when using a custom handler
I'm using StackAuth with a custom handler (`urls: { handler: "/auth" }`) but when I visit `/auth/account-settings`, after a few seconds I am redirected to `/handler/account-settings` which results in an error.

I believe the bug is on this line of code:

https://github.com/stack-auth/stack/blob/25fc625f7d53721dd5ecd7d8cb1918cb852602e1/packages/stack/src/components-page/account-settings.tsx#L90

tanguyMichardiere:
Fixes #202


fomalhautb:
Looks good, thanks!

PR #205, opened 2024-08-25
Add create new project btn under project switcher

Iamsidar07:
I think placing the "Create New Project" button below the project switcher is a good idea.
image


fomalhautb:
Sounds good, feel free to work on it if you are interested

fomalhautb:
Looks great, thanks!

PR #220, opened 2024-09-02
feat: Enhance API Key Visibility and Offer Download Option

Iamsidar07:
[Screencast from 2024-09-02 10-10-52.webm](https://github.com/user-attachments/assets/972fffe3-81a1-4251-86ab-58a04328a26c)


fomalhautb:
Can you maybe put the button to the `EnvKeys` component? There are two places (onboarding and api keys) where it is used and we probably want the button to be in both places

Iamsidar07:
> Can you maybe put the button to the `EnvKeys` component? There are two places (onboarding and api keys) where it is used and we probably want the button to be in both places

Moved to env-keys component.

fomalhautb:
thanks!

PR #259, opened 2024-09-19
fixed verification code, added tests

fomalhautb:


apps/e2e/tests/backend/endpoints/api/v1/contact-channels/verify.test.ts
```diff
@@ -59,3 +59,59 @@ it("each verification code that was already requested can be used exactly once",
     `);
   }
 });
+
+it("should not allow verify a code that doesn't exist", async ({ expect }) => {
```
N2D4, line 63:
```suggestion
it("should not allow verifying a code that doesn't exist", async ({ expect }) => {
```

N2D4 (approved):
merge when fixed

PR #260, opened 2024-09-20
Added handling for user canceling the oauth process

fomalhautb:


packages/stack/src/components-page/error-page.tsx
```diff
@@ -18,13 +18,16 @@ export function ErrorPage(props: { fullPage?: boolean, searchParams: Record<stri
 
   const unknownErrorCard = <PredefinedMessageCard type='unknownError' fullPage={!!props.fullPage} />;
 
-  if (!errorCode || !message || !details) {
+  console.log(errorCode, message, details);
```
N2D4:
```suggestion
```

packages/stack/src/components-page/error-page.tsx
```diff
@@ -61,5 +64,22 @@ export function ErrorPage(props: { fullPage?: boolean, searchParams: Record<stri
...
 
+  if (error instanceof KnownErrors.OAuthProviderAccessDenied) {
+    return (
+      <MessageCard
+        title="OAuth provider access denied"
+        fullPage={!!props.fullPage}
+        primaryButtonText="Sign in again"
+        primaryAction={() => stackApp.redirectToSignIn()}
+        secondaryButtonText="Go to Home"
+        secondaryAction={() => stackApp.redirectToHome()}
```
N2D4: translate this

packages/stack/src/components-page/error-page.tsx
```diff
@@ -61,5 +64,22 @@ export function ErrorPage(props: { fullPage?: boolean, searchParams: Record<stri
...
+      >
+        <Typography>
+          {t("Did you maybe pressed cancel on the OAuth provider's page?")}
```
N2D4: grammar
```suggestion
          {t("The sign-in operation has been cancelled. Please try again. [access_denied]")}
```

packages/stack/src/components-page/error-page.tsx
```diff
@@ -61,5 +64,22 @@ export function ErrorPage(props: { fullPage?: boolean, searchParams: Record<stri
     );
   }
 
+  if (error instanceof KnownErrors.OAuthProviderAccessDenied) {
```
N2D4, line 65: I think instead of rendering this card, we should just redirect back to sign-in or sign-up (wherever we came from)

fomalhautb: Not a big fan of that UX, I think the user should still see that it failed

PR #262, opened 2024-09-22
fix overflow issue with long team names

Iamsidar07:
This PR fixes an issue where long team names would overflow their container in the UI. The changes ensure that team names are properly truncated.

mobile
desktop


fomalhautb:
thx

PR #263, opened 2024-09-23
OTP

fomalhautb:


packages/stack/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@stackframe/stack",
-  "version": "2.5.37",
+  "version": "2.5.38",
```
N2D4: shouldn't we undo this until we bump the package versions?

N2D4 (approved):
just one comment, otherwise looks good

packages/stack/package.json
fomalhautb: Sure, will also add tests for the endpoints that checks the client version

PR #272, opened 2024-09-25
Add noRepeatPassword props

### LINKED ISSUE #267: Make password repeat on sign up configurable on the component
image

Create a prop option "noPasswordRepeat" on the component so that it hides the second "password repeat" field

Iamsidar07:
fixes #267 
Screenshot from 2024-09-25 08-49-21
Screenshot from 2024-09-25 08-49-31


packages/stack/src/components/credential-sign-up.tsx
```diff
@@ -12,7 +12,7 @@ import { useStackApp } from "..";
 import { useTranslation } from "../lib/translations";
 import { FormWarningText } from "./elements/form-warning";
 
-export function CredentialSignUp() {
+export function CredentialSignUp(props: { noPasswordRepeat?: boolean }) {
   const { t } = useTranslation();
 
   const schema = yupObject({
```
fomalhautb, line 18: Did you test if this signs in? I think the schema checks is not modified so it should not pass the "password must match" validation

fomalhautb (changes requested):

PR #282, opened 2024-09-27
feat: show error message when no auth method enabled

### LINKED ISSUE #269: Show an error message in SignIn component when no login methods are available
Currently it is just empty when you disable all the auth methods.

Iamsidar07:
fixes #269 
image


fomalhautb:
This only checks magic link and password email, but doesn't check if oauth is also empty

fomalhautb:
looks good, thx!

PR #287, opened 2024-09-30
New contact channels

fomalhautb:
This is a redesign of the contact channels/auth:
- Each contact channel has `isVerified`, `usedForAuth`, `isPrimary` attributes
- You can login with any contact channel that is `usedForAuth`
- otp/password auth methods now are not bound with identifier/contact channel


apps/e2e/tests/backend/endpoints/api/v1/auth/password/sign-up.test.ts
```diff
@@ -27,20 +27,27 @@ it("should sign up new users", async ({ expect }) => {
...
+        "profile_image_url": null,
+        "requires_totp_mfa": false,
+        "selected_team": null,
+        "selected_team_id": null,
+        "signed_up_at_millis": <stripped field 'signed_up_at_millis'>,
       },
-    ]
+      "headers": Headers { <some fields may have been hidden> },
+    }
   `);
```
N2D4, line 51: this should still test the auth methods

packages/stack-shared/src/interface/crud/users.ts
```diff
@@ -27,6 +36,18 @@ export const usersCrudServerReadSchema = fieldSchema.yupObject({
   profile_image_url: fieldSchema.profileImageUrlSchema.nullable().defined(),
   signed_up_at_millis: fieldSchema.signedUpAtMillisSchema.required(),
   has_password: fieldSchema.yupBoolean().required().meta({ openapiField: { description: 'Whether the user has a password associated with their account', exampleValue: true } }),
+  client_metadata: fieldSchema.userClientMetadataSchema,
+  client_read_only_metadata: fieldSchema.userClientReadOnlyMetadataSchema,
+  server_metadata: fieldSchema.userServerMetadataSchema,
+  last_active_at_millis: fieldSchema.userLastActiveAtMillisSchema.required(),
+  contact_channels: fieldSchema.yupArray(contactChannelSchema).required(),
```
N2D4: do we want this on here? I'm kinda split, if we have this, why don't we have auth_methods? contact_channels also has its own endpoint, so there's probably a world where we don't wanna duplicate the information on the users endpoint

(GraphQL would answer this for us)

packages/stack-shared/src/interface/crud/users.ts
```diff
@@ -35,41 +56,6 @@ export const usersCrudServerReadSchema = fieldSchema.yupObject({
...
   requires_totp_mfa: fieldSchema.yupBoolean().required().meta({ openapiField: { hidden: true, description: 'Whether the user is required to use TOTP MFA to sign in', exampleValue: false } }),
-  /**
-   * @deprecated
-   */
-  oauth_providers: fieldSchema.yupArray(fieldSchema.yupObject({
-    id: fieldSchema.yupString().required(),
-    account_id: fieldSchema.yupString().required(),
-    email: fieldSchema.yupString().nullable(),
-  }).required()).required().meta({ openapiField: { hidden: true, description: 'A list of OAuth providers connected to this account', exampleValue: [{ id: 'google', account_id: '12345', email: 'john.doe@gmail.com' }] } }),
-  auth_methods: fieldSchema.yupArray(fieldSchema.yupUnion(
```
N2D4, line 46: was there no place where we used auth_methods and connected_accounts in the frontend?

fomalhautb: I don't think they are used, I can double check

PR #296, opened 2024-10-10
Auto detect https for cookies

fomalhautb:

### DIFF
--- a/packages/stack/src/lib/cookie.ts
+++ b/packages/stack/src/lib/cookie.ts
@@ -4,31 +4,14 @@ import { calculatePKCECodeChallenge, generateRandomCodeVerifier, generateRandomS
 
 type SetCookieOptions = { maxAge?: number };
 
-function isRscCookieUnavailableError(e: any) {
-  const allowedMessageSnippets = ["was called outside a request scope", "cookies() expects to have requestAsyncStorage"];
-  return typeof e?.message === "string" && allowedMessageSnippets.some(msg => e.message.includes(msg));
-}
-
-function isRscHeadersUnavailableError(e: any) {
-  const allowedMessageSnippets = ["was called outside a request scope"];
-  return typeof e?.message === "string" && allowedMessageSnippets.some(msg => e.message.includes(msg));
-}
-
 export function getCookie(name: string): string | null {
-  try {
-    return rscCookies().get(name)?.value ?? null;
-  } catch (e: any) {
-    if (isRscCookieUnavailableError(e)) {
-      return Cookies.get(name) ?? null;
-    } else {
-      throw e;
-    }
-  } finally {
+  if (typeof window !== "undefined") {
     // This is a flag to automatically detect whether we're on https for the next server request
     // Check out the comment in setCookie for more details
-    if (typeof window !== "undefined" && window.location.protocol === "https:") {
-      Cookies.set("stack-is-https", "true");
-    }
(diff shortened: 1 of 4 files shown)


packages/stack/src/lib/cookie.ts
```diff
@@ -42,19 +48,34 @@ export function deleteCookie(name: string) {
...
+  if (typeof window !== "undefined" && window.location.protocol === "https:") {
+    isSecureCookie = true;
+  }
+  try {
+    const proto = rscHeaders().get("x-forwarded-proto");
+    if (proto === "https") {
+      isSecureCookie = true;
+    }
+  } catch (e: any) {
+    // Ignore
```
N2D4: never catch everything

packages/stack/src/lib/cookie.ts
```diff
@@ -1,23 +1,19 @@
...
-  } catch (e: any) {
-    if (isRscCookieUnavailableError(e)) {
-      return Cookies.get(name) ?? null;
-    } else {
-      throw e;
+  if (typeof window !== "undefined") {
+    // This is a flag to automatically detect whether we're on https for the next server request
+    // Check out the comment in setCookie for more details
+    if (window.location.protocol === "https:") {
+      Cookies.set("stack-is-https", "true");
```
N2D4: the `Secure` flag should be set

packages/stack/src/lib/cookie.ts
```diff
@@ -1,23 +1,19 @@
...
-  try {
-    return rscCookies().get(name)?.value ?? null;
-  } catch (e: any) {
-    if (isRscCookieUnavailableError(e)) {
-      return Cookies.get(name) ?? null;
-    } else {
-      throw e;
+  if (typeof window !== "undefined") {
+    // This is a flag to automatically detect whether we're on https for the next server request
+    // Check out the comment in setCookie for more details
```
N2D4:
```suggestion
    // If we are on https, set a helper cookie that lets the server know
    // see comment in `setCookie` below
```

packages/stack/src/lib/cookie.ts
```diff
@@ -30,37 +26,38 @@ export function setOrDeleteCookie(name: string, value: string | null, options: S
...
   }
 }
 
 export function setCookie(name: string, value: string, options: SetCookieOptions = {}) {
-  const isProd = process.env.NODE_ENV === "production";
-  try {
+  // Check if the current page is server over HTTPS with:
+  // 1. Check if the https cookie is set on the client (see comment above)
+  // 2. If on the client, check the protocol directly
+  // 3. If on the server, check the X-Forwarded-Proto header
```
N2D4:
```suggestion
  // Whenever the client is on HTTPS, we want to set the Secure flag on the cookie.
  //
  // This is not easy to find out on a Next.js server, so we use the following steps:
  //
  // 1. If we're on the client, we can check window.location.protocol which is the ground-truth
  // 2. Check whether the stack-is-https cookie exists. This cookie is set in various places on
  //      the client if the protocol is known to be HTTPS
  // 3. Check the X-Forwarded-Proto header
  // 4. Otherwise, assume HTTP without the S
  //
  // Note that malicious clients could theoretically manipulate the `stack-is-https` cookie or
  // the `X-Forwarded-Proto` header; that wouldn't cause any trouble except for themselves,
  // though.
```

N2D4 (approved):

PR #300, opened 2024-10-15
SDK docs

fomalhautb:


fomalhautb:
@patchbot ResolveComments

The correct way to use stack handler is like this, update the `docs/fern/docs/pages/sdk/stack-handler.mdx` file
```
import { StackHandler } from "@stackframe/stack";
import { stackServerApp } from "@/stack";

export default function Handler(props: any) {
  return <StackHandler fullPage app={stackServerApp} {...props} />;
}
```

fomalhautb:
@patchbot ResolveComments

The correct way to use the stack handler is like this, update the `docs/fern/docs/pages/sdk/stack-handler.mdx` file
```
import { StackHandler } from "@stackframe/stack";
import { stackServerApp } from "@/stack";

export default function Page(props: any) {
  return <StackHandler app={stackServerApp} {...props} /* filll in other parameters here */ />;
}
```

PR #309, opened 2024-10-21
OTP auth on the client SDK and dashboard

fomalhautb:
- Fixed the bug that user creation doesn't check the otp/password configs correctly
- UX improvement: added otp auth to user creation on the dashboard
- Improved user creation docs
- Added otp auth to user creation SDK


packages/stack-shared/src/schema-fields.ts
```diff
@@ -267,7 +268,11 @@ export const userOAuthProviderSchema = yupObject({
   provider_user_id: yupString().required(),
 });
 export const userLastActiveAtMillisSchema = yupNumber().nullable().meta({ openapiField: { description: _lastActiveAtMillisDescription, exampleValue: 1630000000000 } });
-export const userOtpAuthEnabledSchema = yupBoolean().meta({ openapiField: { hidden: true, description: 'Whether the user has OTP/magic link enabled', exampleValue: true } });
+export const userOtpAuthEnabledSchema = yupBoolean().meta({ openapiField: { hidden: true, description: 'Whether the user has OTP/magic link enabled. ', exampleValue: true } });
+export const userOtpAuthEnabledMutationSchema = yupBoolean().meta({ openapiField: { hidden: true, description: 'Whether the user has OTP/magic link enabled. Note that only verified email can be used for sign-in with OTP. If OTP auth is not enabled for the project, it will throw an error.', exampleValue: true } });
```
N2D4:
```suggestion
export const userOtpAuthEnabledMutationSchema = yupBoolean().meta({ openapiField: { hidden: true, description: 'Whether the user has OTP/magic link enabled. Note that only accounts with verified emails can sign-in with OTP.', exampleValue: true } });
```
make sure to run docstrings through chatgpt/grammarly/whatever — maybe there's a way to automate this?

Also, "it will throw an error" doesn't make sense for a boolean attribute

packages/stack-shared/src/schema-fields.ts
```diff
@@ -267,7 +268,11 @@ export const userOAuthProviderSchema = yupObject({
...
+export const userHasPasswordSchema = yupBoolean().meta({ openapiField: { hidden: true, description: 'Whether the user has a password set. If the user does not have a password set, it will not be able to sign in with email/password.', exampleValue: true } });
```
N2D4:
```suggestion
export const userHasPasswordSchema = yupBoolean().meta({ openapiField: { hidden: true, description: 'Whether the user has a password set. If the user does not have a password set, they will not be able to sign in with email/password.', exampleValue: true } });
```

packages/stack-shared/src/schema-fields.ts
```diff
@@ -267,7 +268,11 @@ export const userOAuthProviderSchema = yupObject({
...
+export const userPasswordSchema = yupString().meta({ openapiField: { description: 'A new password for the user, overwriting the old one (if it exists). Specifying this option revokes all current sessions.', exampleValue: 'my-new-password' } });
```
N2D4:
```suggestion
export const userPasswordMutationSchema = yupString().meta({ openapiField: { description: 'Sets the user's password. Doing so revokes all current sessions.', exampleValue: 'my-new-password' } });
```

We should make it clear that this is for updating the password and not user passwords in general. Also, shouldn't this be nullable or so? (To disable password auth)?

N2D4: Shouldn't this be nullable or so? (To disable password auth)

packages/stack-shared/src/schema-fields.ts
```diff
@@ -267,7 +268,11 @@ export const userOAuthProviderSchema = yupObject({
...
+export const userTotpSecretSchema = base64Schema.meta({ openapiField: { description: 'A TOTP secret for the user, overwriting the old one (if it exists). Set to null to disable 2FA.', exampleValue: 'dG90cC1zZWNyZXQ=' } });
```
N2D4:
```suggestion
export const userTotpSecretMutationSchema = base64Schema.meta({ openapiField: { description: 'Enables 2FA and sets a TOTP secret for the user. Set to null to disable 2FA.', exampleValue: 'dG90cC1zZWNyZXQ=' } });
```

docs/fern/docs/pages/sdk/stack-app.mdx
```diff
@@ -379,39 +379,39 @@ Note that although the `useXYZ` hooks are available on `StackServerApp`, they ar
...
+    </ParamField>
+    <ParamField path="forgotPassword" type="string" default="urls.handler + '/forgot-password'">
+    The URL of the forgot password page.
+    </ParamField>
+    <ParamField path="accountSettings" type="string" default="urls.handler + '/account-settings'">
+    The URL of the account settings page.
+    </ParamField>
+    <ParamField path="handler" type="string" default="'/handler'">
+    The URL of the handler root. When changing this, you should also change the location of the `StackHandler` component and the domain handler path on the dashboard. Currently putting the handler at the root `/` is not supported.
+    </ParamField>
```
N2D4: ?

N2D4: this is extra funny because the containing div's class is `indented` haha

N2D4 (approved):
merge after changes

packages/stack-shared/src/schema-fields.ts
fomalhautb: nullable happens in the place where this is used

docs/fern/docs/pages/sdk/stack-app.mdx
fomalhautb: This is how other parts are formatted. We can change it, but then we should change all the places this occurs

N2D4: Is it hard to do that?

fomalhautb: Sure, I updated it. but it doesn't really belong to this PR

PR #313, opened 2024-10-22
migration from next-auth to stack guide

newtonfav:


fomalhautb:
This doesn't seem to contain any next-auth specific content

lacymorrow (changes requested):
Fix conflicts

PR #326, opened 2024-10-30
Fix updatePassword example

RishikesavanRamesh:
Fixes the `updatePassword` example by correctly passing the required arguments.


fomalhautb:
thx

PR #331, opened 2024-11-07
Added callbackUrl to server user invitation

fomalhautb:


docs/fern/docs/pages/sdk/team.mdx
```diff
@@ -280,6 +280,41 @@ Add a user to the team directly without sending an invitation email.
 await team.addUser('user_id_123');
 ```
 
+### `inviteUser()`
+
+This is the same as the `inviteUser` method on the `Team` object, except that it requires a `callbackUrl` parameter.
```
N2D4: this doesn't have to do with Team/ServerTeam; those concepts are different from whether the code runs on the browser or on a physical server. `ServerTeam` should be strictly more powerful than `Team`

PR #334, opened 2024-11-08
Dashboard UI for adding users to a team

fomalhautb:
image

### DIFF
--- a/apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/teams/[teamId]/page-client.tsx
+++ b/apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/teams/[teamId]/page-client.tsx
@@ -2,11 +2,10 @@
 import { TeamMemberSearchTable } from '@/components/data-table/team-member-search-table';
 import { TeamMemberTable } from '@/components/data-table/team-member-table';
 import { ServerTeam } from '@stackframe/stack';
-import { ActionDialog, Button, useToast } from '@stackframe/stack-ui';
+import { ActionDialog, Button } from '@stackframe/stack-ui';
 import { notFound } from 'next/navigation';
 import { PageLayout } from '../../page-layout';
 import { useAdminApp } from '../../use-admin-app';
-import { KnownErrors } from '@stackframe/stack-shared';
 
 
 export function AddUserDialog(props: {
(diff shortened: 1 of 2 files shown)


apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/teams/[teamId]/page-client.tsx
```diff
@@ -1,8 +1,46 @@
...
+    <TeamMemberSearchTable
+      action={(user) => <div className='flex justify-end w-[60px]'><Button
+        variant="outline"
+        onClick={async () => {
+          try {
+            await props.team.addUser(user.id);
+            toast({ title: 'User added to team', variant: 'success' });
+          } catch (error) {
+            if (error instanceof KnownErrors.TeamMembershipAlreadyExists) {
+              toast({ title: 'User already a member of this team', variant: 'destructive' });
```
Comment (N2D4) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/teams/[teamId]/page-client.tsx]: IMO toasts are for things that are not crucial to the user flow (eg. "everything went as planned") — this should be an alert

N2D4 (approved):

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/teams/[teamId]/page-client.tsx
Comment (fomalhautb) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/teams/[teamId]/page-client.tsx]: Actually, I think it is better to show an "added" button (see the updated image in the PR description). It will be a bit more annoying when we later paginate the team-member tables as we will need to do it on the server side, but we can worry about that later

Comment (N2D4) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/teams/[teamId]/page-client.tsx]: What happens if the user was not added successfully because they're already in the team?

Comment (N2D4) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/teams/[teamId]/page-client.tsx]: Ah nvm I get what you mean now

Comment (N2D4) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/teams/[teamId]/page-client.tsx]: SGTM

Comment (N2D4) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/teams/[teamId]/page-client.tsx]: The toast is not necessary though — from a user perspective, "the user has been added" and "the user has already been on the team" is equivalent

PR #338, opened 2024-11-09
Update SDK and component reference layout

N2D4:
# https://stack-auth-preview-31e9a254-992a-41ec-801b-47f93cf45312.docs.buildwithfern.com/sdk

<hr>

Screenshot 2024-11-09 at 20 04 39

<hr>

Screenshot 2024-11-10 at 16 16 54

<hr>

Screenshot 2024-11-10 at 15 32 59


fomalhautb (changes requested):
This is great! A few things to improve:
- Why do we distinguish between "Object" and "Types"? They are not really that different and I think it only causes confusion.
- Can we change the "T" icon color to purple or something? Red feels like warning.
- There is no navigation on the page anymore, is that intended? It is quite useful for this kind of long references

N2D4:
- Stack apps are more like classes than types; their shapes are different, and also I want them to stick to the top of the documentation

- Yes we can change it (I got the colors from Deno)

- Yes, there's not enough space for another navigation sidebar unless you have an ultrawide screen

fomalhautb (approved):

PR #353, opened 2024-11-28
Self-host docker

fomalhautb:
based on #280


.dockerignore
N2D4, line 1: I think this should also contain all of `.gitignore`, right? eg. .DS_Store, .untracked, .vercel, etc

.github/workflows/docker-build.yaml
```diff
@@ -0,0 +1,48 @@
+name: Docker Build and Push
+
+on:
+  push:
+    branches:
+      - main
+      - dev
```
N2D4, line 7: Does it use different tags for those branches? Is there ever a world where we also publish Docker images for preview branches?

apps/backend/.env.docker
```diff
@@ -0,0 +1,33 @@
...
+# run Prisma seed script on container startup
+STACK_RUN_SEED_SCRIPT="true"
+
+# Self-host seed script options
+# Optionally create a default admin user and disable signups to the platform when the seed script is run.
+# This can be useful for self hosted deployments where you want to restrict access to the Stack "internal" project.
+# If not specified, no user will be created and you will have to create the first user manually by signing up.
+# STACK_DEFAULT_DASHBOARD_USER_EMAIL=admin@example.com
+# STACK_DEFAULT_DASHBOARD_USER_PASSWORD=admin123
+# STACK_DEFAULT_DASHBOARD_USER_DISPLAY_NAME=Admin
```
N2D4: This feels really dangerous, it means that with the default settings everyone can login to other people's admin accounts because the password is always the same. People don't need admin access to use Stack Auth, I would just either disable this by default or leave it away entirely

apps/backend/.env.docker
```diff
@@ -0,0 +1,33 @@
...
+
+STACK_EMAIL_HOST=inbucket
+STACK_EMAIL_PORT=2500
+STACK_EMAIL_SECURE=false
+STACK_EMAIL_USERNAME=does not matter, ignored by Inbucket
+STACK_EMAIL_PASSWORD=does not matter, ignored by Inbucket
+STACK_EMAIL_SENDER=noreply@example.com
+
+STACK_SVIX_SERVER_URL=http://localhost:8113
+STACK_SVIX_API_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE2NTUxNDA2MzksImV4cCI6MTk3MDUwMDYzOSwibmJmIjoxNjU1MTQwNjM5LCJpc3MiOiJzdml4LXNlcnZlciIsInN1YiI6Im9yZ18yM3JiOFlkR3FNVDBxSXpwZ0d3ZFhmSGlyTXUifQ.En8w77ZJWbd0qrMlHHupHUB-4cx17RfzFykseg95SUk
```
N2D4: Is this API key created by default? Does it not expire? (seems to be a JWT)

apps/backend/next.config.mjs
```diff
@@ -47,6 +47,10 @@ const withConfiguredSentryConfig = (nextConfig) =>
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  // optionally set output to "standalone" for Docker builds
+  // https://nextjs.org/docs/pages/api-reference/next-config-js/output
+  output: process.env.NEXT_CONFIG_OUTPUT,
```
N2D4, line 52: When should this *not* be standalone?

apps/backend/prisma/seed-self-host.ts
N2D4, line 1: Why is this a different script from the usual seed script?

apps/dashboard/.env.docker
```diff
@@ -0,0 +1,10 @@
+NEXT_PUBLIC_STACK_URL=http://localhost:8102
+
+NEXT_PUBLIC_STACK_PROJECT_ID=internal
+NEXT_PUBLIC_STACK_PUBLISHABLE_CLIENT_KEY=this-publishable-client-key-is-for-local-development-only
+STACK_SECRET_SERVER_KEY=this-secret-server-key-is-for-local-development-only
+STACK_SUPER_SECRET_ADMIN_KEY=this-super-secret-admin-key-is-for-local-development-only
```
N2D4: Those are bad default values. Do we need them? Can we just throw an error when people try to build a Docker image without explicitly specifying all environment variables, and then get rid of the `.env.docker` files?

apps/dashboard/next.config.mjs
```diff
@@ -61,6 +61,10 @@ const withConfiguredSentryConfig = (nextConfig) =>
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  // optionally set output to "standalone" for Docker builds
+  // https://nextjs.org/docs/pages/api-reference/next-config-js/output
+  output: process.env.NEXT_CONFIG_OUTPUT,
+
```
N2D4, line 67: as above

apps/dashboard/package.json
```diff
@@ -40,6 +41,7 @@
     "jose": "^5.2.2",
     "lucide-react": "^0.378.0",
     "next": "^14.2.5",
+    "next-runtime-env": "^3.2.2",
```
N2D4: We can get rid of this now, right?

docs/fern/docs/pages/others/self-host.mdx
```diff
@@ -26,6 +26,26 @@ On a high level, Stack Auth is composed of the following services:
...
+docker run -d --name db -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password -e POSTGRES_DB=stackframe -p 5432:5432 postgres:latest
+
+# Run the Docker container
+docker run --env-file your-env-file.env -p 8101:8101 -p 8102:8102 stackauth/server:main
+```
+
+Now you can open the dashboard at [http://localhost:8101](http://localhost:8101) and the API backend on port 8102.
+
+Now, login with your admin account on the dashboard and follow the [normal setup process](/getting-started/setup). Add `NEXT_PUBLIC_STACK_URL=https://your-backend-url.com` to your app's environment variables so that it connects to your API backend instead of the default Stack Auth API backend (https://api.stack-auth.com).
+
```
N2D4: Since we're building & publishing the Docker container to Dockerhub, can't we just tell people here to download that image? I think most people don't want to build the image from scratch

packages/stack/package.json
```diff
@@ -18,7 +18,7 @@
   "homepage": "https://stack-auth.com",
   "scripts": {
     "typecheck": "tsc --noEmit",
-    "build": "rimraf dist && pnpm run codegen && tsup-node",
+    "build": "rimraf dist && pnpm run css && tsup-node",
```
N2D4, line 21: Why?

N2D4 (changes requested):

.github/workflows/docker-build.yaml
fomalhautb: Yes, those are different tags. I don't think we need preview docker for now

apps/backend/.env.docker
fomalhautb: They would need it for the trusted domain. But we can disable it by default for now

apps/backend/.env.docker
fomalhautb: This is just for the local svix. Also this .env file should actually be removed, forgot to do that.

apps/backend/next.config.mjs
fomalhautb: If we deploy on vercel, we don't want it to be standalone

apps/backend/prisma/seed-self-host.ts
fomalhautb: The local script sets up a project optimized for testing, and the self-host script sets up a project for prod use. We can merge the two scripts later to reduce code repetition.

apps/dashboard/.env.docker
fomalhautb: This is also a file that I forgot to delete

apps/dashboard/package.json
fomalhautb: yes

docs/fern/docs/pages/others/self-host.mdx
fomalhautb: This is the docker from docker hub, there is no building process

packages/stack/package.json
fomalhautb: the Quetzel code gen breaks the docker build process. I think ideally we shouldn't do network requests during build

.github/workflows/docker-build.yaml
N2D4: If we don't have preview Docker, we don't know whether the build is gonna succeed until we merge to dev. Is it hard to add it?

apps/backend/next.config.mjs
N2D4: Why not?

apps/backend/.env.docker
N2D4: Let's do that. The seed script runs every time we launch the Docker container, right? We can just make it overwrite the trusted domain with an environment variable so it needs no config at build time and can still be configured at runtime

apps/backend/prisma/seed-self-host.ts
N2D4: What's the difference? Should be the same IMO (except that testing should have some envvars set to insecure values by default), else our testing env deviates from prod

packages/stack/package.json
N2D4: Sounds good, but we should do it during CI/CD in the "Lint and Build" script so we get an error when the quetzal translations are out of date (it errs out when Git shows changes at the end of the build)

N2D4 (changes requested):
Mostly looks good! Some responses and smaller comments, but I like it in general

apps/backend/.env.docker
N2D4: FYI, you can call it .env.docker.local or .env.docker.untracked so it won't be committed to git, to avoid accidents

.github/workflows/docker-build.yaml
```diff
@@ -0,0 +1,49 @@
...
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: ./docker/server/Dockerfile
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
N2D4: Also, can we make Docker use the tag `latest` for our main branch? Then we don't need to specify `stackauth/server:main`

apps/backend/next.config.mjs
fomalhautb: For local dev and Vercel, setting output = standalone only increased build size and build time

.github/workflows/docker-build.yaml
fomalhautb: done

.github/workflows/docker-build.yaml
fomalhautb: Ah I see, I added build for PRs and push to docker hub only on dev/main

PR #358, opened 2024-12-08
Neon project provision

N2D4:
Neon Identity v2 changes as discussed with their team, where Neon can automatically provision a project and only later transfer it to a Stack Auth account; for this, we now have two endpoints /api/v1/integrations/neon/projects/provision (which creates a new project) and /api/v1/integrations/neon/projects/transfer/initiate (which returns a URL that can be used to transfer the project to a Stack Auth user account). Both require client_id:client_secret in basic authorization format, just like the OAuth flows.

### DIFF
--- a/apps/backend/src/app/api/v1/integrations/neon/projects/provision/route.tsx
+++ b/apps/backend/src/app/api/v1/integrations/neon/projects/provision/route.tsx
@@ -1,0 +1,64 @@
+import { createApiKeySet } from "@/lib/api-keys";
+import { createProject } from "@/lib/projects";
+import { prismaClient } from "@/prisma-client";
+import { createSmartRouteHandler } from "@/route-handlers/smart-route-handler";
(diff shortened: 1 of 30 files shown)


apps/backend/src/app/api/v1/integrations/neon/projects/provision/route.tsx
```diff
@@ -0,0 +1,64 @@
...
+  response: yupObject({
+    statusCode: yupNumber().oneOf([200]).defined(),
+    bodyType: yupString().oneOf(["json"]).defined(),
+    body: yupObject({
+      project_id: yupString().defined(),
+      super_secret_admin_key: yupString().defined(),
+    }).defined(),
+  }),
+  handler: async (req) => {
+    const { authorization } = req.headers;
```
fomalhautb: these lines appear multiple times in the codebase, maybe it should be a separate function.

fomalhautb (approved):
looks good

apps/backend/src/app/api/v1/integrations/neon/projects/provision/route.tsx
N2D4: good catch

PR #389, opened 2025-01-10
Fixed separator display when only Passkey is enabled on the Sign In page

hopleus:
If you turn off all oauth providers and enable passkey with email or email & password, the separator is not displayed

----

> [!IMPORTANT]
> Fixes separator display on Sign In page when only Passkey is enabled by updating logic in `auth-page.tsx`.
> 
>   - **Behavior**:
>     - Fixes separator display on Sign In page when only Passkey is enabled by updating `enableSeparator` logic in `auth-page.tsx`.
>     - Adds `hasPasskey` check to determine if Passkey is enabled and should display separator.
>   - **UI Logic**:
>     - Updates condition to render OAuth and Passkey buttons in `Inner` function.
>     - Adjusts condition for displaying "No authentication method enabled" message.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=stack-auth%2Fstack&utm_source=github&utm_medium=referral)<sup> for 6e4cd8cb6b6226b6acb6e16f3217058e04379dcb. It will automatically update as commits are pushed.</sup>


fomalhautb:
thanks!

PR #408, opened 2025-01-29
Vanilla JS client library

fomalhautb:
----

> [!IMPORTANT]
> Introduces a vanilla JavaScript client library for Stack with a new example project, Docker updates, and a template-based package generation system.
> 
>   - **New Features**:
>     - Introduced a vanilla JavaScript client library for Stack.
>     - Added `examples/js-example` with `index.ts` and `.env.development` for demonstration.
>   - **Docker**:
>     - Updated `docker/server/Dockerfile` to include `tsx` and run `generate-from-template.ts`.
>   - **Package Management**:
>     - Added `generate-from-template.ts` script for template-based package generation.
>     - Updated `package.json` scripts to include template generation steps.
>   - **Template System**:
>     - Added `packages/template` with `package-template.json` for environment-specific package generation.
>     - Introduced macros in `generate-from-template.ts` for conditional content inclusion.
>   - **Miscellaneous**:
>     - Updated `.dockerignore` to exclude generated and development files.
>     - Removed `email.tsx` from `packages/stack/src/utils`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=stack-auth%2Fstack&utm_source=github&utm_medium=referral)<sup> for b1f10df8f729aab0123f7f51c22220521f98e7fc. It will automatically update as commits are pushed.</sup>


package.json
```diff
@@ -4,42 +4,44 @@
...
-    "typecheck": "only-allow pnpm && turbo typecheck",
-    "build:dev": "NODE_ENV=development pnpm run build",
-    "build": "only-allow pnpm && turbo build",
-    "build:backend": "only-allow pnpm && turbo run build --filter=@stackframe/stack-backend...",
-    "build:dashboard": "only-allow pnpm && turbo run build --filter=@stackframe/stack-dashboard...",
-    "build:demo": "only-allow pnpm && turbo run build --filter=demo-app...",
-    "build:packages": "only-allow pnpm && turbo run build --filter=./packages/*",
-    "clean": "only-allow pnpm && turbo run clean && rimraf --glob **/.next && rimraf --glob **/.turbo && rimraf .turbo && rimraf --glob **/node_modules",
-    "codegen": "only-allow pnpm && turbo run codegen",
+    "pre": "npx -y only-allow pnpm && node -e \"if(require('fs').existsSync('./scripts/generate-from-template.ts')) require('child_process').execSync('npx --package=ts-node ts-node ./scripts/generate-from-template.ts')\"",
```
N2D4: why do we only do this if the file exists?

packages/template/package-template.json
```diff
@@ -0,0 +1,127 @@
+{
+  "//": "IF_PLATFORM template",
+  "name": "@stackframe/template",
+  "//": "ELSE_IF_PLATFORM js",
+  "name": "@stackframe/js",
+  "//": "ELSE_IF_PLATFORM next",
+  "name": "@stackframe/stack",
+  "//": "END_PLATFORM",
+  "//": "NEXT_LINE_PLATFORM template",
```
N2D4, line 9: why do we need these if we commit the package.json anyways? (btw while duplicate keys are technically allowed by the JSON spec, we should probably avoid them if we can)

packages/template/src/lib/cookie.ts
```diff
@@ -20,6 +20,21 @@ export type CookieHelper = {
   delete: (name: string, options: DeleteCookieOptions) => void,
 };
 
+export async function createEmptyCookieHelper(): Promise<CookieHelper> {
+  function throwError() {
+    throw new Error("Empty cookie helper is just a placeholder. This should never be called");
```
N2D4: what is it for? can you explain it in the code/error message?

N2D4:
```suggestion
    throw new StackAssertionError("Empty cookie helper is just a placeholder. This should never be called");
```

N2D4 (changes requested):
Looks mostly good!

packages/template/src/lib/cookie.ts
fomalhautb: Sure, this is for creating a placeholder cookie helper that doesn't call the Next.js 15 cookie creation if the token store is not cookie

packages/template/package-template.json
fomalhautb: This is the template to generate the `package.json` in `template`, `js`, and `stack`. It is very hard and requires a lot of tricks to make a valid package.json that is at the same time a valid template so I created a separate file for that

package.json
fomalhautb: This is used for the environment where all the code that is not linked to the entry file will be removed(for example in turbo docker build)

PR #414, opened 2025-01-31
User ID Search, Again

TheCactusBlue:


apps/backend/src/app/api/v1/users/crud.tsx
```diff
@@ -465,6 +466,11 @@ export const usersCrudHandlers = createLazyProxy(() => createCrudHandlers(usersC
       } : {},
       ...query.query ? {
         OR: [
+          {
```
fomalhautb: you can do 

```jsx
OR: [
   ...isUuid(query.query) ? [{
     equals: query.query
   }] : [],
]
```

instead of using a ZERO_UUID


also, why don't we use contains for the id? We can then search for partial ID if people type it in

fomalhautb (changes requested):

N2D4 (changes requested):
Can you also update the documentation of the `query` parameter?

apps/backend/src/app/api/v1/users/crud.tsx
TheCactusBlue: fix

packages/stack-shared/src/utils/uuids.tsx
```diff
@@ -10,3 +10,5 @@ export function generateUuid() {
 export function isUuid(str: string) {
   return /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/.test(str);
 }
+
+export const ZERO_UUID = "00000000-0000-0000-0000-000000000000" as const;
```
N2D4, line 14:
```suggestion
```

N2D4 (approved):

PR #436, opened 2025-02-13
add metadata fields to `createStackUser` on `@stackframe/stack`

TheCactusBlue:
> [!IMPORTANT]
> Add metadata fields to user creation options in `stack-app.ts`.
> 
>   - **Behavior**:
>     - Add `clientMetadata`, `clientReadOnlyMetadata`, and `serverMetadata` fields to `ServerUserCreateOptions` in `stack-app.ts`.
>     - Update `serverUserCreateOptionsToCrud()` to map new metadata fields to `UsersCrud["Server"]["Create"]`.
>   - **Misc**:
>     - No changes to existing functionality, only additions to support metadata.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=stack-auth%2Fstack&utm_source=github&utm_medium=referral)<sup> for 5a3dc9322dae1d0acb15de99df7a396e9b14ba03. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/template/src/lib/stack-app.ts
+++ b/packages/template/src/lib/stack-app.ts
@@ -3039,6 +3039,9 @@ type ServerUserCreateOptions = {
   otpAuthEnabled?: boolean,
   displayName?: string,
   primaryEmailVerified?: boolean,
+  clientMetadata?: any,
+  clientReadOnlyMetadata?: any,
+  serverMetadata?: any,
 }
 function serverUserCreateOptionsToCrud(options: ServerUserCreateOptions): UsersCrud["Server"]["Create"] {
   return {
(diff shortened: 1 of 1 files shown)


N2D4 (approved):
maybe add the same fields to `signIn()`? (I think they don't have displayName, etc yet, but they should probably also have that)

fomalhautb:
> maybe add the same fields to `signIn()`? (I think they don't have displayName, etc yet, but they should probably also have that)

That is only easy to do for credential sign-in. For magic link and OAuth we need to handle the data storage in-between and I would say that should be the scope of another PR

N2D4:
so this is mergable?

fomalhautb (approved):

PR #441, opened 2025-02-17
Config json

fomalhautb:
----

> [!IMPORTANT]
> Add configuration parsing and validation using `yup` schemas with merging and schema definition utilities.
> 
>   - **Configuration Parsing and Merging**:
>     - Add `mergeConfigs` and `mergeConfig` functions in `parser.ts` to merge and validate configurations using `yup` schemas.
>     - Handles nested keys, array indices, and null values for key removal.
>   - **Schema Definitions**:
>     - Introduce `getConfigSchema` in `schema.ts` to define configuration structure with `yup`.
>     - Supports environment-specific schema adjustments using `envSchema`.
>   - **Schema Utilities**:
>     - Add `yupRecord` function in `schema-fields.ts` for validating object records with key validation.
>     - Enhance `yup` with custom methods like `nonEmpty` and `getNested` for string and schema manipulation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=stack-auth%2Fstack&utm_source=github&utm_medium=referral)<sup> for f40b691e0e535e921644e0fc580dfc23781258ae. It will automatically update as commits are pushed.</sup>


packages/stack-shared/src/schema-fields.ts
```diff
@@ -197,6 +197,44 @@ export function yupUnion<T extends yup.ISchema<any>[]>(...args: T): yup.MixedSch
...
+        if (keyValidation && !keyValidation(key)) {
+          return createError({
+            path: path,
+            message: `Invalid key: ${key}`,
+          });
+        }
+
+        try {
+          valueSchema.validateSync((value as Record<string, unknown>)[key]);
+        } catch (e: any) {
```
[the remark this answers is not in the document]
N2D4: shitty codemod

packages/stack-shared/src/schema-fields.ts
```diff
@@ -197,6 +197,44 @@ export function yupUnion<T extends yup.ISchema<any>[]>(...args: T): yup.MixedSch
...
+        // Validate the key if a validation function was provided
+        if (keyValidation && !keyValidation(key)) {
+          return createError({
+            path: path,
+            message: `Invalid key: ${key}`,
+          });
+        }
+
+        try {
+          valueSchema.validateSync((value as Record<string, unknown>)[key]);
```
N2D4: this should be `validateAsync`, and you should pass the context and options along. Is there no better way to do this?

packages/stack-shared/src/config/parser.ts
```diff
@@ -0,0 +1,163 @@
...
+    for (let i = 0; i < pathParts.length - 1; i++) {
+      const part = pathParts[i];
+      // Check if part is a valid array index
+      const arrayIndex = parseInt(part);
+      if (!isNaN(arrayIndex) && Array.isArray(target)) {
+        // Handle array index
+        if (!target[arrayIndex]) {
+          target[arrayIndex] = {};
+        }
+        target = target[arrayIndex];
```
N2D4: array indices are just strings, we can simplify this

packages/stack-shared/src/config/parser.ts
```diff
@@ -0,0 +1,163 @@
...
+      defaultConfig: mergedConfig,
+      overrideConfig: config,
+    });
+  }
+
+  return mergedConfig;
+}
+
+export function mergeConfig(options: {
+  configSchema: yup.AnySchema,
```
N2D4: I don't think the schema should be involved here — a merge should just take an override and apply another override on top of it (you currently only allow overriding normalized configs, which is fine for now, but eventually we'll need both — eg. for the PATCH function. The function in stack-info already takes care of that)

packages/stack-shared/src/config/parser.ts
```diff
@@ -0,0 +1,163 @@
...
+  defaultConfig: any,
+  overrideConfig: any,
+}): yup.InferType<typeof options.configSchema> {
+  const newConfig = deepPlainClone(options.defaultConfig) as any;
+
+  Object.keys(options.overrideConfig).forEach((key) => {
```
N2D4: use a for loop instead

packages/stack-shared/src/config/parser.ts
```diff
@@ -0,0 +1,163 @@
...
+      // Check if part is a valid array index
+      const arrayIndex = parseInt(part);
+      if (!isNaN(arrayIndex) && Array.isArray(target)) {
+        // Handle array index
+        if (!target[arrayIndex]) {
+          target[arrayIndex] = {};
+        }
+        target = target[arrayIndex];
+      } else if (typeof target[part] !== 'object' || target[part] === null) {
+        target[part] = {};
```
N2D4: we should throw an error instead

packages/stack-shared/src/config/parser.ts
```diff
@@ -0,0 +1,163 @@
...
+      } else {
+        target = target[part];
+      }
+    }
+
+    const finalKey = pathParts[pathParts.length - 1];
+    if (overrideValue === null) {
+      delete target[finalKey];
```
N2D4: I think setting to null here is more consistent (from the perspective of the merge algorithm, null is just like any other primitive value then). It also has the same behavior anyways

packages/stack-shared/src/config/parser.ts
```diff
@@ -0,0 +1,163 @@
...
+    } else {
+      target[finalKey] = overrideValue;
+    }
+  });
+
+  return options.configSchema.validateSync(newConfig);
```
N2D4: validate asynchronously

packages/stack-shared/src/config/parser.ts
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  return mergedConfig;
+}
+
+export function mergeConfig(options: {
+  configSchema: yup.AnySchema,
+  defaultConfig: any,
+  overrideConfig: any,
+}): yup.InferType<typeof options.configSchema> {
```
N2D4: can you add some validation to this function? in particular:

- throw an error if a key does not follow the format
- throw an error when deeply trying to set a nested key that doesn't exist
- throw an error if the input object is not an object

StackAssertionErrors on all of them

N2D4 (changes requested):
- prototype pollution
- multiple types of schemas (one for projects, one for overrides, etc etc)

PR #448, opened 2025-02-20
Added magic link, OTP, and OAuth examples to JS docs

fomalhautb:


packages/template/src/lib/stack-app.ts
```diff
@@ -46,11 +46,11 @@ isReactServer = sc.isReactServer;
 // NextNavigation.useRouter does not exist in react-server environments and some bundlers try to be helpful and throw a warning. Ignore the warning.
 const NextNavigation = scrambleDuringCompileTime(NextNavigationUnscrambled);
 
-const clientVersion = process.env.STACK_COMPILE_TIME_CLIENT_PACKAGE_VERSION ?? throwErr("Missing STACK_COMPILE_TIME_CLIENT_PACKAGE_VERSION. This should be a compile-time variable set by Stack's build system.");
+const clientVersion = "STACK_COMPILE_TIME_CLIENT_PACKAGE_VERSION_SENTINEL";
```
N2D4, line 49: can we make this throw an error if it throws for any reason? (I prefer the process.env version because it's more
```suggestion
const clientVersion = "STACK_COMPILE_TIME_CLIENT_PACKAGE_VERSION_SENTINEL";
if (clientVersion.startsWith("STACK_COMPILE_TIME") {
  throw new StackAssertionError("Client version was not replaced. Something went wrong during build!");
}
```

PR #451, opened 2025-02-20
Typo fix in auth method settings page

ShrootBuck:
just a small typo I noticed


TheCactusBlue:
Thanks for the contrib!

PR #457, opened 2025-02-21
Fix MS OAuth

fomalhautb:


apps/backend/package.json
```diff
@@ -28,12 +28,14 @@
...
     "db-seed-script": "pnpm run with-env tsx prisma/seed.ts",
-    "verify-data-integrity": "pnpm run with-env tsx scripts/verify-data-integrity.ts"
+    "verify-data-integrity": "pnpm run with-env tsx scripts/verify-data-integrity.ts",
+    "postinstall": "patch-package"
   },
   "prisma": {
     "seed": "pnpm run db-seed-script"
   },
   "dependencies": {
+    "@milahu/patch-package": "^6.4.14",
```
N2D4: this was last updated 3 years ago by someone whose name I never heard before with 6k downloads — is there no other way to solve this?

apps/backend/src/oauth/providers/microsoft.tsx
```diff
@@ -14,12 +14,17 @@ export class MicrosoftProvider extends OAuthBaseProvider {
     clientSecret: string,
     microsoftTenantId?: string,
   }) {
+    const tenantId = encodeURIComponent(options.microsoftTenantId || "consumers");
     return new MicrosoftProvider(...await OAuthBaseProvider.createConstructorArgs({
-      issuer: `https://login.microsoftonline.com${"/" + options.microsoftTenantId || ""}`,
-      authorizationEndpoint: `https://login.microsoftonline.com/${options.microsoftTenantId || 'consumers'}/oauth2/v2.0/authorize`,
-      tokenEndpoint: `https://login.microsoftonline.com/${options.microsoftTenantId || 'consumers'}/oauth2/v2.0/token`,
+      // Note that it is intentional to have tenantid instead of tenantId, also intentional to not be a template literal.
+      // This will be replaced by the openid-client library. The package patch is applied to enable aadIssValidation for this
```
N2D4: any link explaining where you got this patch from that you could add here?

N2D4 (changes requested):

apps/backend/package.json
fomalhautb: just realized pnpm has this feature in-built. Moved to that

apps/backend/src/oauth/providers/microsoft.tsx
fomalhautb: added more comments

N2D4 (approved):

PR #526, opened 2025-03-10
Auto migration

fomalhautb:
----

> [!IMPORTANT]
> Introduces an automated database migration system, replacing manual Prisma commands with new scripts and updating workflows, configurations, and tests accordingly.
> 
>   - **Auto-Migration System**:
>     - Introduces `db-migrations.ts` script for handling database migrations automatically.
>     - Adds utility functions in `utils.tsx` for managing migration files.
>     - Implements `applyMigrations` and `runMigrationNeeded` in `index.tsx` for executing migrations.
>   - **Workflow and Scripts**:
>     - Updates GitHub workflows (`check-prisma-migrations.yaml`, `e2e-api-tests.yaml`) to use new migration commands.
>     - Replaces `prisma migrate` commands with `db:init`, `db:migrate`, etc., in `package.json` and `README.md`.
>   - **Testing**:
>     - Adds `auto-migration.tests.ts` for testing migration logic and concurrency handling.
>   - **Configuration**:
>     - Updates `.env.development` and `vitest.config.ts` for new environment variables and paths.
>     - Modifies `turbo.json` and `package.json` to include new migration tasks and scripts.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=stack-auth%2Fstack-auth&utm_source=github&utm_medium=referral)<sup> for 2c241838793f54da34f25b0078f0803ae6770554. You can [customize](https://app.ellipsis.dev/stack-auth/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


apps/backend/prisma/migrations/20240910211533_remove_shared_facebook/migration.sql
```diff
@@ -37,10 +37,8 @@ DELETE FROM "ProxiedOAuthProviderConfig"
 WHERE "type" = 'FACEBOOK';
 
 -- AlterEnum
-BEGIN;
 CREATE TYPE "ProxiedOAuthProviderType_new" AS ENUM ('GITHUB', 'GOOGLE', 'MICROSOFT', 'SPOTIFY');
 ALTER TABLE "ProxiedOAuthProviderConfig" ALTER COLUMN "type" TYPE "ProxiedOAuthProviderType_new" USING ("type"::text::"ProxiedOAuthProviderType_new");
 ALTER TYPE "ProxiedOAuthProviderType" RENAME TO "ProxiedOAuthProviderType_old";
 ALTER TYPE "ProxiedOAuthProviderType_new" RENAME TO "ProxiedOAuthProviderType";
 DROP TYPE "ProxiedOAuthProviderType_old";
-COMMIT;
```
N2D4: can we no longer have transactions inside migration files now?

apps/backend/prisma/migrations/20250304200822_add_project_user_count/migration.sql
```diff
@@ -8,6 +8,8 @@ UPDATE "Project" SET "userCount" = (
 );
 
 -- Create function to update userCount
+-- SPLIT_STATEMENT_SENTINEL
+-- SINGLE_STATEMENT_SENTINEL
```
N2D4, line 12: what does this do?

apps/backend/scripts/db-migrations.ts
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const getDropDBPrompt = async () => {
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout
+  });
+
+  const answer = await new Promise<string>(resolve => {
+    rl.question('Are you sure you want to drop everything in the database? This action cannot be undone. (Y/n): ', resolve);
```
N2D4: this should be default N

apps/backend/scripts/db-migrations.ts
```diff
@@ -0,0 +1,76 @@
...
+  await prismaClient.$executeRaw`CREATE SCHEMA public`;
+  await prismaClient.$executeRaw`GRANT ALL ON SCHEMA public TO postgres`;
+  await prismaClient.$executeRaw`GRANT ALL ON SCHEMA public TO public`;
+};
+
+const seed = async () => {
+  execSync('pnpm run db-seed-script', { stdio: 'inherit' });
+};
+
+const getDropDBPrompt = async () => {
```
N2D4:
```suggestion
const promptDropDb = async () => {
```

apps/backend/scripts/db-migrations.ts
```diff
@@ -0,0 +1,76 @@
...
+    migrationFiles: getMigrationFiles(MIGRATION_FILES_DIR)
+  });
+  await seed();
+};
+
+const main = async () => {
+  const args = process.argv.slice(2);
+  const command = args[0];
+
+  switch (command) {
```
N2D4, line 63: can we add a help command that documents each command?

apps/backend/scripts/generate-migration-imports.ts
```diff
@@ -0,0 +1,9 @@
+import { writeFileSyncIfChanged } from '@stackframe/stack-shared/dist/utils/fs';
+import path from 'path';
+import { MIGRATION_FILES_DIR, getMigrationFiles } from '../src/auto-migrations/utils';
+
+const migrationFiles = getMigrationFiles(MIGRATION_FILES_DIR);
+
+writeFileSyncIfChanged(path.join(process.cwd(),  'src', 'auto-migrations', 'migration-files.ts'), `
```
N2D4:
```suggestion
writeFileSyncIfChanged(path.join(process.cwd(), 'src', 'auto-migrations', 'migration-files.ts'), `
```

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+}
+
+
+export async function applyMigrations(options: {
+  prismaClient: PrismaClient,
+  migrationFiles?: Array<{ migrationName: string, sql: string }>,
+  artificialDelayInSeconds?: number,
+}): Promise<{
+  newlyAppliedMigrationNames: string[],
+}> {
```
N2D4: add race condition test

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+  for (const [index, appliedMigration] of appliedMigrations.entries()) {
+    if (appliedMigration !== migrationFiles[index].migrationName) {
+      throw new StackAssertionError(`Migration is applied out of order`);
+    }
+  }
+
+  const transactions = [];
+  const newMigrationFiles = migrationFiles.slice(appliedMigrations.length);
+
+  for (const migration of migrationFiles.slice(appliedMigrations.length)) {
```
N2D4:
```suggestion
  for (const migration of newMigrationFiles) {
```

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+    prismaClient: options.prismaClient,
+  });
+
+  for (const [index, appliedMigration] of appliedMigrations.entries()) {
+    if (appliedMigration !== migrationFiles[index].migrationName) {
+      throw new StackAssertionError(`Migration is applied out of order`);
+    }
+  }
+
+  const transactions = [];
```
N2D4:
```suggestion
  const transaction = [];
```

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+    } else {
+      throw error;
+    }
+  }
+
+  return { newlyAppliedMigrationNames: newMigrationFiles.map(x => x.migrationName) };
+};
+
+export function getMigrationCheckQuery(options?: {
+  migrationFiles?: Array<{ migrationName: string, sql: string }>,
```
N2D4:
```suggestion
  migrationFiles?: { migrationName: string, sql: string }[],
```

apps/backend/src/prisma-client.tsx
```diff
@@ -123,7 +132,14 @@ async function rawQueryArray<Q extends RawQuery<any>[]>(queries: Q): Promise<[]
...
 
-    const rawResult = await prismaClient.$queryRaw(query) as { type: string, json: any }[];
+    const [_, rawResult] = await runQueryAndMigrateIfNeeded({
+      prismaClient,
+      fn: async () => await prismaClient.$transaction([
+        prismaClient.$queryRaw(getMigrationCheckQuery()),
+        prismaClient.$queryRaw(query),
+      ])
+    }) as [unknown, { type: string, json: any }[]];
+
```
N2D4: make single roundtrip

apps/backend/src/auto-migrations/utils.tsx
```diff
@@ -0,0 +1,31 @@
...
+import path from "path";
+
+export const MIGRATION_FILES_DIR = path.join(process.cwd(),  'prisma', 'migrations');
+
+export function getMigrationFiles(migrationDir: string): Array<{ migrationName: string, sql: string }> {
+  const folders = fs.readdirSync(migrationDir).filter(folder =>
+    fs.statSync(path.join(migrationDir, folder)).isDirectory()
+  );
+
+  const result: Array<{ migrationName: string, sql: string }> = [];
```
N2D4:
```suggestion
  const result: { migrationName: string, sql: string }[] = [];
```

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+  fn: () => Promise<T>,
+  artificialDelayInSeconds?: number,
+}): Promise<T> {
+  try {
+    return await options.fn();
+  } catch (e) {
+    const migrationError = getMigrationError(e);
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (migrationError === 'MIGRATION_NEEDED') {
```
N2D4: switch-case

N2D4 (changes requested):
- update migration logic

what do I need to know about this before pushing new DB changes in the future? what do we need to do to port existing databases over?

apps/backend/prisma/migrations/20240910211533_remove_shared_facebook/migration.sql
fomalhautb: We can, but then we need the SPLIT_STATEMENT_SENTINEL

apps/backend/package.json
N2D4, line 1: how will the new scripts work with source-of-truth? assumably we also have to update the new github workflows when we merge the branch? (just putting here as a reminder for when you merge branches)

apps/backend/.gitignore
```diff
@@ -36,3 +36,5 @@ yarn-error.log*
 # typescript
 *.tsbuildinfo
 next-env.d.ts
+
+src/auto-migrations/migration-files.tsx
```
N2D4: can we put this in src/generated/auto-migrations or so?

N2D4: and then the .gitignore can be in src/generated like in the client library

apps/backend/src/prisma-client.tsx
```diff
@@ -48,6 +49,13 @@ export async function retryTransaction<T>(fn: (tx: PrismaClientTransaction) => P
             return Result.ok(await prismaClient.$transaction(async (tx, ...args) => {
               let res;
               try {
+                await runQueryAndMigrateIfNeeded({
+                  prismaClient,
+                  fn: async () => {
+                    await tx.$queryRaw(getMigrationCheckQuery());
+                  },
+                });
+
```
N2D4: run when the prismaclient is created instead

apps/backend/src/auto-migrations/auto-migration.tests.ts
```diff
@@ -0,0 +1,313 @@
...
+const TEST_DB_PREFIX = 'stack_auth_test_db';
+
+const getTestDbURL = (testDbName: string) => {
+  // @ts-ignore - ImportMeta.env is provided by Vite
+  const base = import.meta.env.STACK_DIRECT_DATABASE_CONNECTION_STRING.replace(/\/[^/]*$/, '');
+  return {
+    full: `${base}/${testDbName}`,
+    base,
+  };
+};
```
N2D4, line 15: this feels kinda hacky, also you'd expect that tests outside `e2e` wouldn't need any infrastructure running. can we use [pglite-socket](https://pglite.dev/docs/pglite-socket) to create a new DB every time we call this function (it needs to be cleaned up at the end of the test of course) on a random port and return that?

N2D4: then we can also replace all the teardown with just closing the pglite + pglite-socket objects, instead of having to remove stuff with sql

apps/backend/vitest.config.ts
```diff
@@ -1,7 +1,25 @@
...
+        ...loadEnv('development', process.cwd(), ''),
+      },
+    },
+    resolve: {
+      alias: {
+        '@': resolve(__dirname, './src')
+      }
+    },
+    envDir: __dirname,
+    envPrefix: 'STACK_',
```
N2D4: can we simplify this if we don't need the envvars?

```suggestion
    resolve: {
      alias: {
        '@': resolve(__dirname, './src')
      }
    },
```

apps/backend/scripts/db-migrations.ts
```diff
@@ -0,0 +1,104 @@
...
+    case 'reset': {
+      await promptDropDb();
+      await dropPublicSchema();
+      await migrate();
```
N2D4, line 67: should seed

apps/backend/scripts/db-migrations.ts
```diff
@@ -0,0 +1,104 @@
...
+  console.log(`Database Migration Script
+
+Usage: pnpm db-migrations <command>
+
+Commands:
+  reset                    Drop all data and recreate the database, then apply migrations and seed
+  generate-migration-file  Generate a new migration file using Prisma, then reset and migrate
+  seed                     Run database seeding only
+  init                     Apply migrations and seed the database
+  help                     Show this help message
```
N2D4:
```suggestion
  seed                     [Advanced] Run database seeding only
  init                     Apply migrations and seed the database
  help                     Show this help message
```

apps/backend/scripts/db-migrations.ts
```diff
@@ -0,0 +1,104 @@
...
+};
+
+const main = async () => {
+  const args = process.argv.slice(2);
+  const command = args[0];
+
+  switch (command) {
+    case 'reset': {
+      await promptDropDb();
+      await dropPublicSchema();
```
N2D4: this should drop the current schema, not necessarily the public schema

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,249 @@
+import { Prisma, PrismaClient } from '@prisma/client';
+import { MIGRATION_FILES } from './migration-files';
+
+
+function getMigrationError(error: unknown): string {
+  if (error instanceof Prisma.PrismaClientKnownRequestError && error.code === 'P2010') {
+    if (error.meta?.code === 'P0001') {
```
N2D4, line 17: can you add a comment with the error messages/meaning of these codes?

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+      const errorName = (error.meta as { message: string }).message.split(' ')[1];
+      return errorName;
+    }
+  }
+  throw error;
+}
+
+function isMigrationNeeded(error: unknown): boolean {
```
N2D4:
```suggestion
function isMigrationNeededError(error: unknown): boolean {
```

updated the function name, because it returning `false` doesn't mean that no migration is needed

apps/backend/src/auto-migrations/auto-migration.tests.ts
```diff
@@ -0,0 +1,313 @@
...
+    runMigrationAndInsert('test_value2'),
+  ]);
+
+  const result1 = await prismaClient.$queryRaw`SELECT name FROM test` as { name: string }[];
+  expect(Array.isArray(result1)).toBe(true);
+  expect(result1.length).toBe(2);
+  expect(result1.some(r => r.name === 'test_value1')).toBe(true);
+  expect(result1.some(r => r.name === 'test_value2')).toBe(true);
+}));
+
```
N2D4, line 286: a migration that fails for whatever reasons rolls back all statements successfully, and then reapplying a fixed version of the migration is also successful

apps/backend/scripts/db-migrations.ts
```diff
@@ -0,0 +1,104 @@
...
+  await seed();
+};
+
+const showHelp = () => {
+  console.log(`Database Migration Script
+
+Usage: pnpm db-migrations <command>
+
+Commands:
+  reset                    Drop all data and recreate the database, then apply migrations and seed
```
N2D4, line 50: add "apply-migrations"

apps/backend/src/auto-migrations/auto-migration.tests.ts
```diff
@@ -0,0 +1,313 @@
...
+    sql: `ALTER TABLE "User" ADD COLUMN "age" INTEGER NOT NULL DEFAULT 0;`,
+  },
+];
+
+
+import.meta.vitest?.test("connects to DB", runTest(async ({ expect, prismaClient }) => {
+  const result = await prismaClient.$executeRaw`SELECT 1`;
+  expect(result).toBe(1);
+}));
+
```
N2D4, line 156: does not apply a migration again if all migrations are already applied

does not apply a migration again if all migrations are already applied, and some future migrations are also applied (we need this for rollbacks)

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+  ]);
+
+  return (appliedMigrations as { migrationName: string }[]).map((migration) => migration.migrationName);
+}
+
+async function acquireMigrationLock(options: {
+  prismaClient: PrismaClient,
+}) {
+  const maxRetries = 10;
+  let backOffInMs = 100;
```
N2D4: since it's exponential, we'll never hit 10 max retries. this is the sort of thing where I think we actually don't need exponential backoff — exponential backoff is useful to prevent overloading a service, but in this case we are just waiting gracefully until the migration is over. It probably doesn't harm to have exponential backoff regardless, but the basis should be much lower (like randomly between 1.5 and 2, with 15 maxRetries)

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+            
+            -- Check if started_at exists and is within 10 seconds
+            IF started_at_value IS NOT NULL AND started_at_value >= clock_timestamp() - INTERVAL '10 seconds' THEN
+              RAISE EXCEPTION 'MIGRATION_IN_PROGRESS';
+            ELSIF started_at_value IS NOT NULL AND started_at_value < clock_timestamp() - INTERVAL '10 seconds' THEN
+              -- Update the startedAt to current timestamp since it's older than 10 seconds and start this migration
+              UPDATE "SchemaMigrationLock" SET "startedAt" = clock_timestamp() WHERE id = 1;
+            END IF;
+          END IF;
+        END $$;
```
N2D4: this lock might still cause errors actually — since it's only isolation level READ_COMMITTED, the values read *within* this rawQuery might change from line to line. instead, the right abstraction to use is probably advisory locks, which are built-in postgres locks that you can easily create (and release). so we won't need to make our own SchemaMigrationLock table. then, we also don't need to backoff here (we can just wait until the lock is released)

apps/backend/src/auto-migrations/index.tsx
N2D4: but I think you don't need to implement this, see the comments below, you can use pg advisory locks

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+      if (getMigrationError(e) === 'MIGRATION_IN_PROGRESS') {
+        await new Promise(resolve => setTimeout(resolve, backOffInMs));
+        backOffInMs *= 2 * (1 + Math.random());
+      } else {
+        throw e;
+      }
+    }
+  }
+
+  throw new Error('Failed to acquire migration lock');
```
N2D4:
```suggestion
  throw new StackAssertionError('Failed to acquire migration lock');
```

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+      transaction.push(options.prismaClient.$executeRaw`
+        UPDATE "SchemaMigration" SET "finishedAt" = clock_timestamp()
+        WHERE "migrationName" = ${migration.migrationName}
+      `);
+
+      await options.prismaClient.$transaction(transaction);
+    } catch (e) {
+      await getRemoveMigrationLockQuery({ prismaClient: options.prismaClient });
+      throw e;
+    }
```
N2D4: better to use try-finally for this (also avoids code duplication, and prevents situations such as us not releasing the lock if it throws on line 197)

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+  prismaClient: PrismaClient,
+  migrationFiles?: { migrationName: string, sql: string }[],
+  artificialDelayInSeconds?: number,
+  logging?: boolean,
+}): Promise<{
+  newlyAppliedMigrationNames: string[],
+}> {
+  const migrationFiles = options.migrationFiles ?? MIGRATION_FILES;
+
+  await acquireMigrationLock({ prismaClient: options.prismaClient });
```
N2D4:
```suggestion
  // TODO we should not acquire a lock if the migrations are already up-to-date, and at most once for every migration that is being applied. (right now, in a long-running migration which will cause many servers to restart, all those servers after the first one would queue up and lock one after the other just to see that we're already up-to-date and release the lock again. in the future, we could use a mix of shared and exclusive locks to make sure that only one client at a time attempts to get an exclusive lock, and if it fails to get that, goes back to acquiring a shared lock instead)
  await acquireMigrationLock({ prismaClient: options.prismaClient });
```

would be better if we don't acquire a lock when all migrations are up-to-date, but I think we can just leave a todo here and worry later

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+            -- Return the startedAt value since the row already existed
+            SELECT "startedAt" FROM "SchemaMigrationLock" WHERE id = 1 INTO started_at_value;
+            
+            -- Check if started_at exists and is within 10 seconds
+            IF started_at_value IS NOT NULL AND started_at_value >= clock_timestamp() - INTERVAL '10 seconds' THEN
+              RAISE EXCEPTION 'MIGRATION_IN_PROGRESS';
+            ELSIF started_at_value IS NOT NULL AND started_at_value < clock_timestamp() - INTERVAL '10 seconds' THEN
+              -- Update the startedAt to current timestamp since it's older than 10 seconds and start this migration
+              UPDATE "SchemaMigrationLock" SET "startedAt" = clock_timestamp() WHERE id = 1;
+            END IF;
```
N2D4: I think this will still potentially cause multiple migrations at the same time? the transactions are not cancelled if the lock is given to someone else after 10 seconds

N2D4 (changes requested):

apps/backend/src/auto-migrations/index.tsx
```diff
@@ -1,8 +1,17 @@
+import { sqlQuoteIdent } from '@/prisma-client';
 import { Prisma, PrismaClient } from '@prisma/client';
-import { MIGRATION_FILES } from './migration-files';
+import { MIGRATION_FILES } from './../generated/migration-files';
 
+const MIGRATION_LOCK_ID = 59129034;
```
N2D4, line 6: can we give this a more readable name?
```suggestion
const MIGRATION_LOCK_ID = "stack_migration_lock";
```

PR #557, opened 2025-03-19
Payment integration

fomalhautb:
image


TheCactusBlue:
closing and reopening (if zai does not mind) so that ellipsis is less confused

PR #562, opened 2025-03-20
Fix/disable forgot password when auth disabled

### LINKED ISSUE #292: /handler/forgot-password page shouldn't be shown when password auth is disabled
The page should show a message card saying that password authentication is disabled on the client side.
The stack handler should redirect it back to home on the server side.

This is similar to how the other pages work

Naresh-official:
This PR addresses the issue where the /handler/forgot-password page should not be accessible when password authentication is disabled. Additionally, it improves the redirection behavior when a user is already logged in.

### Changes:

Added a redirection check in the useEffect hook to ensure that logged-in users are redirected to the appropriate page instead of staying on the forgot-password page.

### Fixes:
Closes #292

### Test:
[Screencast from 2025-03-20 22-18-36.webm](https://github.com/user-attachments/assets/d626ba95-9d7f-46e6-b232-0320b171d130)

----

> [!IMPORTANT]
> Redirects logged-in users from `ForgotPassword` page and prevents access when password authentication is disabled.
> 
>   - **Behavior**:
>     - Redirects logged-in users from `ForgotPassword` page using `useEffect` in `forgot-password.tsx`.
>     - Prevents access to `ForgotPassword` when password authentication is disabled.
>   - **Functions**:
>     - Adds `useEffect` hook in `ForgotPassword` to handle redirection.
>   - **Fixes**:
>     - Closes issue #292.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=stack-auth%2Fstack-auth&utm_source=github&utm_medium=referral)<sup> for 9c93cb407897b901877f8f823b3c96ec7e7eebfb. It will automatically update as commits are pushed.</sup>


N2D4:
this broke the non-Next packages, so reverting it for now

PR #580, opened 2025-03-26
fix: local development with docker_compose needs this extra var

j-sp4:
When developing Locally with docker compose stack sets the env var NEXT_PUBLIC_STACK_SVIX_SERVER_URL to STACK_SVIX_SERVER_URL, this then breaks webhooks due to the fact that STACK_SVIX_SERVER_URL is called within docker and NEXT_PUBLIC_STACK_SVIX_SERVER_URL is called from the frontend, in my specific case:
NEXT_PUBLIC_STACK_SVIX_SERVER_URL=http://localhost:8071
STACK_SVIX_SERVER_URL=http://svix:8071

Hopefully this helps the next person 😃

### DIFF
--- a/docker/server/.env
+++ b/docker/server/.env
@@ -26,6 +26,7 @@ STACK_EMAIL_SENDER=
 
 # Set these if you want to use webhooks
 STACK_SVIX_SERVER_URL=# this is only needed if you self-host the Svix service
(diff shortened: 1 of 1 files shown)


fomalhautb:
Thanks! Looks great! Can you sign the CLA?

PR #683, opened 2025-05-15
Domain should be always defined in neon domain crud endpoint

fomalhautb:


N2D4 (approved):

N2D4 (approved):
Can you maybe add an E2E test?

PR #707, opened 2025-06-10
Remove commented-out code for sending verification emails

fomalhautb:
There is no redirect url for this feature


fomalhautb:
Will be fixed in #711

PR #720, opened 2025-06-20
Email normalization

fomalhautb:

### DIFF
--- a/apps/backend/src/lib/emails.tsx
+++ b/apps/backend/src/lib/emails.tsx
@@ -367,3 +367,78 @@ export async function getSharedEmailConfig(displayName: string): Promise<EmailCo
     type: 'shared',
   };
 }
+
+// modified from https://github.com/johno/normalize-email/tree/master
+export function normalizeEmail(email: string): string {
+  if (typeof email !== 'string') {
+    throw new TypeError('normalize-email expects a string');
+  }
+
+  const PLUS_ONLY = /\+.*$/;
+  const PLUS_AND_DOT = /\.|\+.*$/g;
+
+  const providers = {
+    'gmail.com': {
+      cut: PLUS_AND_DOT
+    },
+    'googlemail.com': {
+      cut: PLUS_AND_DOT,
+      aliasOf: 'gmail.com'
+    },
+    'hotmail.com': {
+      cut: PLUS_ONLY
+    },
+    'live.com': {
+      cut: PLUS_AND_DOT
+    },
+    'outlook.com': {
+      cut: PLUS_ONLY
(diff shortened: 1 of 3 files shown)


apps/backend/src/lib/emails.tsx
```diff
@@ -367,3 +367,78 @@ export async function getSharedEmailConfig(displayName: string): Promise<EmailCo
...
+    'hotmail.com': {
+      cut: PLUS_ONLY
+    },
+    'live.com': {
+      cut: PLUS_AND_DOT
+    },
+    'outlook.com': {
+      cut: PLUS_ONLY
+    }
+  };
```
N2D4: I don't think we should do this, at least not by default. The fact that you can add + to an email is a feature IMO. Cutting the dot and using googlemail as an alias for gmail sounds reasonable tho

apps/backend/src/lib/emails.tsx
```diff
@@ -367,3 +367,78 @@ export async function getSharedEmailConfig(displayName: string): Promise<EmailCo
     type: 'shared',
   };
 }
+
+// modified from https://github.com/johno/normalize-email/tree/master
```
N2D4:
```suggestion
```

apps/backend/src/lib/emails.tsx
```diff
@@ -367,3 +367,78 @@ export async function getSharedEmailConfig(displayName: string): Promise<EmailCo
...
+    },
+    'live.com': {
+      cut: PLUS_AND_DOT
+    },
+    'outlook.com': {
+      cut: PLUS_ONLY
+    }
+  };
+
+  const emailLower = email.toLowerCase();
```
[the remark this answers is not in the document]
N2D4: ya, I think this function should trim whitespaces

apps/backend/src/lib/emails.tsx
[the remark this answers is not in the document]
N2D4: good bot

apps/backend/src/lib/emails.tsx
```diff
@@ -367,3 +367,78 @@ export async function getSharedEmailConfig(displayName: string): Promise<EmailCo
...
+  const emailParts = emailLower.split(/@/);
+
+  if (emailParts.length !== 2) {
+    return email;
+  }
+
+  let [username, domain] = emailParts;
+
+  if (domain in providers) {
+    const provider = providers[domain as keyof typeof providers];
```
N2D4: map as above

N2D4 (changes requested):

PR #751, opened 2025-07-09
feat: new api endpoints to get and update neon managed projects

Shridhad:
Added GET and PATCH endpoint for Neon managed project with `secret-admin-key` auth. The API handler is same as internal projects API - `/internal/projects/current`


N2D4:
Merging this! Please note that we will deprecate this method of updating projects relatively soon — we'll let you know beforehand though so you can update your API calls first.

PR #787, opened 2025-07-22
template and theme saving, simpler rendering

BilalG1:


N2D4 (approved):
wish I had more to say to these PRs but they look good haha

PR #794, opened 2025-07-23
sending new templates

BilalG1:


apps/backend/src/lib/email-rendering.tsx
```diff
@@ -1,8 +1,22 @@
...
 import { getEnvVariable, getNodeEnvironment } from '@stackframe/stack-shared/dist/utils/env';
 import { Result } from "@stackframe/stack-shared/dist/utils/results";
 import { deindent } from "@stackframe/stack-shared/dist/utils/strings";
 import { bundleJavaScript } from '@stackframe/stack-shared/dist/utils/esbuild';
+import { StackAssertionError } from '@stackframe/stack-shared/dist/utils/errors';
+
+export function getActiveEmailTheme(tenancy: Tenancy) {
+  const themeList = tenancy.completeConfig.emails.themeList;
+  const currentActiveTheme = tenancy.completeConfig.emails.theme;
+  if (!(currentActiveTheme in themeList)) {
```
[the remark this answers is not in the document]
N2D4: this is a good catch actually — not a Map, but we have a `has` function that's immune to prototype pollution


```suggestion
  if (!(has(themeList, "currentActiveTheme"))) {
```

apps/backend/src/lib/email-rendering.tsx
```diff
@@ -1,8 +1,34 @@
...
+      currentActiveTheme,
+    });
+  }
+  return themeList[currentActiveTheme];
+}
+
+export function getEmailThemeForTemplate(tenancy: Tenancy, templateThemeId: string | null | false | undefined) {
+  const themeList = tenancy.completeConfig.emails.themes;
+  if (templateThemeId && templateThemeId in themeList) {
+    return themeList[templateThemeId].tsxSource;
```
N2D4:
```suggestion
    return get(themeList, "templateThemeId").tsxSource;
```

apps/backend/src/lib/emails.tsx
```diff
@@ -12,6 +12,29 @@ import { Result } from '@stackframe/stack-shared/dist/utils/results';
...
+    return templateList[DEFAULT_TEMPLATE_IDS.magic_link];
+  }
+  if (type === "team_invitation") {
+    return templateList[DEFAULT_TEMPLATE_IDS.team_invitation];
+  }
+  if (type === "sign_in_invitation") {
+    return templateList[DEFAULT_TEMPLATE_IDS.sign_in_invitation];
+  }
+  throw new StackAssertionError(`Unknown email template type: ${type}`);
+}
```
N2D4: greptile's suggestion here was also good — I applied it

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/email-templates/[templateId]/page-client.tsx
```diff
@@ -0,0 +1,139 @@
...
+  if (themeId === false) return "false-sentinel";
+  if (themeId === undefined) return "undefined-sentinel";
+  return themeId;
+}
+
+function selectStringToThemeId(value: string): string | undefined | false {
+  if (value === "false-sentinel") return false;
+  if (value === "undefined-sentinel") return undefined;
+  return value;
+}
```
Comment (N2D4) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/email-templates/[templateId]/page-client.tsx]:
```suggestion
function themeIdToSelectString(themeId: string | undefined | false): string {
  return JSON.stringify(themeId ?? null);
}

function selectStringToThemeId(value: string): string | undefined | false {
  return JSON.parse(value) ?? undefined;
}
```

packages/stack-shared/src/config/format.ts
```diff
@@ -9,7 +9,7 @@ export type Config = {
   [keyOrDotNotation: string]: ConfigValue | undefined,  // must support undefined for optional values
 };
 
-export type NormalizedConfigValue = string | number | boolean | NormalizedConfig | NormalizedConfigValue[];
+export type NormalizedConfigValue = string | number | boolean | null | NormalizedConfig | NormalizedConfigValue[];
```
N2D4: Normalized configs should never contain null

```suggestion
export type NormalizedConfigValue = string | number | boolean | NormalizedConfig | NormalizedConfigValue[];
```

apps/backend/src/lib/emails.tsx
BilalG1: I was getting lint issues with this, using maps instead

PR #805, opened 2025-07-25
Update to single changelog source of truth for entire monorepo

madster456:
Consolidates all changelog into single file.


scripts/interactive-changelog.ts
```diff
@@ -0,0 +1,244 @@
...
+    '@stackframe/example-demo-app',
+    '@stackframe/docs-examples',
+    '@stackframe/example-middleware-demo',
+    '@stackframe/cjs-test',
+    '@stackframe/js-example',
+    '@stackframe/react-example',
+    '@stackframe/supabase-example',
+    '@stackframe/e-commerce-example',
+    '@stackframe/partial-prerendering-example'
+  ];
```
N2D4: Can we find those dynamically? They get outdated quickly

N2D4 (commented):
What's the new workflow

PR #858, opened 2025-08-21
Payments test mode

BilalG1:

### DIFF
--- a/apps/dashboard/src/lib/utils.tsx
+++ b/apps/dashboard/src/lib/utils.tsx
@@ -1,4 +1,5 @@
 import { getPublicEnvVar } from "@/lib/env";
+import { parseJson } from "@stackframe/stack-shared/dist/utils/json";
 import { clsx, type ClassValue } from "clsx";
 import { redirect } from "next/navigation";
 import { twMerge } from "tailwind-merge";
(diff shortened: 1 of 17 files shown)


apps/dashboard/src/lib/utils.tsx
```diff
@@ -20,7 +21,7 @@ export function devFeaturesEnabledForProject(projectId: string) {
   if (projectId === "internal") {
     return true;
   }
-  const allowedProjectIds = JSON.parse(getPublicEnvVar("NEXT_PUBLIC_STACK_ENABLE_DEVELOPMENT_FEATURES_PROJECT_IDS") || "[]");
+  const allowedProjectIds = parseJson(getPublicEnvVar("NEXT_PUBLIC_STACK_ENABLE_DEVELOPMENT_FEATURES_PROJECT_IDS") || "[]");
```
N2D4, line 24: accident?

N2D4: I don't think this owrks, parseJson returns a result

N2D4 (approved):

apps/e2e/tests/backend/endpoints/api/v1/payments/purchase-session.test.ts
N2D4, line 1: can you add some more e2e tests here for the test mode (failure cases, etc)

apps/dashboard/src/lib/utils.tsx
BilalG1: i think it is right since I want to check the result status

PR #967, opened 2025-10-20
unblock signup endpoint

BilalG1:
## Summary by CodeRabbit

* **New Features**
  * Sign-up accepts an optional verification callback URL and a new opt-out flag to disable email verification; when opted-out or absent, URL checks and verification emails are skipped.
  * Client APIs and runtime validation updated to forbid providing a callback URL when opting out. Sign-up now retries without a callback if a redirect URL is not whitelisted.

* **Tests**
  * End-to-end tests added for sign-up without verification and for conflicting verification settings.

### DIFF
--- a/packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
+++ b/packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
@@ -1862,22 +1862,22 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
     this._ensurePersistentTokenStore();
     const session = await this._getSession();
     const emailVerificationRedirectUrl = options.noVerificationCallback ? undefined : options.verificationCallbackUrl ?? constructRedirectUrl(this.urls.emailVerification, "verificationCallbackUrl");
-    
+
     let result = await this._interface.signUpWithCredential(
       options.email,
       options.password,
       emailVerificationRedirectUrl,
       session
     );
-    
+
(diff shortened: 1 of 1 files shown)


packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -1853,11 +1853,15 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
     verificationCallbackUrl?: string,
   }): Promise<Result<undefined, KnownErrors["UserWithEmailAlreadyExists"] | KnownErrors['PasswordRequirementsNotMet']>> {
+    if (options.noVerificationCallback && options.verificationCallbackUrl) {
+      throw new StackAssertionError("verificationCallbackUrl is not allowed when noVerificationCallback is true");
+    }
     this._ensurePersistentTokenStore();
     const session = await this._getSession();
-    const emailVerificationRedirectUrl = options.verificationCallbackUrl ?? constructRedirectUrl(this.urls.emailVerification, "verificationCallbackUrl");
+    const emailVerificationRedirectUrl = options.noVerificationCallback ? undefined : options.verificationCallbackUrl ?? constructRedirectUrl(this.urls.emailVerification, "verificationCallbackUrl");
     const result = await this._interface.signUpWithCredential(
```
N2D4: if the emailVerificationRedirectUrl (or its default) is not trusted, we should set it to undefined, that way you can still sign up with the default args even if the domain is not trusted (maybe we should log an error in the console though just to let the user know)

@claude

PR #969, opened 2025-10-22
capture freestyle error

BilalG1:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved email rendering reliability by detecting and handling cases where template executions produce no result, preventing silent failures.
  * Enhanced error reporting for both single and batched email generation so failures are captured and surfaced more consistently for faster diagnosis.

### DIFF
--- a/apps/backend/src/lib/email-rendering.tsx
+++ b/apps/backend/src/lib/email-rendering.tsx
@@ -125,13 +125,14 @@ export async function renderEmailWithTemplate(
     return Result.error(`${executeResult.error}`);
   }
   if (!executeResult.data.result) {
-    captureError("freestyle-no-result", new StackAssertionError("No result from Freestyle", {
+    const noResultError = new StackAssertionError("No result from Freestyle", {
       executeResult,
       templateOrDraftComponent,
       themeComponent,
       options,
-    }));
-    return Result.error("An unknown error occurred while rendering this email. Please try again");
+    });
+    captureError("freestyle-no-result", noResultError);
+    throw noResultError;
   }
   return Result.ok(executeResult.data.result as { html: string, text: string, subject: string, notificationCategory: string });
 }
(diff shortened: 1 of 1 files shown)


apps/backend/src/lib/email-rendering.tsx
```diff
@@ -120,11 +120,20 @@ export async function renderEmailWithTemplate(
...
   }
-  return Result.ok(output.data.result as { html: string, text: string, subject: string, notificationCategory: string });
+  if (!executeResult.data.result) {
+    captureError("freestyle-no-result", new StackAssertionError("No result from Freestyle", {
+      executeResult,
+      templateOrDraftComponent,
+      themeComponent,
+      options,
+    }));
+    return Result.error("An unknown error occurred while rendering this email. Please try again");
```
N2D4: this sounds like it should be a StackAssertionError instead? can we just throw the error above?

apps/backend/src/lib/email-rendering.tsx
```diff
@@ -205,6 +214,15 @@ export async function renderEmailsWithTemplateBatched(
...
     return Result.error(executeResult.error);
   }
+  if (!executeResult.data.result) {
+    captureError("freestyle-no-result", new StackAssertionError("No result from Freestyle", {
+      executeResult,
+      templateOrDraftComponent,
+      themeComponent,
+      inputs,
+    }));
+    return Result.error("An unknown error occurred while rendering this email. Please try again");
```
N2D4: as above

apps/backend/src/lib/email-rendering.tsx
```diff
@@ -205,6 +214,15 @@ export async function renderEmailsWithTemplateBatched(
...
+  }
```
N2D4, line 227: should this maybe be in the executeScript function, since it's duplicated code?

PR #971, opened 2025-10-22
Cookie subdomain sharing

BilalG1:
## Summary by CodeRabbit

- **New Features**
  - Projects now expose a domains field in the client API.
  - Cookie API expanded: domain and secure options added, plus getAll and isSecure helpers.

- **Refactor**
  - Domain-aware cookie and token handling for cross-domain refresh flows.
  - Minor signature/formatting tweaks to IP and URL utilities.

- **Tests**
  - E2E coverage added: refresh-cookie scenarios and a project scaffolding test.
  - Backend snapshot updated to include domains.

- **Chores**
  - Added a new dependency for domain parsing.

### DIFF
--- a/apps/backend/src/app/api/latest/auth/password/sign-in/route.tsx
+++ b/apps/backend/src/app/api/latest/auth/password/sign-in/route.tsx
@@ -20,7 +20,7 @@ export const POST = createSmartRouteHandler({
       tenancy: adaptSchema,
     }).defined(),
     body: yupObject({
-      email: emailSchema.defined(),
+      email: emailSchema.defined().nonEmpty(),
       password: passwordSchema.defined(),
     }).defined(),
   }),
--- a/apps/backend/src/app/api/latest/internal/send-test-webhook/route.tsx
+++ b/apps/backend/src/app/api/latest/internal/send-test-webhook/route.tsx
@@ -1,0 +1,107 @@
+import { getSvixClient } from "@/lib/webhooks";
+import { createSmartRouteHandler } from "@/route-handlers/smart-route-handler";
+import { adaptSchema, adminAuthTypeSchema, yupBoolean, yupNumber, yupObject, yupString } from "@stackframe/stack-shared/dist/schema-fields";
+import { captureError, StackAssertionError } from "@stackframe/stack-shared/dist/utils/errors";
+import { Result } from "@stackframe/stack-shared/dist/utils/results";
+import { MessageStatus } from "svix";
+
+export const POST = createSmartRouteHandler({
+  metadata: {
+    hidden: true,
+  },
+  request: yupObject({
+    auth: yupObject({
+      type: adminAuthTypeSchema,
+      tenancy: adaptSchema.defined(),
+    }).defined(),
+    body: yupObject({
+      endpoint_id: yupString().defined(),
+    }).defined(),
+    method: yupString().oneOf(["POST"]).defined(),
+  }),
+  response: yupObject({
+    statusCode: yupNumber().oneOf([200]).defined(),
+    bodyType: yupString().oneOf(["json"]).defined(),
+    body: yupObject({
+      success: yupBoolean().defined(),
+      error_message: yupString().optional(),
+    }).defined(),
+  }),
+  handler: async ({ auth, body }) => {
+    const projectId = auth.tenancy.project.id;
+    const svix = getSvixClient();
+
+    await svix.application.getOrCreate({ uid: projectId, name: projectId });
+    const endpointResult = await Result.fromPromise(svix.endpoint.get(projectId, body.endpoint_id));
+    if (endpointResult.status === "error") {
+      return {
+        statusCode: 200,
+        bodyType: "json",
+        body: {
+          success: false,
+          error_message: "Endpoint not found. Make sure it still exists.",
+        },
+      };
+    }
+    const endpoint = endpointResult.data;
+
+    const messageResult = await Result.fromPromise(svix.message.create(
+      projectId,
+      {
+        eventType: "stack.test",
+        payload: {
+          type: "stack.test",
+          data: {
+            message: "Stack webhook test event triggered from the Stack dashboard.",
+            endpointUrl: endpoint.url,
+          },
+        },
+      },
+    ));
+    if (messageResult.status === "error") {
+      const errorMessage = messageResult.error instanceof Error ? messageResult.error.message : "Unknown error while sending the test webhook.";
+      captureError("send-test-webhook", new StackAssertionError("Failed to send test webhook", {
+        cause: messageResult.error,
+        project_id: projectId,
+        endpoint_id: body.endpoint_id,
+      }));
+      return {
+        statusCode: 200,
+        bodyType: "json",
+        body: {
+          success: false,
+          error_message: errorMessage,
+        },
+      };
+    }
+
+    const attemptResult = await Result.retry(async () => {
+      const attempts = await svix.messageAttempt.listByMsg(
+        projectId,
+        messageResult.data.id,
+        { status: MessageStatus.Success }
+      );
+      const success = attempts.data.some(a => a.endpointId === body.endpoint_id);
+      return success ? Result.ok(undefined) : Result.error("No successful attempt found");
+    }, 3);
+
+    if (attemptResult.status === "error") {
+      return {
+        statusCode: 200,
+        bodyType: "json",
+        body: {
+          success: false,
+          error_message: "Webhook not delivered. Make sure the endpoint is configured correctly.",
+        },
+      };
+    }
+
+    return {
+      statusCode: 200,
+      bodyType: "json",
+      body: {
+        success: true,
+      },
+    };
+  },
+});
--- a/apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/webhooks/page-client.tsx
+++ b/apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/webhooks/page-client.tsx
@@ -1,19 +1,22 @@
 "use client";
 
-import { FormDialog, SmartFormDialog } from "@/components/form-dialog";
+import { SmartFormDialog } from "@/components/form-dialog";
 import { InputField } from "@/components/form-fields";
 import { useRouter } from "@/components/router";
 import { SettingCard } from "@/components/settings";
 import { getPublicEnvVar } from '@/lib/env';
 import { urlSchema } from "@stackframe/stack-shared/dist/schema-fields";
-import { ActionCell, ActionDialog, Alert, Button, Table, TableBody, TableCell, TableHead, TableHeader, TableRow, Typography } from "@stackframe/stack-ui";
-import { useState } from "react";
+import { yupResolver } from "@hookform/resolvers/yup";
+import { ActionCell, ActionDialog, Alert, AlertDescription, AlertTitle, Button, Form, Table, TableBody, TableCell, TableHead, TableHeader, TableRow, Typography } from "@stackframe/stack-ui";
+import { useMemo, useState } from "react";
+import { useForm } from "react-hook-form";
 import { SvixProvider, useEndpoints, useSvix } from "svix-react";
 import * as yup from "yup";
 import { AppEnabledGuard } from "../app-enabled-guard";
 import { PageLayout } from "../page-layout";
 import { useAdminApp } from "../use-admin-app";
 import { getSvixResult } from "./utils";
+import { runAsynchronously } from "@stackframe/stack-shared/dist/utils/promises";
 
 type Endpoint = {
   id: string,
@@ -24,46 +27,134 @@ type Endpoint = {
 function CreateDialog(props: {
   trigger: React.ReactNode,
   updateFn: () => void,
+  onTestRequested?: (endpoint: Endpoint) => void,
 }) {
   const { svix, appId } = useSvix();
+  const [open, setOpen] = useState(false);
+  const [createdEndpoint, setCreatedEndpoint] = useState<Endpoint | null>(null);
 
   const formSchema = yup.object({
     url: urlSchema.defined().label("URL"),
     description: yup.string().label("Description"),
   });
 
-  return <FormDialog
-    trigger={props.trigger}
-    title={"Create new endpoint"}
-    formSchema={formSchema}
-    okButton={{ label: "Create" }}
-    onSubmit={async (values) => {
-      await svix.endpoint.create(appId, { url: values.url, description: values.description });
+  const form = useForm<yup.InferType<typeof formSchema>>({
+    resolver: yupResolver(formSchema),
+    defaultValues: {
+      url: "",
+      description: "",
+    },
+    mode: "onChange",
+  });
+
+  const handleOpenChange = (value: boolean) => {
+    setOpen(value);
+    if (!value) {
       props.updateFn();
-    }}
-    render={(form) => (
-      <>
-        <Alert>
-          Make sure this is a trusted URL that you control.
-        </Alert>
-        <InputField
-          label="URL"
-          name="url"
-          control={form.control}
-        />
-        <InputField
-          label="Description"
-          name="description"
-          control={form.control}
-        />
-        {(form.watch('url') as any)?.startsWith('http://') && (
-          <Alert variant="destructive">
-            Using HTTP endpoints is insecure. This can expose your user data to attackers. Only use HTTP endpoints in development environments.
+    }
+    form.reset();
+    setCreatedEndpoint(null);
+  };
+
+  const handleSubmit = form.handleSubmit(async (values) => {
+    const created = await svix.endpoint.create(appId, { url: values.url, description: values.description });
+    setCreatedEndpoint({
+      id: created.id,
+      url: created.url,
+      description: created.description,
+    });
+  });
+
+  const allowInsecureWarning = form.watch("url").startsWith("http://");
+
+  return (
+    <ActionDialog
+      trigger={props.trigger}
+      title={"Create new endpoint"}
+      open={open}
+      onOpenChange={handleOpenChange}
+      okButton={false}
+      cancelButton={false}
+    >
+      {createdEndpoint ? (
+        <div className="space-y-6">
+          <Alert variant="success">
+            <AlertTitle>Endpoint created</AlertTitle>
+            <AlertDescription>
+              <span className="block">
+                <span className="font-semibold">URL:</span> {createdEndpoint.url}
+              </span>
+              {createdEndpoint.description ? (
+                <span className="block">
+                  <span className="font-semibold">Description:</span> {createdEndpoint.description}
+                </span>
+              ) : null}
+              <span className="block pt-2">
+                You can now send a test event to verify your integration.
+              </span>
+            </AlertDescription>
           </Alert>
-        )}
-      </>
-    )}
-  />;
+          <div className="flex justify-end gap-2">
+            <Button
+              variant="outline"
+              onClick={() => handleOpenChange(false)}
+            >
+              Close
+            </Button>
+            <Button
+              onClick={() => {
+                props.onTestRequested?.(createdEndpoint);
+                handleOpenChange(false);
+              }}
+            >
+              Test endpoint
+            </Button>
+          </div>
+        </div>
+      ) : (
+        <Form {...form}>
+          <form
+            onSubmit={(e) => runAsynchronously(handleSubmit(e))}
+            className="space-y-4"
+          >
+            <Alert>
+              Make sure this is a trusted URL that you control.
+            </Alert>
+            <InputField
+              label="URL"
+              name="url"
+              control={form.control}
+            />
+            <InputField
+              label="Description"
+              name="description"
+              control={form.control}
+            />
+            {allowInsecureWarning && (
+              <Alert variant="destructive">
+                Using HTTP endpoints is insecure. This can expose your user data to attackers. Only use HTTP endpoints in development environments.
+              </Alert>
+            )}
+            <div className="flex justify-end gap-2 pt-2">
+              <Button
+                variant="outline"
+                type="button"
+                onClick={() => handleOpenChange(false)}
+              >
+                Cancel
+              </Button>
+              <Button
+                type="submit"
+                disabled={!form.formState.isValid}
+              >
+                Create
+              </Button>
+            </div>
+          </form>
+        </Form>
+      )}
+    </ActionDialog>
+  );
 }
 
 export function EndpointEditDialog(props: {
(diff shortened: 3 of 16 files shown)


packages/stack-shared/src/utils/urls.tsx
```diff
@@ -225,6 +226,31 @@ import.meta.vitest?.test("isLocalhost", ({ expect }) => {
...
+}
+import.meta.vitest?.test("extractBaseDomainFromHost", ({ expect }) => {
+  expect(extractBaseDomainFromHost("app.example.com")).toBe("example.com");
+  expect(extractBaseDomainFromHost("sub.app.example.com")).toBe("example.com");
+  expect(extractBaseDomainFromHost("example.com")).toBe("example.com");
+  expect(extractBaseDomainFromHost("localhost:3000")).toBe("localhost");
+  expect(extractBaseDomainFromHost("127.0.0.1")).toBe("127.0.0.1");
+  expect(extractBaseDomainFromHost("127.0.0.1:3000")).toBe("127.0.0.1");
+  expect(extractBaseDomainFromHost("app.example.co.uk")).toBe("example.co.uk");
+});
```
N2D4: i think we won't need this function, and given the security issues, I'd remove it

```suggestion
```

N2D4: i think we won't need this function, and given the security issues, I'd remove it

```suggestion
```

packages/template/src/lib/cookie.ts
```diff
@@ -153,6 +170,23 @@ export async function getCookie(name: string): Promise<string | null> {
...
+  const headers = await rscHeaders();
+  if (headers.get("x-forwarded-proto") === "https") {
+    return true;
+  }
+  if (cookies.get("stack-is-https")) {
+    return true;
+  }
+  // END_PLATFORM
+  return false;
+}
```
N2D4, line 210: this logic is duplicated from the set function. can we dedupe this somehow?

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -418,13 +419,98 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+      if (!refreshToken) {
+        return null;
+      }
+      return {
+        refreshToken,
+        updatedAt,
+      };
+    } catch {
+      return null;
+    }
```
N2D4: generic try-catch is bad. use the parseJson(...) function instead

packages/template/src/lib/cookie.ts
```diff
@@ -94,6 +99,13 @@ function createNextCookieHelper(
       }
       return rscCookiesAwaited.get(name)?.value ?? null;
     },
+    getAll: () => {
+      const all = rscCookiesAwaited.getAll();
+      return all.reduce((acc, entry) => {
+        acc[entry.name] = entry.value;
+        return acc;
+      }, {} as Record<string, string>);
+    },
```
N2D4, line 107: <p class='graphite__hidden'><i>[Re: lines +82 to +108]</i></p>
can we make `get` use `getAll` to dedupe the code? and can we set `stack-is-https` in `getAll`?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/stack-auth/stack-auth/971?utm_source=unchanged-line-comment">Graphite</a>.</p>

packages/template/src/lib/cookie.ts
```diff
@@ -51,6 +53,9 @@ export async function createCookieHelper(): Promise<CookieHelper> {
 export function createBrowserCookieHelper(): CookieHelper {
   return {
     get: getCookieClient,
+    getAll: () => {
+      return Cookies.get();
+    },
```
N2D4: just like in the nextjs cookie helper, can we make getCookieClient use a new function getAllCookiesClient?

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -418,13 +419,98 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+          parsed.length === 2 &&
+          typeof parsed[0] === "string" &&
+          typeof parsed[1] === "string" &&
+          parsed[0] === refreshToken
+        ) {
+          accessToken = parsed[1];
+        }
+      } catch {
+        // ignore invalid cookie contents
+      }
```
N2D4: as above; never use a catch-all. use parseJson instead

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -418,13 +419,98 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+    return `${secure ? "__Host-" : ""}${this._refreshTokenCookieName}--default`;
+  }
+  private _getCustomRefreshCookieName(domain: string): string {
+    const encoded = encodeBase32(new TextEncoder().encode(domain.toLowerCase()));
+    return `${this._refreshTokenCookieName}--custom-${encoded}`;
+  }
+  private _formatRefreshCookieValue(refreshToken: string, updatedAt: number): string {
+    return JSON.stringify({
+      refresh_token: refreshToken,
+      updated_at: updatedAt,
```
N2D4: kinda late, I know, but maybe we can rename this to `updated_at_millis` to be consistent with the endpoints?

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
     // the access token on page reload.
     return `stack-access`;
   }
+  private _getAllBrowserCookies(): Record<string, string> {
+    if (!isBrowserLike()) {
+      return {};
```
N2D4: can we make this throw an error instead?

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+    }
+    return cookie.parse(document.cookie || "");
+  }
+  private _getBrowserBaseDomain(): string | undefined {
+    if (!isBrowserLike()) {
+      return undefined;
+    }
+    const host = window.location.host;
+    return host ? extractBaseDomainFromHost(host) : undefined;
+  }
```
N2D4: we can remove this now
```suggestion
```

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+    return undefined;
+  }
+  private _deleteLegacyBrowserRefreshCookies() {
+    deleteCookieClient(this._refreshTokenCookieName);
+    deleteCookieClient("stack-refresh");
+    const domain = this._getBrowserBaseDomain();
+    if (domain) {
+      deleteCookieClient(this._refreshTokenCookieName, { domain });
+      deleteCookieClient("stack-refresh", { domain });
+    }
```
N2D4: why is this second half necessary? i thought we never set the domain previously

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+      const hostHeader = resolvedHeaders?.get("x-forwarded-host") ?? resolvedHeaders?.get("host");
+      return hostHeader ? extractBaseDomainFromHost(hostHeader) : undefined;
+    } catch {
+      return undefined;
+    }
+    // END_PLATFORM
+    return undefined;
+  }
+  private _deleteLegacyBrowserRefreshCookies() {
+    deleteCookieClient(this._refreshTokenCookieName);
```
N2D4:
```suggestion
    deleteCookieClient(this._legacyRefreshTokenCookieName);
```

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+      setOrDeleteCookie(this._refreshTokenCookieName, null, { noOpIfServerComponent: true }),
+      setOrDeleteCookie("stack-refresh", null, { noOpIfServerComponent: true }),
+    ];
+    const domain = await this._getServerBaseDomain();
+    if (domain) {
+      operations.push(
+        setOrDeleteCookie(this._refreshTokenCookieName, null, { domain, noOpIfServerComponent: true }),
+        setOrDeleteCookie("stack-refresh", null, { domain, noOpIfServerComponent: true }),
+      );
+    }
```
N2D4: as above, why is this necessary?

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+    await Promise.all(operations);
```
N2D4: i think we can just await the statements above, that'll make the code cleaner


```suggestion
    await setOrDeleteCookie(this._refreshTokenCookieName, null, { noOpIfServerComponent: true });
    await setOrDeleteCookie("stack-refresh", null, { noOpIfServerComponent: true });
    
    const domain = await this._getServerBaseDomain();
    if (domain) {
      await setOrDeleteCookie(this._refreshTokenCookieName, null, { domain, noOpIfServerComponent: true });
      await setOrDeleteCookie("stack-refresh", null, { domain, noOpIfServerComponent: true });
    }
```

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+        : setOrDeleteCookie;
+
+      const deleteCookie = async (targetDomain: string) => {
+        const name = this._getCustomRefreshCookieName(targetDomain);
+        await setOrDeleteCookieWrapper(name, null, { domain: targetDomain, noOpIfServerComponent: true });
+      };
+      const setCookie = async (targetDomain: string, value: string) => {
+        const name = this._getCustomRefreshCookieName(targetDomain);
+        await setOrDeleteCookieWrapper(name, value, { maxAge: 60 * 60 * 24 * 365, domain: targetDomain, noOpIfServerComponent: true });
+      };
```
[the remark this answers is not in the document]
N2D4: good call, maybe setCookieClient should set Secure automatically?

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+        setOrDeleteCookie(this._refreshTokenCookieName, null, { domain, noOpIfServerComponent: true }),
+        setOrDeleteCookie("stack-refresh", null, { domain, noOpIfServerComponent: true }),
+      );
+    }
+    await Promise.all(operations);
+  }
+  private _queueCustomRefreshCookieUpdate(refreshToken: string | null, updatedAt: number | null, context: "browser" | "server") {
+    runAsynchronously(async () => {
+      const domain = await this._fetchTrustedParentDomain();
+      const previousDomain = this._lastCustomRefreshCookieDomain;
```
N2D4: i think we can simplify the logic here a bit. we don't have to store previousDomain, for `set` we can just update the `domain` (it will take precedence since it has the highest updated_at all other) and for `delete` we can delete all the cookies (for all the domains)

N2D4: the `delete` part can happen inside the refresh token onChange, it doesn't need to be in the queued one, since it doesn't require fetching the trusted domains

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+        if (previousDomain) {
+          await deleteCookie(previousDomain);
+          this._lastCustomRefreshCookieDomain = null;
+        }
+        return;
+      }
+
+      if (previousDomain && previousDomain !== domain) {
+        await deleteCookie(previousDomain);
+      }
```
N2D4: as above

```suggestion
```

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+
+      if (refreshToken && updatedAt !== null) {
+        if (this._lastCustomRefreshCookieUpdatedAt !== null && updatedAt < this._lastCustomRefreshCookieUpdatedAt) {
```
N2D4: instead of making this a timestamp, I'd make this an index; that's more resilient in case cookies are set within the same timestamp (which is quite realistic)


```suggestion
        if (updateIndex !== this._mostRecentQueuedCookieRefreshIndex) {
```

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+        }
+        const value = this._formatRefreshCookieValue(refreshToken, updatedAt);
+        await setCookie(domain, value);
+        this._lastCustomRefreshCookieDomain = domain;
+        this._lastCustomRefreshCookieUpdatedAt = updatedAt;
+      } else {
+        await deleteCookie(domain);
+        this._lastCustomRefreshCookieDomain = null;
+        this._lastCustomRefreshCookieUpdatedAt = null;
+      }
```
N2D4: if we get rid of the delete in here that cleans up a lot:


```suggestion
        if (this._lastCustomRefreshCookieUpdatedAt !== null && updatedAt < this._lastCustomRefreshCookieUpdatedAt) {
          // Ignore stale writes
          return;
        }
        const value = this._formatRefreshCookieValue(refreshToken, updatedAt);
        await setCookie(domain, value);
        this._lastCustomRefreshCookieDomain = domain;
        this._lastCustomRefreshCookieUpdatedAt = updatedAt;
```

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+    });
+  }
+  private async _fetchTrustedParentDomain(): Promise<string | null> {
```
N2D4: we should cache the return value of this function somewhere, otherwise we will do a lot of requests


```suggestion
  private async _getTrustedParentDomain(): Promise<string | null> {
```

N2D4: this function needs to take the current domain as input somewhere

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -437,17 +523,137 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+      const bParts = b.split(".").length;
+      if (aParts !== bParts) {
+        return aParts - bParts;
+      }
+      return stringCompare(a, b);
+    });
+    const selected = candidates[0] ?? null;
+    return selected;
+
+  }
```
N2D4: i think a simpler implementation of this function takes the current domain as an argument, and then iterates by going up one level in the domain each; for each parent domain, it checks whether both the parent domain itself *and* `**.parent-domain` are trusted domains




```suggestion
  private async _fetchTrustedParentDomain(currentDomain?: string): Promise<string | null> {
    const project = Result.orThrow(await this._interface.getClientProject());
    const domains = project.config.domains.map(d => d.domain.trim().replace(/^https?:\/\//, "").split("/")[0]?.toLowerCase();

    const parts = currentDomain.split('.');
    for (let i = part.length - 2; i >= 0; i--) {
      const parentDomain = parts.slice(i).join('.');
      if (domains.includes(parentDomain) && trustedWildcards.includes("**." + parentDomain)) {
        return parentDomain;
      }
    }
    
    return null;
  }
```

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -467,9 +673,16 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+          const refreshToken = value.refreshToken;
+          const updatedAt = refreshToken ? Date.now() : null;
+          const refreshCookieValue = refreshToken && updatedAt !== null ? this._formatRefreshCookieValue(refreshToken, updatedAt) : null;
+          const secure = window.location.protocol === "https:";
+          const defaultName = this._getRefreshTokenDefaultCookieNameForSecure(secure);
+          setOrDeleteCookieClient(defaultName, refreshCookieValue, { maxAge: 60 * 60 * 24 * 365, secure });
+          const accessTokenPayload = refreshToken && value.accessToken ? JSON.stringify([refreshToken, value.accessToken]) : null;
+          setOrDeleteCookieClient(this._accessTokenCookieName, accessTokenPayload, { maxAge: 60 * 60 * 24 });
+          this._deleteLegacyBrowserRefreshCookies();
+          this._queueCustomRefreshCookieUpdate(refreshToken, updatedAt, "browser");
```
N2D4, line 671: as above, this logic is only necessary for setting cookies. for deleting cookies, we can just delete all the refresh-token-related cookies (we should probably have a function that returns the list of ALL potential cookie names, and use that same list for the get and the delete functions)

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -495,28 +708,21 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
-              //
-              // so hack it is
+              const refreshToken = value.refreshToken;
+              const updatedAt = refreshToken ? Date.now() : null;
+              const refreshCookieValue = refreshToken ? this._formatRefreshCookieValue(refreshToken, updatedAt!) : null;
+              const secure = await isSecureCookieContext();
+              const defaultName = this._getRefreshTokenDefaultCookieNameForSecure(secure);
               await Promise.all([
-                setOrDeleteCookie(this._refreshTokenCookieName, value.refreshToken, { maxAge: 60 * 60 * 24 * 365, noOpIfServerComponent: true }),
-                setOrDeleteCookie(this._accessTokenCookieName, value.accessToken ? JSON.stringify([value.refreshToken, value.accessToken]) : null, { maxAge: 60 * 60 * 24, noOpIfServerComponent: true }),
```
N2D4, line 518: why did you delete the comment here?

packages/template/src/lib/cookie.ts
```diff
@@ -142,23 +155,70 @@ function createNextCookieHelper(
...
+// IF_PLATFORM next
+function determineSecureFromServerContext(
+  cookies: Awaited<ReturnType<typeof rscCookies>>,
+  headers: Awaited<ReturnType<typeof rscHeaders>>,
+): boolean {
+  let isSecureCookie = !!cookies.get("stack-is-https");
+  if (headers.get("x-forwarded-proto") === "https") {
+    isSecureCookie = true;
+  }
+  return isSecureCookie;
```
N2D4: always a good idea to compress code

```suggestion
  return cookies.has("stack-is-https") || headers.get("x-forwarded-proto") === "https";
```

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -418,13 +424,101 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
   protected get _refreshTokenCookieName() {
     return `stack-refresh-${this.projectId}`;
   }
-  protected _getTokensFromCookies(cookies: { refreshTokenCookie: string | null, accessTokenCookie: string | null }): TokenObject {
-    const refreshToken = cookies.refreshTokenCookie;
-    const accessTokenObject = cookies.accessTokenCookie?.startsWith('[\"') ? JSON.parse(cookies.accessTokenCookie) : null;  // gotta check for validity first for backwards-compat, and also in case someone messes with the cookie value
-    const accessToken = accessTokenObject && refreshToken === accessTokenObject[0] ? accessTokenObject[1] : null;  // if the refresh token has changed, the access token is invalid
+  protected get _refreshTokenDefaultCookieName() {
+    return `${this._refreshTokenCookieName}--default`;
+  }
```
N2D4: unused?

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -418,13 +424,101 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+  private _formatAccessCookieValue(refreshToken: string | null, accessToken: string | null): string | null {
+    return refreshToken && accessToken ? JSON.stringify([refreshToken, accessToken]) : null;
+  }
+  private _parseStructuredRefreshCookie(value: string | null): { refreshToken: string, updatedAt: number | null } | null {
+    if (!value) {
+      return null;
+    }
+    const parsed = parseJson(value);
+    if (parsed.status !== "ok" || typeof parsed.data !== "object" || parsed.data === null) {
+      return null;
```
N2D4, line 457: can we add a console.warn here with the error?

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -418,13 +424,101 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+    }
+    const data = parsed.data;
+    const refreshToken = "refresh_token" in data && typeof data.refresh_token === "string" ? data.refresh_token : null;
+    const updatedAt = "updated_at_millis" in data && typeof data.updated_at_millis === "number" ? data.updated_at_millis : null;
+    if (!refreshToken) {
+      return null;
```
N2D4, line 464: also console.warn here right? afaict this could never happen if nothing is broken, correct?

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -418,13 +424,101 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+        typeof parsed.data === "object" &&
+        parsed.data !== null &&
+        Array.isArray(parsed.data) &&
+        parsed.data.length === 2 &&
+        typeof parsed.data[0] === "string" &&
+        typeof parsed.data[1] === "string" &&
+        parsed.data[0] === refreshToken
+      ) {
+        accessToken = parsed.data[1];
+      }
```
N2D4, line 522: can we also have a console.warn here if the access token cookie format is incorrect? (some of these are incorrect formats that should never happen, others can actually happen in normal execution, so we have to distinguish between those two and only console.warn in the ones that should never happen)

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
N2D4: the reason why i would like to have console.warn everywhere is because it will help us debug if we ever have cookie issues, as well as our users who try to set refresh and access tokens manually. (there are quite a lot of these, eg. for electron etc)

N2D4 (approved):

PR #978, opened 2025-10-23
Init stack telegram alerts

BilalG1:
## Summary by CodeRabbit

* **New Features**
  * Stack initialization now sends completion callbacks (success/failure) to a configured notification service.
  * Backend endpoint added to receive and forward initialization completion notifications.

* **Chores**
  * Added configuration entries for Telegram callback integration.
  * Local init tooling updated to respect a configurable API base URL for callback delivery.


apps/backend/.env
```diff
@@ -83,3 +83,5 @@ STACK_OPENAI_API_KEY=# enter your openai api key
 STACK_FEATUREBASE_API_KEY=# enter your featurebase api key
 STACK_STRIPE_SECRET_KEY=# enter your stripe api key
 STACK_STRIPE_WEBHOOK_SECRET=# enter your stripe webhook secret
+STACK_INIT_TELEGRAM_BOT_TOKEN= # enter you telegram bot token
+STACK_INIT_TELEGRAM_CHAT_ID=# enter your telegram chat id
```
BilalG1: reminder: need to set these before merging

N2D4: what does the INIT stand for? why not just STACK_TELEGRAM_BOT_TOKEN?

apps/backend/src/app/api/latest/integrations/telegram/route.tsx
```diff
@@ -0,0 +1,54 @@
...
+    body: yupObject({
+      success: yupBoolean().oneOf([true]).defined(),
+    }).defined(),
+  }),
+  async handler({ body: { message } }) {
+    const botToken = getEnvVariable("STACK_INIT_TELEGRAM_BOT_TOKEN", "");
+    const chatId = getEnvVariable("STACK_INIT_TELEGRAM_CHAT_ID", "");
+
+    if (!botToken || !chatId) {
+      throw new Error("Telegram integration is not configured.");
```
N2D4: this should be a StackAssertionError

apps/backend/.env
N2D4: ahh because of init script? I think let's keep it without the INIT, we can add more TG bot actions later in that case

apps/backend/src/app/api/latest/integrations/telegram/route.tsx
N2D4, line 1: feels like a better name for this would be `api/latest/internal/init-script-callback`?

packages/init-stack/src/index.ts
```diff
@@ -354,6 +355,16 @@ async function main(): Promise<void> {
...
 
+  await notifyTelegramOnCompletion({
+    success: true,
+    distinctId,
+    options,
+    args: program.args,
+    isNonInteractive: isNonInteractiveEnv(),
+    timestamp: new Date().toISOString(),
+    projectPath,
+  });
```
N2D4: like the endpoint name change, maybe a better name for this function would be `invokeCallback`

N2D4: like the endpoint name change, maybe a better name for this function would be `invokeCallback`

packages/init-stack/src/telegram.ts
```diff
@@ -0,0 +1,76 @@
...
+    `Args: ${argSummary}`,
+    `Error: ${errorSummary}`,
+  ].filter((line): line is string => Boolean(line));
+
+  if (error?.stack) {
+    lines.push(`Stack: ${truncate(error.stack, 4000)}`);
+  }
+
+  return "_".repeat(50) + "\n\n" + lines.join("\n");
+}
```
N2D4: this should probably be in the endpoint, not here (otherwise clients can send whatever they want into our telegram chat)

N2D4 (changes requested):
some comments!

PR #983, opened 2025-10-27
fix team invitations with server actions

BilalG1:
## Summary by CodeRabbit

- **New Features**
    - Invite users to teams by email with customizable callback URLs.
    - View and revoke pending invitations from the team management UI.
    - Track and enforce team seat capacity, disabling invites when full.
- **Improvements**
    - Upgrade flow now redirects to the checkout URL from the team UI.

### DIFF
--- a/apps/dashboard/src/app/(main)/(protected)/(outside-dashboard)/projects/page-client.tsx
+++ b/apps/dashboard/src/app/(main)/(protected)/(outside-dashboard)/projects/page-client.tsx
@@ -3,16 +3,16 @@
 import { ProjectCard } from "@/components/project-card";
 import { useRouter } from "@/components/router";
 import { SearchBar } from "@/components/search-bar";
-import { AdminOwnedProject, StackAdminApp, Team, useUser } from "@stackframe/stack";
+import { AdminOwnedProject, Team, useUser } from "@stackframe/stack";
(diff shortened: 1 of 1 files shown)


apps/dashboard/src/app/(main)/(protected)/(outside-dashboard)/projects/page-client.tsx
```diff
@@ -159,39 +156,42 @@ function TeamAddUserDialog(props: {
...
+  const [invitations, setInvitations] = useState<Awaited<ReturnType<typeof listInvitations>>>();
+
+  const fetchInvitations = useCallback(async () => {
+    const invitations = await listInvitations(props.team.id);
+    setInvitations(invitations);
+  }, [props.team.id]);
+
+  useEffect(() => {
+    runAsynchronously(fetchInvitations());
+  }, [fetchInvitations]);
```
N2D4: async useEffect that modifies state should always be cancelable:


```suggestion
  useEffect(() => {
    let canceled = false;
    runAsynchronously(async () => {
      const invitations = await listInvitations(props.team.id);
      if (!canceled) {
        setInvitations(invitations);
      }
    }
    return () => { canceled = true };
  }, [fetchInvitations]);
```

PR #1030, opened 2025-11-26
Email outbox backend

N2D4:
I'm sorry for whoever reviews this. I tried to split it up as much as possible so this is only the existing functionality migrated onto the new email outbox table; the new endpoints or UI changes are not here yet.

Tests not passing yet but already ready for reviews.

How to review:

- The most important file is schema.prisma. It contains the explanation of the new EmailOutbox table, and all its constraints.
- After that, check the updated emails.tsx. It writes to the EmailOutbox table.
- Then, check email-queue-step.tsx. It contains the logic that renders, queues, and sends an email (alongside emails-low-level.tsx). This is called once every ~second (in a Vercel cron job in prod or a script in dev) and whenever a new email is sent (see the callers of runEmailQueueStep for more info).
- Then, look at the updated endpoints. Would love to have some eyes on the ones that @BilalG1 wrote in particular.
- Finally, everything else. Everything else should be relatively straight-forward, I think, mostly migrations.

The prompt that generated the first version (but note that the final version is somewhat different from what was generated, but it might help get some overview):

```
i want you to implement a new email sending backend infrastructure. apps/backend is the right folder

let's start with the database. For this purpose, I've already created an EmailOutbox model in the prisma schema. Note that, for now, canHaveDeliveryInfo is always false (we never fetch delivery info rn). Note that the EmailOutbox is on the global database, not the source-of-truth database (which means it's not in the same database as the project users, for example, and can't have foreign key relations to those). You must create a new migration for this new schema; you will have to manually write the migration SQL for it (can't use the Prisma command). After that, create another subsequent migration that migrates from the old SentEmail table to the new one.

Create a file in `lib` that has functions for the email outbox for listing all emails (returning the database objects), editing an email, resetting an email, deleting an email (which should only work under the same conditions as the retry one). This file should be SUPER clean, make sure to have many explained/documented functions with concise docs.

You also want to update emails.tsx to use the new email outbox. The usual current email-sending functions should instead of sending the email directly, create an EmailOutbox row in the database instead. It should also queue the email-queue-step below onto QStash so that it can start queuing itself repeatedly (this should probably be exported as a function so. Keep the existing function implementations somewhere for your own reference until you've implemented the email queue step endpoint down below, as you'll need to implement the email queue step endpoint down below with them. Note that if you send from a templtae to many users, each user is its own row in the email outbox system. Feel free to update the signature of the send email functions to match the information that is required in EmailOutbox, just update all the callers.

Also, update the `email-rendering.tsx` file to export a new function that allows you to render many email templates/drafts/themes at once, as long as they are for the same tenancy ID. They should all be in a single call to Freestyle.

Note that difference between InternalError and ExternalError in the EmailOutbox schema. Internal errors contain all the information, but maybe have sensitive implementaiton-speciifc data that should not exposed to the end user who sent the email. External errors are guaranteed to be safe to expose to the end user who sent the email. There is already a similar handling that's implicit in the emails.tsx file, so take some of its logic and make it more explicit (for both rendering & sending server errors, make functions that take internal full error objects and returns an external error object, with just a string like "Internal server error" if it's an unknown error). 

Also, create the following endpoints (please look at other endpoints in the backend to see how they're implemented, and make sure the security is good):

- GET /api/latest/emails/delivery-info: This endpoint returns some stats on the emails. It should return the number of emails sent in the last hour, last day, last week, and last month, and how many of those emails (per hour, day, week, and month) were marked as spam or bounced. It should also return the "capacity rate": the number of emails that can be sent per second. It should be calculated like this: `max(max(<the number of emails sent in the last week> * 4, 200) * penaltyFactor, 5) / 7 / 60 / 60`, where penaltyFactor is a number between 0 and 1 depending on what % of emails were bounced or marked as spam (I will build a precise metric for this later, for now you can think about it and make your own).
- GET /api/latest/emails/outbox: List all emails in the outbox, paginated like the users endpoint. Should return a nice type for the individual rows which should at least contain the status and simple status, and probably some other useful fields that you think should be displayed to the user in the UI, but not all those from the database (if you're not sure whether it's needed, don't include it, we can include it later). The function which converts DB EmailOutbox objects to the API type should be separate from the other stuff as we will use it elsewhere too.
- PATCH /api/latest/emails/outbox/{id}: Edit an email in the outbox, if the conditions for it are met. Only some fields can be edited, see the Prisma schema for an explanation
- POST /api/latest/emails/outbox/{id}/cancel: Cancel an email in the outbox, if the conditions for it are met
- POST /api/latest/emails/outbox/{id}/retry: Retry an email in the outbox, if the conditions for it are met. See the Prisma schema for an explanation
- POST /api/latest/internal/email-queue-step: This is a QStash receiver that is constantly called by QStash in an infinite loop. Make sure that it is super clean and easy to understand. Feel free to split it up into multiple functions and files as needed.
- - First thing it should do is add a call to itself back into the QStash queue, with Flow Control with a flow control key equal to the URL, a rate limit of 1 per second, and a parallelism limit of 1.
- - Then, it should acquire a Postgres exclusive lock, to make sure that only one email-queue-step can be running at a time (make sure this is easy to understand and free of data races). It should be released automatically at the end of this endpoint (think about how we can make it that there's no deadlock if the endpoint fails for whatever reason, eg. the server it's running on dies).
- - Now, it should get the timestamp of the last execution of this endpoint (which is stored in some metadata table in the global database), and calculate the precise time delta in seconds. The new time should be updated in the metadata table. (Make sure that it's not negative, which can happen due to clock skew, so make sure to handle that by just not doing anything if the new time is before the last execution time).
- - Then, it should fetch all the email outbox rows that do not have a renderedByWorkerId, and set renderedByWorkerId to a random UUID and startedRenderingAt to now (it should do this in an atomic statement so that an email is never rendered by multiple workers at the same time). It should then use email-rendering.tsx to render the email, using the previously created batch-by-tenancy function (you need to group the email outbox rows by tenancy ID for this). Once done, it should (again, in a single atomic statement) check whether the renderedByWorkerId is still set to the same UUID (this is to make sure there are no race conditions), and if so, set the appropriate fields in the EmailOutbox row.
- - Next, it should run a SQL query that updates all rows with scheduledAt < now() and isQueued = false to set isQueued = true.
- - Now, it should fetch a deduplicated list of all tenancy IDs of all email outbox rows that have isQueued = true and startedSendingAt being null, and set startedSendingAt to now.
- - For each tenancy ID in the list:
- - - Calculate the capacity rate. Multiply it by the time delta since the last endpoint execution. Use stochastic rounding to round to the nearest integer.
- - - For that tenancy, fetch at most that many email outbox rows that have isQueued = true and startedSendingAt being null, and set startedSendingAt to now. Fetch those rows that have the highest priority (see the Prisma schema for an explanation).
- - - Then, asynchronously in parallel, process each email that needs to be sent. Use the email-sending functions to send the emails, and update finishedSendingAt and canHaveDeliveryInfo (recall it's always set to false for now) for each row when the email is sent. Before you send the email, check whether the user still exists in the DB, and use the same notification category logic from the send-email endpoint (extract it into a helper function) to find out which users have already unsubscribed and should not receive the email, and set skippedReason accordingly if necessary. If the email encountered an error while sending, set the sendServerError-ish fields in the same atomic statement. `waitUntil` each promise (to make sure that they continue running in the background). Ensure that a bug when handling one of the emails does not mess with the other emails. Don't use `.catch`, ever, use try-catch blocks, and use `captureError("<some-machine-readable-location-id>", ...)` to capture all unknown errors aggressively.


While you're at it, we now have some new features like scheduling, add those to the send-email endpoint.

Make sure to make VERY comprehensive E2E tests for ALL the edge cases.

The code should be ULTRA clean and NEVER repetitive, feel free to create helper functions and files as needed. Definitely make sure everything is clean and easy to understand. When there's some logic that is used in multiple places, create a helper function for it. In the end, write a human-readable README.md file in the email endpoint folder that explains how the email infrastructure works, meant to be read by a technical human (it should be VERY concise and understandable!)

```

---

> [!NOTE]
> Migrates email to an async outbox pipeline with queueing/rendering, delivery stats, updated APIs, and extensive tests.
> 
> - **Database/Migrations**:
>   - Add `EmailOutbox` schema with generated columns, constraints, and indices; migrate data from `SentEmail` and drop it.
> - **Backend/Logic**:
>   - Implement outbox-driven flow: enqueue via `sendEmailToMany`, batch render (Freestyle), queue, and send (low-level SMTP/provider) with robust error handling and skip reasons.
>   - Add periodic processor: `runEmailQueueStep` and internal endpoint `/api/latest/internal/email-queue-step` with locking, delta-based capacity, and stochastic quotas.
>   - Extract low-level email sender, batched renderer, and delivery stats calculator.
> - **APIs**:
>   - Update `POST /api/latest/emails/send-email` (per-user enqueue, variables JSON, priority/scheduling, notification category validation).
>   - Add `GET /api/latest/emails/delivery-info` (hour/day/week/month stats and capacity).
>   - Switch internal email listing to outbox-backed data shape.
> - **SDK/Template**:
>   - Expose `getEmailDeliveryStats` and `useEmailDeliveryStats`; server interface method added.
> - **Tests**:
>   - Comprehensive E2E for queue step, delivery stats, unsubscribe, drafts/templates, variables, all-users sends, and edge cases.
>   - Test helpers improved (mailbox wait utilities, auth fast signup).
> - **Infra/Tooling**:
>   - Add Vercel cron for email queue step; dev script to run queue locally.
>   - Docker deps: add Drizzle Gateway; tweak Postgres and compose; CI workflow renamed to build/run.
>   - Misc refactors (ESM config, Next/Sentry adjustments), README for email pipeline.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 313c229d09ddc5ab12b5d27be1a4244ad64ae7dc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Asynchronous Email Outbox pipeline (queuing, rendering, scheduling, priority, drafts) and internal queue runner
  * Delivery statistics API with capacity metrics
  * Send-email API: draft/template/html support, high-priority flag, and results now return user_id
  * Unsubscribe handling and two‑pass template rendering; unsubscribe links

* **Bug Fixes**
  * Improved deliverability checks, error reporting, retries and skip conditions

* **Documentation**
  * Email infrastructure README added

* **Tests**
  * Extensive e2e and unit tests for email flows and delivery-info endpoint

### DIFF
--- a/apps/backend/prisma/migrations/20251212180000_email_outbox/migration.sql
+++ b/apps/backend/prisma/migrations/20251212180000_email_outbox/migration.sql
@@ -165,9 +165,9 @@ CREATE TABLE "EmailOutbox" (
                     ("renderedHtml" IS NULL AND "renderedText" IS NULL AND "renderedSubject" IS NULL AND "renderedIsTransactional" IS NULL AND "renderedNotificationCategoryId" IS NULL)
                     AND (
                         "renderErrorExternalMessage" IS NOT NULL
-                        OR "renderErrorExternalDetails" IS NOT NULL
-                        OR "renderErrorInternalMessage" IS NOT NULL
-                        OR "renderErrorInternalDetails" IS NOT NULL
+                        AND "renderErrorExternalDetails" IS NOT NULL
+                        AND "renderErrorInternalMessage" IS NOT NULL
+                        AND "renderErrorInternalDetails" IS NOT NULL
                     )
                 )
             )
@@ -219,7 +219,12 @@ CREATE TABLE "EmailOutbox" (
             <= 1
         ),
     CONSTRAINT "EmailOutbox_click_implies_open_check"
-        CHECK ("clickedAt" IS NULL OR "openedAt" IS NOT NULL)
+        CHECK ("clickedAt" IS NULL OR "openedAt" IS NOT NULL),
+    CONSTRAINT "EmailOutbox_send_server_error_all_or_none_check"
+        CHECK (
+            ("sendServerErrorExternalMessage" IS NULL AND "sendServerErrorExternalDetails" IS NULL AND "sendServerErrorInternalMessage" IS NULL AND "sendServerErrorInternalDetails" IS NULL)
+            OR ("sendServerErrorExternalMessage" IS NOT NULL AND "sendServerErrorExternalDetails" IS NOT NULL AND "sendServerErrorInternalMessage" IS NOT NULL AND "sendServerErrorInternalDetails" IS NOT NULL)
+        )
 );
 
 -- CreateTable


apps/backend/src/lib/email-delivery-stats.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const ratePerSecond = ratePerHour / 60 / 60;
+  return { ratePerSecond, penaltyFactor };
+}
+
+const deliveryStatsQuery = (tenancyId: string): RawQuery<EmailDeliveryStats> => ({
+  supportedPrismaClients: ["global"],
+  sql: Prisma.sql`
+    SELECT
+      SUM(CASE WHEN "finishedSendingAt" >= NOW() - INTERVAL '1 hour' AND "sendServerErrorInternalMessage" IS NULL AND "skippedReason" IS NULL THEN 1 ELSE 0 END)::bigint AS sent_last_hour,
+      SUM(CASE WHEN "finishedSendingAt" >= NOW() - INTERVAL '1 day' AND "sendServerErrorInternalMessage" IS NULL AND "skippedReason" IS NULL THEN 1 ELSE 0 END)::bigint AS sent_last_day,
```
BilalG1, line 50: can probably make this a clickhouse query later

N2D4: ooo that would be great, this query is p slow right now

BilalG1 (approved):

PR #1031, opened 2025-11-26
convex fix auth config bug

BilalG1:
## Summary by CodeRabbit

* **New Features**
  * Added new `./convex-auth.config` export path for Convex authentication configuration across all packages.

* **Documentation**
  * Updated documentation and examples with new import paths for Convex auth integration in JavaScript, React, and Next.js projects.

### DIFF
--- a/examples/convex/convex/auth.config.ts
+++ b/examples/convex/convex/auth.config.ts
@@ -1,4 +1,4 @@
-import { getConvexProvidersConfig } from "@stackframe/stack/convex";
+import { getConvexProvidersConfig } from "@stackframe/stack/convex-auth.config";
 
 export default {
   providers: getConvexProvidersConfig({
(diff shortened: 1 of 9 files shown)


examples/convex/convex/auth.config.ts
```diff
@@ -1,4 +1,4 @@
-import { getConvexProvidersConfig } from "@stackframe/stack";
+import { getConvexProvidersConfig } from "@stackframe/stack/convex";
```
N2D4: can we call this import `@stackframe/stack/convex-auth.config`? that way it's consistent with `@stackframe/stack/convex.config`, and also makes it clear that this file is exclusively for auth stuff (and we can never accidentally import it)

also, this code now also works again even when importing from `@stackframe/stack`, right?

N2D4 (approved):

PR #1036, opened 2025-11-30
External db sync

aadesh18:
Screenshot 2026-02-04 at 9 47 16 AM

**This PR revolves around the following components**
1. Sequencer - sequences the updates in the internal db
2. Poller - polls for the latest updates to sync with the external db
3. Outgoing Request Handler - essentially a trigger that can make http requests based on a change in the internal db
4. Sync Engine - syncs with the latest changes from the internal db to the external db

**What has been done**
- Added a global sequence id for ProjectUser, ContactChannel and DeletedRow.
- Added the deletedRow table to keep track of the rows that were deleted across ProjectUser and ContactChannel.
- Added the OutgoingRequest table to keep track of the outgoing requests 
- Added function for the sequencer to call to sequence updates
- Added a sequencer that sequences all the changes in the internal db every 50 ms
- Added a poller that polls for the latest changes in the internal db every 50 ms, and adds to a queue
- Added a Vercel cron that calls sequencer and poller every minute
- Added a queue that fulfills the outgoing requests by making http calls (for external db sync, it calls the sync engine endpoint)
- Added a sync engine that uses the defined sql mapping query in the user's schema to pull in the changes for the user, and sync them with the external db
- Added tests to test out each functionality

**How to review this PR:**
1. Review the migrations (sequence id, deletedRow,  triggers, backlog sync) (all files created under the migrations folder)
2. Review sequencer
3. Review poller
4. Review the changes in schema
5. Review sync-engine (the function, and it's helper file)
6. Review the schema changes, and query mappings
7. Review the tests (basic, advanced and race, along with the helper file)
8. Review the changes made in Dockerfile to support local testing using the postgres docker 

---

> [!NOTE]
> Introduces a cron-driven external DB sync pipeline with global sequencing, internal poller and webhook sync engine, new DB tables/functions, config schema/mappings, and comprehensive e2e tests.
> 
> - **Database (Prisma/Migrations)**:
>   - Add global sequence (`global_seq_id`) and `sequenceId`/`shouldUpdateSequenceId` to `ProjectUser`, `ContactChannel`, `DeletedRow` with partial indexes.
>   - Create `DeletedRow` (capture deletes) and `OutgoingRequest` (queue) tables; add unique/indexes.
>   - Add triggers/functions: `log_deleted_row`, `reset_sequence_id_on_update`, `backfill_null_sequence_ids`, `enqueue_tenant_sync`.
> - **Backend/API**:
>   - New internal routes: `GET /api/latest/internal/external-db-sync/sequencer`, `GET /poller`, `POST /sync-engine` (Upstash-verified) for sync orchestration.
>   - Add cron wiring: `vercel.json` schedules and local `scripts/run-cron-jobs.ts`; start in dev via `dev` script.
>   - Tweak route handler (remove noisy logging) without behavior change.
> - **Sync Engine**:
>   - Implement `src/lib/external-db-sync.ts` to read tenant mappings and upsert to external Postgres (schema bootstrap, param checks, sequencing).
>   - Add default mappings `DEFAULT_DB_SYNC_MAPPINGS` and config schema `dbSync.externalDatabases` in shared config.
> - **Testing/Infra**:
>   - Add extensive e2e tests (basics, advanced, race conditions) for sequencing, idempotency, deletes, pagination, multi-mapping, and permissions.
>   - Docker compose: add `external-db-test` Postgres for tests; e2e deps for `pg` types.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3f2a8efcfbd94789c8ffd83cf967e06ca942014f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * External PostgreSQL sync: automatic, batched replication with mappings, resume/idempotency, and on-demand enqueueing.

* **Admin UI**
  * Real-time External DB Sync dashboard and status API showing per-mapping backlog, sequencer/poller/sync-engine telemetry, and fusebox controls.

* **Tests**
  * Large e2e suite: basic, advanced, race, high-volume tests and test utilities for external DB sync.

* **Chores**
  * DB migrations, CI/workflow updates, background cron runner and local/dev test support.

### DIFF
--- a/apps/e2e/tests/backend/endpoints/api/v1/users.test.ts
+++ b/apps/e2e/tests/backend/endpoints/api/v1/users.test.ts
@@ -9,7 +9,7 @@ describe("without project access", () => {
     projectKeys: "no-project",
   });
 
-  it.only("should not be able to read own user", async ({ expect }) => {
+  it("should not be able to read own user", async ({ expect }) => {
     await backendContext.with({
       projectKeys: InternalProjectKeys,
     }, async () => {


apps/backend/prisma/migrations/20251125030551_add_sequence_id/migration.sql
N2D4, line 1: can you merge all these migrations into a single one and document what each of the steps do?

apps/backend/prisma/migrations/20251128210004_add_sync_functions/migration.sql
```diff
@@ -0,0 +1,100 @@
...
+    SELECT 1
+    FROM "OutgoingRequest"
+    WHERE "fulfilledAt" IS NULL
+      AND ("qstashOptions"->'body'->>'tenantId')::uuid = p_tenant_id
+  );
+END;
+$$ LANGUAGE plpgsql;
+-- SPLIT_STATEMENT_SENTINEL
+-- SINGLE_STATEMENT_SENTINEL
+CREATE FUNCTION backfill_null_sequence_ids()
```
N2D4: this is no longer on pgcron now, right? do we still need this? can't we just have this query in the sequencer endpoint?

apps/backend/prisma/schema.prisma
```diff
@@ -173,6 +173,9 @@ model ProjectUser {
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
 
+  sequenceId BigInt? @unique
+  shouldUpdateSequenceId Boolean @default(true)
```
N2D4: these fields should be indexed, i think?

apps/backend/prisma/schema.prisma
```diff
@@ -874,3 +880,32 @@ model SubscriptionInvoice {
...
+
+model OutgoingRequest {
+  id String @id @default(uuid()) @db.Uuid
+
+  createdAt DateTime @default(now())
+
+  qstashOptions Json
+  fulfilledAt   DateTime?
+
+  @@index([fulfilledAt])
```
N2D4: maybe we can rename this to "startedFulfillingAt" just to make it clearer for the future

apps/backend/prisma/schema.prisma
```diff
@@ -252,6 +255,9 @@ model ContactChannel {
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
 
+  sequenceId BigInt? @unique
+  shouldUpdateSequenceId Boolean @default(true)
```
N2D4: as above

apps/backend/prisma/schema.prisma
```diff
@@ -874,3 +880,32 @@ model SubscriptionInvoice {
...
+  @@index([fulfilledAt])
+}
+
+model DeletedRow {
+  id        String @id @default(uuid()) @db.Uuid
+  tenancyId String @db.Uuid
+  tableName String
+
+  sequenceId             BigInt?  @unique
+  shouldUpdateSequenceId Boolean  @default(true)
```
N2D4: as above

apps/backend/prisma/schema.prisma
```diff
@@ -874,3 +880,32 @@ model SubscriptionInvoice {
...
+
+  primaryKey Json
+  data       Json?
+
+  deletedAt   DateTime  @default(now())
+  fulfilledAt DateTime?
+
+  @@index([tableName])
+  @@index([tenancyId])
```
N2D4, line 1113: shouldn't we index by tenancy > tableName > primaryKey or something like that?

apps/backend/src/app/api/latest/internal/external-db-sync/poller/route.ts
```diff
@@ -0,0 +1,129 @@
...
+      let processed = 0;
+
+      for (const request of requests) {
+        try {
+          const options = request.qstashOptions as any;
+          const baseUrl = getEnvVariable("NEXT_PUBLIC_STACK_API_URL");
+
+          let fullUrl = options.url.startsWith("http")
+            ? options.url
+            : new URL(options.url, baseUrl).toString();
```
N2D4: can you just do this?

```suggestion
          let fullUrl = new URL(options.url, baseUrl).toString();
```

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,77 @@
...
+    }).defined(),
+  }),
+  handler: async ({ headers }) => {
+    const authHeader = headers.authorization[0];
+    if (authHeader !== `Bearer ${getEnvVariable("CRON_SECRET")}`) {
+      throw new StatusError(401, "Unauthorized");
+    }
+
+    const startTime = performance.now();
+    const maxDurationMs = 2 * 60 * 1000;
```
N2D4: thinking about this, maybe we should set it to 3min, because in the worst case vercel cron might schedule the calls 2min apart of each other (start of minute :05 and end of minute :06)

apps/backend/src/app/api/latest/internal/external-db-sync/poller/route.ts
```diff
@@ -0,0 +1,129 @@
...
+    }).defined(),
+  }),
+  handler: async ({ headers }) => {
+    const authHeader = headers.authorization[0];
+    if (authHeader !== `Bearer ${getEnvVariable("CRON_SECRET")}`) {
+      throw new StatusError(401, "Unauthorized");
+    }
+
+    const startTime = performance.now();
+    const maxDurationMs = 2 * 60 * 1000;
```
N2D4: thinking about this, maybe we should set it to 3min, because in the worst case vercel cron might schedule the calls 2min apart of each other (start of minute :05 and end of minute :06)

apps/backend/scripts/run-cron-jobs.ts
```diff
@@ -0,0 +1,36 @@
...
+    });
+    if (!res.ok) throw new StackAssertionError(`Failed to call ${endpoint}: ${res.status} ${res.statusText}\n${await res.text()}`, { res });
+    console.log(`${endpoint} completed.`);
+  });
+
+  for (const endpoint of endpoints) {
+    setInterval(() => {
+      run(endpoint);
+    }, 60000);
+  }
```
N2D4, line 37: vercel actually doesn't guarantee execution every exact 60s but instead just that it will be scheduled somewhere within the minute, so this might actually be random between 0s and 120s. maybe we should do that as well in the local emulator?

```suggestion
  for (const endpoint of endpoints) {
    runAsynchronously(async () => {
      while (true) {
        run(endpoint);
        // Vercel only guarantees minute-granularity for cron jobs, so we randomize the interval
        await wait(Math.random() * 120_000);
      }
    }); 
  }
```

apps/backend/src/app/api/latest/internal/external-db-sync/poller/route.ts
```diff
@@ -0,0 +1,129 @@
...
+          SET "fulfilledAt" = NOW()
+          WHERE "id" IN (
+            SELECT id
+            FROM "OutgoingRequest"
+            WHERE "fulfilledAt" IS NULL
+            ORDER BY "createdAt"
+            LIMIT 100
+            FOR UPDATE SKIP LOCKED
+          )
+          RETURNING *;
```
N2D4: make sure that we have an index for [fulfilledAt, createdAt] so this is fast

apps/backend/src/app/api/latest/internal/external-db-sync/poller/route.ts
```diff
@@ -0,0 +1,129 @@
...
+            body: options.body,
+          });
+
+          processed++;
+        } catch (error) {
+          console.error(
+            `Failed to process outgoing request ${request.id}:`,
+            error,
+          );
+        }
```
N2D4: BAD BAD BAD BAD BAD BAD BAD BAD BAD BAD

apps/backend/src/app/api/latest/internal/external-db-sync/poller/route.ts
```diff
@@ -0,0 +1,129 @@
...
+            ? options.url
+            : new URL(options.url, baseUrl).toString();
+
+          if (getNodeEnvironment().includes("development") || getNodeEnvironment().includes("test")) {
+            const url = new URL(fullUrl);
+            if (url.hostname === "localhost" || url.hostname === "127.0.0.1") {
+              url.hostname = "host.docker.internal";
+              fullUrl = url.toString();
+            }
+          }
```
N2D4: add a comment explaining why this is needed

apps/backend/src/app/api/latest/internal/external-db-sync/poller/route.ts
```diff
@@ -0,0 +1,129 @@
...
+  }),
+  handler: async ({ headers }) => {
+    const authHeader = headers.authorization[0];
+    if (authHeader !== `Bearer ${getEnvVariable("CRON_SECRET")}`) {
+      throw new StatusError(401, "Unauthorized");
+    }
+
+    const startTime = performance.now();
+    const maxDurationMs = 2 * 60 * 1000;
+    const busySleepMs = 50;
```
N2D4: i wouldn't call this busySleep, a busy loop is what we call just wasting CPU cycles in an infinite loop, but the sleep you're doing here is not busy (it's just a normal sleep)

apps/backend/src/app/api/latest/internal/external-db-sync/poller/route.ts
N2D4:
```suggestion
        } catch (error) {
          captureError(
            `poller-iteration-error`,
            error,
          );
        }
```

N2D4: try-catch-everything considered harmful. (we should ban it with eslint at some point). if you do really want to catch everything, use captureError to ensure that the data captured alerts us

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,77 @@
...
+    const sleepMs = 50;
+
+    let iterations = 0;
+
+    while (performance.now() - startTime < maxDurationMs) {
+      try {
+        await globalPrismaClient.$executeRaw`SELECT backfill_null_sequence_ids()`;
+      } catch (error) {
+        console.warn('[sequencer] Failed to run backfill_null_sequence_ids:', error);
+      }
```
N2D4: BAD BAD BAD BAD BAD BAD

N2D4: fix with captureError as above

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const startTime = performance.now();
+    const maxDurationMs = 2 * 60 * 1000;
+    const sleepMs = 50;
+
+    let iterations = 0;
+
+    while (performance.now() - startTime < maxDurationMs) {
+      try {
+        await globalPrismaClient.$executeRaw`SELECT backfill_null_sequence_ids()`;
```
N2D4: we should probably just put the sql query inline in here instead of calling a function

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,77 @@
...
+      } catch (error) {
+        console.warn('[sequencer] Failed to run backfill_null_sequence_ids:', error);
+      }
+
+      iterations++;
+
+      const elapsed = performance.now() - startTime;
+      if (elapsed >= maxDurationMs) {
+        break;
+      }
```
N2D4: this seems a bit unnecessary?

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }).defined(),
+    method: yupString().oneOf(["POST"]).defined(),
+  }),
+  response: yupObject({
+    statusCode: yupNumber().oneOf([200]).defined(),
+    bodyType: yupString().oneOf(["json"]).defined(),
+    body: yupObject({
+      success: yupBoolean().defined(),
+      tenantId: yupString().defined(),
+      timestamp: yupString().defined(),
```
N2D4: what does this do?

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
```diff
@@ -0,0 +1,68 @@
...
+      console.error(`Tenant not found: ${tenantId}`);
+      return {
+        statusCode: 200,
+        bodyType: "json",
+        body: {
+          success: false,
+          tenantId,
+          timestamp,
+        },
+      };
```
N2D4, line 65: status code 200 with success: false seems wrong?

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    try {
+      await syncExternalDatabases(tenancy);
+    } catch (error: any) {
+      console.error(` Error syncing external databases for tenant ${tenantId}:`, error);
+    }
```
N2D4: BAD

N2D4: this should probably just not catch the error and return a 5xx

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
N2D4: you probably want to return a 4xx error here instead (if I understand correctly, this will trigger if a project is deleted, right?)

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }),
+  handler: async ({ body }, fullReq) => {
+    await ensureUpstashSignature(fullReq);
+
+    const { tenantId } = body;
+    const timestamp = new Date().toISOString();
+
+    const tenancy = await getTenancy(tenantId);
+    if (!tenancy) {
+      console.error(`Tenant not found: ${tenantId}`);
```
N2D4: there are very few good use cases for console.error in backend code. in this case, maybe a console.warn makes sense (describing what happened and why it happened, eg. "tenancy on queue not found, assuming it was deleted")

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
```diff
@@ -0,0 +1,68 @@
...
+    description: "Receives webhook from QStash to trigger external database sync for a tenant",
+    tags: ["External DB Sync"],
+    hidden: true,
+  },
+  request: yupObject({
+    headers: yupObject({
+      "upstash-signature": yupTuple([yupString()]).defined(),
+    }).defined(),
+    body: yupObject({
+      tenantId: yupString().defined(),
```
N2D4: it's called tenancyId

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+  } catch (error: any) {
+    console.error(`Error syncing external DB ${dbId}:`, error);
+  } finally {
+    await externalClient.end();
+  }
+}
+
+
+export async function syncExternalDatabases(tenancy: Tenancy) {
+  const externalDatabases = getExternalDatabases(tenancy.config);
```
N2D4: i think we can just inline this, no need for an extra function

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+  if (Object.keys(externalDatabases).length === 0) {
+    return;
+  }
```
N2D4: this is unnecessary

N2D4: (checks like these that single out special cases can hide implementation bugs as they might make bugs not trigger in the special cases)

N2D4: (also, concise & minimal code is always good. if you want to let the reader know "this code does that when there are no external databases", write it as a comment)

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+        mappingId,
+        mapping as any,
+        internalPrisma,
+        dbId,
+        tenancyId,
+      );
+    }
+
+  } catch (error: any) {
+    console.error(`Error syncing external DB ${dbId}:`, error);
```
N2D4: BAD BAD

N2D4: i'll stop commenting this now, i think you get the gist of it :]

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+  internalPrisma: PrismaClientTransaction,
+  tenancyId: string,
+) {
+
+  const mappings = dbConfig.mappings;
+
+  const isArray = Array.isArray(mappings);
+  const mappingCount = mappings
+    ? (isArray ? (mappings as any[]).length : Object.keys(mappings as Record<string, unknown>).length)
+    : 0;
```
N2D4: what is this for? remnant of an old version?

```suggestion
  const isArray = Array.isArray(mappings);
  const mappingCount = Object.keys(mappings as Record<string, unknown>).length;
```

packages/stack-shared/src/config/schema.ts
```diff
@@ -200,6 +201,27 @@ export const branchConfigSchema = canNoLongerBeOverridden(projectConfigSchema, [
 
   payments: branchPaymentsSchema,
 
+  dbSync: yupObject({
+    externalDatabases: yupRecord(
+      userSpecifiedIdSchema("externalDatabaseId"),
+      yupObject({
+        type: yupString().oneOf(['neon', 'postgres']).defined(),
```
N2D4: should probably just be `postgres`, no need to have a separate one for Neon

packages/stack-shared/src/config/schema.ts
```diff
@@ -200,6 +201,27 @@ export const branchConfigSchema = canNoLongerBeOverridden(projectConfigSchema, [
...
+        connectionString: yupString().defined(),
+        mappings: yupRecord(
```
N2D4: users should NOT be able to configure this, this should not be in the schema — the schema is fully configurable to any value, so users would be able to run any query they want on Stack Auth's DB

packages/stack-shared/src/config/schema.ts
```diff
@@ -200,6 +201,27 @@ export const branchConfigSchema = canNoLongerBeOverridden(projectConfigSchema, [
...
+          userSpecifiedIdSchema("mappingId"),
+          yupObject({
+            sourceTables: yupRecord(yupString(), yupString()).optional(),
+            targetTable: yupString().defined(),
+            targetTableSchema: yupString().optional(),
+            targetTablePrimaryKey: yupTuple([yupString().defined()]).optional(),
+            internalDbFetchQuery: yupString().defined(),
+            externalDbUpdateQuery: yupString().optional(),
+          })
+        ).default(() => DEFAULT_DB_SYNC_MAPPINGS as any),
```
N2D4: default here does nothing, you need to put it into the organizationConfigDefaults

packages/stack-shared/src/config/schema.ts
```diff
@@ -863,7 +894,7 @@ export async function getConfigOverrideErrors<T extends yup.AnySchema>(schema: T
         // This is how the implementation would look like, but we don't support arrays in config JSON files (besides tuples)
         // const arraySchema = schema as yup.ArraySchema<any, any, any, any>;
         // const innerType = arraySchema.innerType;
-        // return yupArray(innerType ? getRestrictedSchema(path + ".[]", innerType as any) : undefined);
+        // return yupArray(innerType ? getRestrictedSchema(path + ".[]", innerType as any) : yupMixed());
```
N2D4: <p class='graphite__hidden'><i>[Re: lines +893 to +897]</i></p>
i think we can just delete this comment (not very useful)

```gt_suggestion
```
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/stack-auth/stack-auth/1036?utm_source=unchanged-line-comment">Graphite</a>.</p>

vercel.json
```diff
@@ -1,4 +1,14 @@
 {
+  "crons": [
+    {
+      "path": "/api/latest/internal/external-db-sync/poller",
+      "schedule": "* * * * *"
+    },
+    {
+      "path": "/api/latest/internal/external-db-sync/sequencer",
+      "schedule": "* * * * *"
```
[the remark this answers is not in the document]
N2D4: that's intentional

package.json
```diff
@@ -65,6 +65,7 @@
     "@changesets/cli": "^2.27.9",
     "@testing-library/react": "^15.0.7",
     "@types/node": "20.17.6",
+    "@types/pg": "^8.15.6",
```
N2D4: did you mean to add this to apps/backend/package.json?

packages/stack-shared/src/config/db-sync-mappings.ts
```diff
@@ -0,0 +1,166 @@
+export const DEFAULT_DB_SYNC_MAPPINGS = {
+  "PartialUsers": {
+    sourceTables: { "ContactChannel": "ContactChannel", "ProjectUser": "ProjectUser" },
+    targetTable: "PartialUsers",
+    targetTablePrimaryKey: ["id"],
+    targetTableSchema: `
```
N2D4:
```suggestion
    postgresTargetTableSchema: `
```

packages/stack-shared/src/config/db-sync-mappings.ts
```diff
@@ -0,0 +1,166 @@
+export const DEFAULT_DB_SYNC_MAPPINGS = {
+  "PartialUsers": {
+    sourceTables: { "ContactChannel": "ContactChannel", "ProjectUser": "ProjectUser" },
+    targetTable: "PartialUsers",
```
N2D4: I think we can just call this `Users`​

N2D4: actually, what is this used for? don't the targetTableSchema and externalDbUpdateQuery already solve this?

packages/stack-shared/src/config/db-sync-mappings.ts
```diff
@@ -0,0 +1,166 @@
...
+        "userUpdatedAt" timestamp with time zone,
+        "profileImageUrl" text,
+        "displayName" text,
+        "userCreatedAt" timestamp with time zone,
+        "isAnonymous" boolean
+      );
+      CREATE INDEX ON "PartialUsers" ("sequenceId");
+      REVOKE ALL ON "PartialUsers" FROM PUBLIC;
+      GRANT SELECT ON "PartialUsers" TO PUBLIC;
+    `.trim(),
```
N2D4: can we replace this by just having some code in the externalDbUpdateQuery which does something along the lines of "if table does not exist yet, create it"? that seems like a safer and cleaner way to initialize the schema (eg. imagine what happens when a user wipes the database)

packages/stack-shared/src/config/db-sync-mappings.ts
```diff
@@ -0,0 +1,166 @@
...
+        FROM "DeletedRow"
+        WHERE
+          "DeletedRow"."tenancyId" = $1::uuid
+          AND "DeletedRow"."tableName" = 'ContactChannel'
+      ) AS "_src"
+      WHERE "sequenceId" IS NOT NULL
+      ORDER BY "sequenceId" ASC
+      LIMIT 1000
+    `.trim(),
+    externalDbUpdateQuery: `
```
N2D4:
```suggestion
    postgresExternalDbUpdateQuery: `
```

packages/stack-shared/src/config/db-sync-mappings.ts
```diff
@@ -0,0 +1,166 @@
...
+        "sequenceId" bigint,
+        "userUpdatedAt" timestamp with time zone,
+        "profileImageUrl" text,
+        "displayName" text,
+        "userCreatedAt" timestamp with time zone,
+        "isAnonymous" boolean
+      );
+      CREATE INDEX ON "PartialUsers" ("sequenceId");
+      REVOKE ALL ON "PartialUsers" FROM PUBLIC;
+      GRANT SELECT ON "PartialUsers" TO PUBLIC;
```
N2D4: this seems dangerous, why is this here?

packages/stack-shared/src/config/db-sync-mappings.ts
```diff
@@ -0,0 +1,166 @@
...
+    targetTableSchema: `
+   CREATE TABLE IF NOT EXISTS "PartialUsers" (
+        "id" uuid PRIMARY KEY,
+        "createdAt" timestamp with time zone,
+        "updatedAt" timestamp with time zone,
+        "type" text,
+        "isPrimary" boolean,
+        "isVerified" boolean,
+        "value" text,
+        "sequenceId" bigint,
```
N2D4: the sequenceId shouldn't be on the Users table on each row. rather, we should have a metadata table on the target postgres, that has a field "most recently sync'd sequenceId" instead

packages/stack-shared/src/config/db-sync-mappings.ts
```diff
@@ -0,0 +1,166 @@
...
+    sourceTables: { "ContactChannel": "ContactChannel", "ProjectUser": "ProjectUser" },
+    targetTable: "PartialUsers",
+    targetTablePrimaryKey: ["id"],
+    targetTableSchema: `
+   CREATE TABLE IF NOT EXISTS "PartialUsers" (
+        "id" uuid PRIMARY KEY,
+        "createdAt" timestamp with time zone,
+        "updatedAt" timestamp with time zone,
+        "type" text,
+        "isPrimary" boolean,
```
N2D4: isPrimary does not make sense as an attribute on the user

packages/stack-shared/src/config/db-sync-mappings.ts
```diff
@@ -0,0 +1,166 @@
...
+        "isVerified" boolean,
```
N2D4: as above with isPrimary

packages/stack-shared/src/config/db-sync-mappings.ts
```diff
@@ -0,0 +1,166 @@
...
+        "value" text,
```
N2D4: ?

packages/stack-shared/src/config/db-sync-mappings.ts
N2D4: actually `users`, because postgres traditionally uses snake_case

packages/stack-shared/src/config/db-sync-mappings.ts
```diff
@@ -0,0 +1,166 @@
+export const DEFAULT_DB_SYNC_MAPPINGS = {
+  "PartialUsers": {
+    sourceTables: { "ContactChannel": "ContactChannel", "ProjectUser": "ProjectUser" },
+    targetTable: "PartialUsers",
+    targetTablePrimaryKey: ["id"],
+    targetTableSchema: `
+   CREATE TABLE IF NOT EXISTS "PartialUsers" (
```
N2D4: let's keep the PartialUsers resembling the [API user type](https://docs.stack-auth.com/api/client/users/users/me/get). they should also all be snake_case

```
id: non-null UUID
display_name: nullable string,
profile_image_url: nullable string,
primary_email: nullable string,
primary_email_verified: non-null boolean,
signed_up_at: non-null timestamp (without timezone),
client_metadata: non-null jsonb,  // the jsonb itself can be null ofc, but no dbnull
client_readonly_metadata: non-null jsonb,  // the jsonb itself can be null ofc, but no dbnull
server_metadata: non-null jsonb,  // the jsonb itself can be null ofc, but no dbnull
is_anonymous: non-null boolean,
```

packages/stack-shared/src/config/db-sync-mappings.ts
N2D4: or maybe even externalDbUpdateQueries: { postgres: `...` } ?

docker/dependencies/docker.compose.yaml
```diff
@@ -204,6 +204,19 @@ services:
...
+    image: "docker.io/postgres:16.1"
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: external-db-test-password
+      POSTGRES_DB: postgres
+    ports:
+      - "${NEXT_PUBLIC_STACK_PORT_PREFIX:-81}32:5432"
+    volumes:
+      - external-db-test-data:/var/lib/postgresql/data
+
```
N2D4: why a new docker container? can't we use the same postgres but a different db on the other one?

apps/backend/src/route-handlers/smart-route-handler.tsx
N2D4, line 1: why would you delete all of this

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  if (!dbConfig.connectionString) {
+    return;
+  }
```
N2D4, line 435: somewhere here we should also check the type, and just throw an error "unimplemented" if it's not postgres

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+  });
+
+  try {
+    await externalClient.connect();
+
+    if (!mappings || mappingCount === 0) {
+      return;
+    }
+
+    for (const [mappingId, mapping] of Object.entries(mappings)) {
```
N2D4:
```suggestion
    for (const [mappingId, mapping] of Object.entries(DB_DEFAULT_MAPPINGS)) {
```

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+        await externalClient.query(mapping.targetTableSchema);
+      } catch (err: any) {
+        if (err.code !== '23505' || err.constraint !== 'pg_type_typname_nsp_index') {
+          throw err;
+        }
+      }
+    }
+  }
+
+  const rows = await internalPrisma.$queryRawUnsafe<any[]>(fetchQuery, tenancyId);
```
N2D4: fetchQuery currently only fetches 1000 rows at max, we should iterate over that loop so that it fetches more than 1000 rows if the database was down  
  
also please add an E2E test for this behavior (if we can) — create a ton of users while the external db is down, and then make sure that a single external db call catches up all of them

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  const rawSourceTables: any = (mapping as any).sourceTables;
+  const sourceTables: string[] = rawSourceTables
+    ? Object.values(rawSourceTables)
+    : [];
+
+  const rawTargetPk: any = (mapping as any).targetTablePrimaryKey;
+  const targetTablePrimaryKey: string[] = rawTargetPk
+    ? Object.values(rawTargetPk)
+    : [];
```
N2D4: what's the point of this? it seems unused

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+  dbId: string,
+  tenancyId: string,
+) {
+
+  const rawSourceTables: any = (mapping as any).sourceTables;
+  const sourceTables: string[] = rawSourceTables
+    ? Object.values(rawSourceTables)
+    : [];
+
+  const rawTargetPk: any = (mapping as any).targetTablePrimaryKey;
```
N2D4: we can remove all the casts and stuff once `mapping` is just the db default mapping, instead of from the config schema

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+    );
+    return;
+  }
+
+  if (targetTablePrimaryKey.length === 0) {
+    console.error(
+      ` Invalid configuration for mapping #${mappingId}: 'targetTablePrimaryKey' resolved to an empty list.`,
+    );
+    return;
+  }
```
N2D4: "console.error and return undefined" is a deadly pattern. `throw new StackAssertionError` instead

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  const fetchQuery = mapping.internalDbFetchQuery;
+  if (!fetchQuery || !mapping.targetTable) {
+    return;
+  }
```
N2D4: generally whenever i see something like this it's a bit of an antipattern. why exactly would either of these be undefined, and why does that mean we should just skip a table? a comment can help explain

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+    const res = await externalClient.query(checkTableQuery, [tableName]);
+    if (!res.rows[0].exists) {
+      try {
+        await externalClient.query(mapping.targetTableSchema);
+      } catch (err: any) {
+        if (err.code !== '23505' || err.constraint !== 'pg_type_typname_nsp_index') {
+          throw err;
+        }
+      }
+    }
```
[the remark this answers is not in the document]
N2D4: yep! if we put the creation into the schema update query (as explained on the mapping defaults) this won't be an issue btw :]

apps/backend/src/lib/external-db-sync.ts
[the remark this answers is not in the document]
N2D4: good bot!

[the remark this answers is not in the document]
N2D4: good bot

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+async function pushRowsToExternalDb(
+  externalClient: Client,
+  tableName: string,
+  newRows: any[],
+  upsertQuery?: string,
+) {
+  if (!upsertQuery) {
+    throw new Error(
```
N2D4: StackAssertionError instead. actually, what's the point of this? you can just make it required in the function arguments and remove this check

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+      `Cannot sync table "${tableName}": No upsertQuery configured.`
+    );
+  }
+
+  if (newRows.length === 0) return;
```
N2D4, line 274: unnecessary

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+  const orderedKeys = Object.keys(restSample);
+
+  if (orderedKeys.length !== expectedParamCount) {
+    throw new Error(
+      ` Column count mismatch for table "${tableName}".\n` +
+        `→ upsertQuery expects ${expectedParamCount} parameters.\n` +
+        `→ internalDbFetchQuery returned ${orderedKeys.length} columns (excluding tenancyId).\n` +
+        `Fix your SELECT column order or your SQL parameter order.`
+    );
+  }
```
N2D4:
```suggestion
  // Just for our own sanity, make sure that we have the right number of positional parameters
  const placeholderMatches = upsertQuery.match(/\$\d+/g) ?? throwErr(`Could not find any positional parameters ($1, $2, ...) in the update SQL query.`);"
  const expectedParamCount = Math.max(...placeholderMatches.map((m) => Number(m.slice(1))));
  const sampleRow = newRows[0];
  const orderedKeys = Object.keys(omit(sampleRow, "tenancyId"));
  if (orderedKeys.length !== expectedParamCount) {
    throw new StackAssertionError(deindent`
      Column count mismatch for table ${JSON.stringify(tableName)}
       → upsertQuery expects ${expectedParamCount} parameters.
       → internalDbFetchQuery returned ${orderedKeys.length} columns (excluding tenancyId).
      Fix your SELECT column order or your SQL parameter order.
    `);
  }
```

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+  for (const row of newRows) {
+    const { tenancyId, ...rest } = row;
+    const rowKeys = Object.keys(rest);
+
+    const validShape =
+      rowKeys.length === orderedKeys.length &&
+      rowKeys.every((k, i) => k === orderedKeys[i]);
+
+    if (!validShape) {
+      throw new Error(
```
N2D4:
```suggestion
      throw new StackAssertionError(
```

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+          `Your SELECT must be explicit, ordered, and NEVER use SELECT *.\n` +
+          `Fix the SELECT in internalDbFetchQuery immediately.`
+      );
+    }
+  }
+  for (const row of newRows) {
+    const { tenancyId, ...rest } = row;
+    await externalClient.query(upsertQuery, Object.values(rest));
+  }
+}
```
N2D4, line 321: these two for-loops can be merged into one

apps/backend/src/lib/external-db-sync.ts
```diff
@@ -0,0 +1,203 @@
...
+          `Received column order: [${rowKeys.join(", ")}]\n` +
+          `Your SELECT must be explicit, ordered, and NEVER use SELECT *.\n` +
+          `Fix the SELECT in internalDbFetchQuery immediately.`
+      );
+    }
+  }
+  for (const row of newRows) {
+    const { tenancyId, ...rest } = row;
+    await externalClient.query(upsertQuery, Object.values(rest));
+  }
```
N2D4, line 320: should we check that the tenancyId is always the same? i think right now we don't do anything with it, right?

apps/backend/prisma/migrations/20251125030551_add_sequence_id/migration.sql
aadesh18: done - merged them into one and added comments

apps/backend/prisma/migrations/20251128210004_add_sync_functions/migration.sql
aadesh18: done - moved it into sequencer

apps/backend/prisma/schema.prisma
aadesh18: done - added indexes

apps/backend/prisma/schema.prisma
aadesh18: done - added indexes

apps/backend/prisma/schema.prisma
aadesh18: done - renamed

apps/backend/prisma/schema.prisma
aadesh18: done

apps/backend/prisma/schema.prisma
aadesh18: done - added index

apps/backend/src/app/api/latest/internal/external-db-sync/poller/route.ts
aadesh18: done

apps/backend/src/app/api/latest/internal/external-db-sync/poller/route.ts
aadesh18: done

apps/backend/src/app/api/latest/internal/external-db-sync/poller/route.ts
aadesh18: done - added index

apps/backend/src/app/api/latest/internal/external-db-sync/poller/route.ts
aadesh18: done - added

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
aadesh18: done

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
aadesh18: done

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
aadesh18: done

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
aadesh18: done - removed

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
aadesh18: done - renamed

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
aadesh18: done - simplified it

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
aadesh18: done

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
aadesh18: done

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
aadesh18: done

apps/backend/src/lib/external-db-sync.ts
aadesh18: done

apps/backend/src/lib/external-db-sync.ts
aadesh18: done - included a check

apps/backend/src/lib/external-db-sync.ts
aadesh18: done - merged

apps/backend/src/lib/external-db-sync.ts
aadesh18: done

apps/backend/src/lib/external-db-sync.ts
aadesh18: done - removed

apps/backend/src/lib/external-db-sync.ts
aadesh18: done

apps/backend/src/lib/external-db-sync.ts
aadesh18: done

apps/backend/src/lib/external-db-sync.ts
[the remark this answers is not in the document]
aadesh18: done

apps/backend/src/lib/external-db-sync.ts
aadesh18: done

apps/backend/src/lib/external-db-sync.ts
aadesh18: done

apps/backend/src/lib/external-db-sync.ts
aadesh18: done

apps/backend/src/lib/external-db-sync.ts
aadesh18: done

apps/backend/src/lib/external-db-sync.ts
aadesh18: done

apps/backend/src/lib/external-db-sync.ts
aadesh18: done

apps/backend/src/route-handlers/smart-route-handler.tsx
aadesh18: my bad, didnt realize cursor deleted this. ill be more careful. reverted back.

docker/dependencies/docker.compose.yaml
aadesh18: done

package.json
aadesh18: done

packages/stack-shared/src/config/db-sync-mappings.ts
aadesh18: done

packages/stack-shared/src/config/db-sync-mappings.ts
aadesh18: done

packages/stack-shared/src/config/db-sync-mappings.ts
aadesh18: done

packages/stack-shared/src/config/db-sync-mappings.ts
aadesh18: done

packages/stack-shared/src/config/db-sync-mappings.ts
aadesh18: done

packages/stack-shared/src/config/db-sync-mappings.ts
aadesh18: done

packages/stack-shared/src/config/db-sync-mappings.ts
aadesh18: done

packages/stack-shared/src/config/db-sync-mappings.ts
aadesh18: basically, the sync process connects using the users provided connection string (which im assuming has owner/superuser privileges), so it can write to the table. after revoking the all the permissions, we can select what permissions we wanna give to the users (its just select now, but we can also change it)

packages/stack-shared/src/config/db-sync-mappings.ts
aadesh18: done

packages/stack-shared/src/config/schema.ts
aadesh18: done

packages/stack-shared/src/config/schema.ts
aadesh18: done

packages/stack-shared/src/config/schema.ts
aadesh18: done

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,320 @@
...
+        AND pu."projectUserId" = r."projectUserId"
+      RETURNING pu."tenancyId"
+    )
+    SELECT DISTINCT "tenancyId" FROM updated_rows
+  `;
+
+  // Enqueue sync for each affected tenant
+  for (const { tenancyId } of projectUserTenants) {
+    assertUuid(tenancyId, "projectUserTenants.tenancyId");
+    await enqueueExternalDbSync(tenancyId);
```
N2D4: we currently enqueue every tenancy individually, but this means that enqueueExternalDbSync may  be called 1000x every time for each table. instead, we should do a single query which enqueues all 1000 tenancies at the same time

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,320 @@
...
+  if (value === "true") return true;
+  if (value === "false") return false;
+  throw new StatusError(400, "stopWhenIdle must be 'true' or 'false'");
+}
+
+function assertUuid(value: unknown, label: string): asserts value is string {
+  if (typeof value !== "string" || value.trim().length === 0 || !UUID_REGEX.test(value)) {
+    throw new StatusError(500, `${label} must be a valid UUID. Received: ${JSON.stringify(value)}`);
+  }
+}
```
N2D4: unnecessary as this should never not be a uuid (bigger things are wrong if it is). and if we do have it, it should use StackAssertionError

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,320 @@
...
+  for (const { id } of tenancyIds) {
+    const tenancy = await getTenancy(id);
+    if (!tenancy) continue;
+    if (tenancy.config.sourceOfTruth.type !== "hosted") {
+      tenancies.push(tenancy);
+    }
+  }
+
+  return tenancies;
+}
```
N2D4: this won't terminate, we have millions of tenancies

N2D4: I think we can just remove the entire logic surrounding this and keep a TODO comment somewhere; we don't do source-of-truth anymore and the DB doesn't contain any source of truth tenancies either

apps/backend/src/app/api/latest/internal/external-db-sync/sync-engine/route.tsx
```diff
@@ -0,0 +1,46 @@
...
+  }),
+  handler: async ({ body }, fullReq) => {
+    await ensureUpstashSignature(fullReq);
+
+    const { tenancyId } = body;
+
+    const tenancy = await getTenancy(tenancyId);
+    if (!tenancy) {
+console.warn(`[sync-engine] Tenancy ${tenancyId} in queue but not found.`);
+throw new StatusError(404, `Tenancy ${tenancyId} not found.`);
```
N2D4:
```suggestion
throw new StatusError(400, `Tenancy ${tenancyId} not found.`);
```

the resource (the sync engine) was found, just an argument passed on was not

apps/backend/prisma/migrations/20251125030551_external_db_sync/migration.sql
```diff
@@ -0,0 +1,234 @@
...
+ALTER TABLE "ProjectUser" ADD COLUMN  "sequenceId" BIGINT;
+
+-- SPLIT_STATEMENT_SENTINEL
+-- Creates unique indexes on sequenceId columns to ensure no duplicate sequence IDs exist.
+-- This guarantees each row has a unique position in the change sequence.
+CREATE UNIQUE INDEX  "ContactChannel_sequenceId_key" ON "ContactChannel"("sequenceId");
+
+-- SPLIT_STATEMENT_SENTINEL
+CREATE UNIQUE INDEX  "ProjectUser_sequenceId_key" ON "ProjectUser"("sequenceId");
+
```
N2D4, line 25: shouldn't we also have indexes on (tenancyId, sequenceId) for the sync-engine additionally to these ones?

apps/backend/prisma/migrations/20251125030551_external_db_sync/migration.sql
```diff
@@ -0,0 +1,234 @@
...
+    "data" JSONB,
+    "deletedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "startedFulfillingAt" TIMESTAMP(3),
+
+    CONSTRAINT "DeletedRow_pkey" PRIMARY KEY ("id")
+);
+
+-- SPLIT_STATEMENT_SENTINEL
+-- Creates indexes on DeletedRow table for efficient querying by sequence, table name, and tenant.
+CREATE UNIQUE INDEX "DeletedRow_sequenceId_key" ON "DeletedRow"("sequenceId");
```
N2D4, line 74: as above

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,320 @@
...
+  let didUpdate = false;
+  const projectUserTenants = await globalPrismaClient.$queryRaw<{ tenancyId: string }[]>`
+    WITH rows_to_update AS (
+      SELECT "tenancyId", "projectUserId"
+      FROM "ProjectUser"
+      WHERE "shouldUpdateSequenceId" = TRUE
+      OR "sequenceId" IS NULL
+      LIMIT 1000
+      FOR UPDATE SKIP LOCKED
+    ),
```
N2D4: would probably actually be a smart idea to ORDER BY tenancy_id here as well so the updates are less fragmented. (this requires us to change the index to (shouldUpdateSequenceId, tenancyId) alongside the (tenancyId, shouldUpdateSequenceId) as the order matters)

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,320 @@
...
+
+  const contactChannelTenants = await globalPrismaClient.$queryRaw<{ tenancyId: string }[]>`
+    WITH rows_to_update AS (
+      SELECT "tenancyId", "projectUserId", "id"
+      FROM "ContactChannel"
+      WHERE "shouldUpdateSequenceId" = TRUE
+      OR "sequenceId" IS NULL
+      LIMIT 1000
+      FOR UPDATE SKIP LOCKED
+    ),
```
N2D4: as above

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,320 @@
...
+
+  const deletedRowTenants = await globalPrismaClient.$queryRaw<{ tenancyId: string }[]>`
+    WITH rows_to_update AS (
+      SELECT "id", "tenancyId"
+      FROM "DeletedRow"
+      WHERE "shouldUpdateSequenceId" = TRUE
+      OR "sequenceId" IS NULL
+      LIMIT 1000
+      FOR UPDATE SKIP LOCKED
+    ),
```
N2D4: as above

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,320 @@
...
+        AND cc."projectUserId" = r."projectUserId"
+        AND cc."id"            = r."id"
+      RETURNING cc."tenancyId"
+    )
+    SELECT DISTINCT "tenancyId" FROM updated_rows
+  `;
+
+  for (const { tenancyId } of contactChannelTenants) {
+    assertUuid(tenancyId, "contactChannelTenants.tenancyId");
+    await enqueueExternalDbSync(tenancyId);
```
N2D4: as above

apps/backend/src/app/api/latest/internal/external-db-sync/sequencer/route.ts
```diff
@@ -0,0 +1,320 @@
...
+  }
+  if (deletedRowTenants.length > 0) {
+    didUpdate = true;
+  }
+
+  return didUpdate;
+}
+
+async function backfillSequenceIdsForTenancy(prisma: PrismaClientTransaction, tenancyId: string): Promise<boolean> {
```
N2D4: lots of duplicated code in here. although we can remove this entirely ifw e just don't support non-hosted  source-of-truth projects anymore (see below)

N2D4 (changes requested):
we should definitely do loadtesting here somehow, let's chat to see how

PR #1038, opened 2025-12-02
Clickhouse events

BilalG1:

### DIFF
--- a/apps/backend/src/lib/events.tsx
+++ b/apps/backend/src/lib/events.tsx
@@ -166,6 +166,8 @@ export async function logEvent<T extends EventType[]>(
   const branchId = typeof dataRecord === "object" && dataRecord && typeof dataRecord.branchId === "string" ? dataRecord.branchId : DEFAULT_BRANCH_ID;
   const userId = typeof dataRecord === "object" && dataRecord && typeof dataRecord.userId === "string" ? dataRecord.userId : "";
   const teamId = typeof dataRecord === "object" && dataRecord && typeof dataRecord.teamId === "string" ? dataRecord.teamId : "";
+  const sessionId = typeof dataRecord === "object" && dataRecord && typeof dataRecord.sessionId === "string" ? dataRecord.sessionId : "";
+  const isAnonymous = typeof dataRecord === "object" && dataRecord && typeof dataRecord.isAnonymous === "boolean" ? dataRecord.isAnonymous : false;
 
 
   // rest is no more dynamic APIs so we can run it asynchronously
@@ -204,6 +206,8 @@ export async function logEvent<T extends EventType[]>(
         branch_id: branchId,
         user_id: userId,
         team_id: teamId,
+        is_anonymous: isAnonymous,
+        session_id: sessionId,
       })),
       format: "JSONEachRow",
       clickhouse_settings: {
--- a/apps/backend/src/app/api/latest/internal/clickhouse/migrate-events/route.tsx
+++ b/apps/backend/src/app/api/latest/internal/clickhouse/migrate-events/route.tsx
@@ -1,0 +1,208 @@
+import { clickhouseAdminClient, createClickhouseClient } from "@/lib/clickhouse";
+import { DEFAULT_BRANCH_ID } from "@/lib/tenancies";
+import { globalPrismaClient } from "@/prisma-client";
+import { createSmartRouteHandler } from "@/route-handlers/smart-route-handler";
+import { yupNumber, yupObject, yupString } from "@stackframe/stack-shared/dist/schema-fields";
+import { StatusError } from "@stackframe/stack-shared/dist/utils/errors";
+import type { Prisma } from "@prisma/client";
+
+type Cursor = {
+  created_at_millis: number,
+  id: string,
+};
+
+const parseMillisOrThrow = (value: number | undefined, field: string) => {
+  if (typeof value !== "number" || !Number.isFinite(value)) {
+    throw new StatusError(400, `Invalid ${field}`);
+  }
+  const parsed = new Date(value);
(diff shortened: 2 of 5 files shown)


apps/backend/src/app/api/latest/internal/analytics/query/route.ts
```diff
@@ -42,7 +42,8 @@ export const POST = createSmartRouteHandler({
       query_id: queryId,
       query_params: body.params,
       clickhouse_settings: {
-        SQL_tenancy_id: auth.tenancy.id,
+        SQL_project_id: auth.tenancy.project.id,
+        SQL_branch_id: auth.tenancy.branchId,
```
N2D4, line 46: here i guess we should do something different for include_all_branches = true?

apps/backend/src/lib/tokens.tsx
```diff
@@ -164,6 +164,7 @@ export async function generateAccessTokenFromRefreshTokenIfValid(options: {
       projectId: options.tenancy.project.id,
       branchId: options.tenancy.branchId,
       userId: options.refreshTokenObj.projectUserId,
+      teamId: user.selected_team_id ?? "",
```
N2D4: is this necessary? if so, can we call it selectedTeamId?

apps/backend/src/lib/events.tsx
```diff
@@ -179,6 +193,29 @@ export async function logEvent<T extends EventType[]>(
       },
     });
 
+    const clickhouseClient = createClickhouseClient("admin");
```
N2D4: is it not a problem if we create a new clicnt every time?

apps/backend/src/lib/tokens.tsx
N2D4: if this was just to test the teamId column, i'd say we should probably skip it here

apps/backend/scripts/clickhouse-migrations.ts
```diff
@@ -5,20 +5,39 @@ export async function runClickhouseMigrations() {
...
+    data        JSON,
+    project_id  String,
+    branch_id   String,
+    user_id     String,
+    team_id     String,
+    created_at DateTime64(3, 'UTC') DEFAULT now64(3)
+)
+ENGINE MergeTree
+PARTITION BY toYYYYMM(event_at)
+ORDER BY (project_id, branch_id, event_at);
```
N2D4, line 46: maybe some more columns are useful here thinking forward for web analytics. refresh_token_id and session_id seems important, probably also is_impersonation (inferred from session_id). ip and geolocation also come to mind. i think it's better too have too many columns here than too few. chat has some more but not all of them make sense:







## **Request / Transport**





These help debug reality vs dashboards.



- **event_id (UUID / ULID)** – idempotency, deduping, late arrivals
- **received_at** – when _you_ got it (≠ client time)
- **client_timestamp** – raw client time (often lies; still useful)
- **latency_ms** – client → ingestion delta
- **sdk_name / sdk_version** – breaks explain themselves
- **protocol / transport** – fetch, beacon, websocket, edge, worker
- **ingest_source** – browser, server, edge, replay, backfill





---





## **Session / Navigation**





These unlock funnels and causality.



- **session_id** – your definition, not the browser’s
- **previous_event_id** – cheap causal graph
- **referrer** – raw + normalized
- **utm_source / medium / campaign / term / content**
- **entry_url / current_url**
- **route / pathname** – stripped of noise
- **navigation_type** – hard load, soft nav, bfcache restore





_bfcache (= back/forward cache; browser snapshot restore)_



---





## **Client / Environment**





This is where 80% of “why is this broken?” lives.



- **user_agent (raw)** – always keep the original
- **browser / browser_version**
- **os / os_version**
- **device_type** – mobile / desktop / tablet
- **device_brand / model** – mostly mobile, very useful
- **screen_width / height**
- **viewport_width / height**
- **pixel_ratio**
- **locale / timezone**





---





## **Network / Performance**





Gold for debugging slow or flaky users.



- **connection_type** – wifi / cellular / ethernet
- **effective_bandwidth** – eg 4g / 3g (if available)
- **rtt_ms** – rough but telling
- **cdn_pop / edge_region**
- **is_prerender / is_prefetch**





---





## **Identity / Auth State**





Not _who_ they are, but _how_ they’re known.



- **auth_state** – anonymous / logged_in / service
- **account_id / org_id / workspace_id** (in addition to team)
- **role / plan / tier** – feature gating explains behavior
- **impersonated_by** – support & admins





---





## **Experimentation / Flags**





Absolutely worth the bytes.



- **experiment_assignments** – map or compact encoding
- **feature_flags** – snapshot at event time
- **variant_hash** – cheap joins, fast group-bys





---





## **Trust / Data Quality**





You’ll thank yourself later.



- **is_bot / bot_reason**
- **is_sampled** – and **sample_rate**
- **is_replay / is_backfill**
- **ingestion_version** – schema + pipeline version
- **integrity_hash / signature** – tampering detection
- **consent_state** – gdpr/ccpa snapshot





---





## **App / Build Info**





Critical for regressions.



- **app_version** – semver or git sha
- **build_id** – deploy-level grouping
- **environment** – prod / staging / preview
- **release_channel** – canary / stable





---





## **Minimal but Powerful “Extras”**





If I had to pick a few “surprisingly high ROI” ones:



- **previous_url** (not just referrer)
- **client_clock_skew_ms**
- **experiment bucket hash**
- **build_id**
- **sdk_version**
- **is_bot**

N2D4 (approved):

apps/backend/src/app/api/latest/internal/analytics/query/route.ts
BilalG1: yeah, for now i'll just throw an error

PR #1070, opened 2025-12-16
Payments update user payment method

BilalG1:
Screenshot 2025-12-16 at 11 46 18 AM
Screenshot 2025-12-16 at 11 46 55 AM


packages/stack-shared/src/interface/client-interface.ts
N2D4, line 1: from our discussion: Payments tab should be shown/hidden based on whether the user has already interacted with payments before or not

PR #1094, opened 2026-01-09
[dashboard][fix] - Hide dev port banner in prod

madster456:
# Hide dev port banner in production

## Problem
The "PORT:81xx" development banner displays in production Docker deployments, even with `ENNVIRONMENT=production`.

## Fix:
Add `process.env.NODE_ENV === "production"` check to the existing condition:

```ts
if (!prefix || !isVisible || process.env.NODE_ENV === "production") return null;
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security by preventing development port information from being displayed in production environments.


N2D4:
just fixed this on dev

PR #1100, opened 2026-01-12
Update SDK dependencies to latest version and handle the breaking changes

nams1570:
We update the sdk dependencies (the ones present in `package-template.json`) to the latest versions. Since several packages have major version bumps, this results in a variety of breaking changes that have been handled here. Incidentally, when possible, we update similar dependencies across the codebase.
We decide to defer the tailwind update to another PR owing to its scale. The rest of the updates and changes have been catalogued below:

1. [Bumping](https://github.com/panva/oauth4webapi/blob/v3.x/CHANGELOG.md) `oauth4webapi` to 3.8.3: this was a major version changed. While there were no compatibility issues in the sdk, there were several breaking changes in `stack-shared`. Namely:
    a. The removal of `isOauth2Error`. We used this to check if the results of our `oauth4webapi` api invocations had issues. The functions were changed to explicitly throw either `ResponseBodyErrors` or `AuthorizationResponseErrors`, so the code was reworked to account for that with no loss in error handling.
    b. Dropping of support for http broadly: `oauth4webapi` now only accepts https. This is desired, but I add a carve out for our test environments only.
    c. `refreshTokenGrantRequest` and `authorizationCodeGrantRequest` now require `clientAuthentication` to be passed explicitly to them.
    d. Changes in how we handle our `MultiFactorAuthenticationRequired` error: This is an error that we created and is passed to the `oauth4webapi` API if there are MFA issues. Since the `processAuthorizationCodeResponse` now explicitly throws a `ResponseBodyError`, we access the error cause from the body of the error instead.
2. [Bumping](https://github.com/Qix-/color/releases) `color` to 5.0.4: this was a major version bump. Simple type checking change, I checked the API for the correct interface.
3. [Bumping](https://github.com/MasterKale/SimpleWebAuthn/blob/master/CHANGELOG.md) `simplewebauthn` to 13.2.2: two major version bumps, but no incompatibilities surprisingly
4. [Bumping](https://github.com/jshttp/cookie/releases) `cookie` to 1.1.1: this was a major version bump.
    a. Changing `parse` to `parseCookie`. In the most recent version, `parse` is still maintained as an alias for `parseCookie` for backwards compatibility, but I thought it would be best to change it over now. No change in functionality.
    b. Typing is now strongly enforced. A cookie can be `string | undefined`, and the `Cookies` are now `Record<string, string | undefined>`. We already have code to handle if a cookie is returned as undefined/ null, so the changes here were more to ensure type compatibility rather than big changes in functionality.
5. [Bumping ](https://github.com/isaacs/rimraf#readme)`rimraf` to 6.1.2: No breaking changes, mostly just bug fixes.
6. [Bumping](https://github.com/panva/jose/releases?page=1) `jose` to 6.1.3: This is another major version bump. We update it across the codebase to ensure compatibility. We use this for importing and processing jwk tokens. There are a few big changes in the version bump, but the only one that applies to us is that `importJwk` now yields a `CryptoKey` instead of a `KeyObject` in Node.js. However, this doesn't appear to break our code. We use `importJwk` in `stack-auth/packages/stack-shared/src/utils/jwt.tsx`.
7.  [Bumping](https://github.com/react-hook-form/resolvers/releases) `hookform/resolvers` to 5.2.2 (two major version jumps), and consequently bumping `react-hook-form` to 7.70.0: We already use the patterns that `hookform/resolvers`' latest versions seem to be enforcing. The only other breaking change is that it requires version 7.55.0+ of `react-hook-form`. Though we should pay attention to any interactions with zod and `hookform/resolvers`, some people have reported compatibility issues if they aren't using the latest compatible versions of both.
8. [Bumping](https://github.com/jquense/yup/blob/master/CHANGELOG.md) `yup` to 1.7.1: this was a minor version change, but we had incompatibility issues with this change. Versions 1.4.1 and 1.7.1 cannot exist in the same codebase due to incompatibility, so we bumped it up across the codebase, including in peer dependencies. 
9. Some minor version changes for some packages, but these were mostly bug fixes.
10. **Edited to add**: Bumping freestyle to 0.1.6, and reworking the freestyle mock server. In 0.1.6, freestyle changed their API in two ways:
    a. We're now supposed to hit their `execute/v2/...` endpoint and 
    b. They've flattened the `config` argument to `serverless.runs.create`.
These changes are minor, but are important. As part of a general suite of dependency bumps, this was judged to fit here.

We have linked the changelogs for the packages on each line.

## Summary by CodeRabbit

* **Chores**
  * Updated many dependency versions across the project for improved security, compatibility, and tooling.

* **Bug Fixes**
  * Hardened OAuth/token flows with stronger client auth, clearer error mapping (including MFA), and explicit handling for insecure test endpoints.
  * Improved cookie parsing and session token handling for more reliable sign-in/out behavior.

* **Other**
  * Minor backend and sandbox execution payload adjustments and expanded mock endpoint acceptance (no user-facing API changes).


packages/stack-shared/src/interface/client-interface.ts
```diff
@@ -190,17 +197,20 @@ export class StackClientInterface {
...
+      }
 
-    return AccessToken.createIfValid(result.access_token) ?? throwErr("Access token in fetchNewAccessToken is invalid, looks like the backend is returning an invalid token!", { result });
+      return AccessToken.createIfValid(result.access_token) ?? throwErr("Access token in fetchNewAccessToken is invalid, looks like the backend is returning an invalid token!", { result });
+    } catch (e){
+      if (e instanceof oauth.ResponseBodyError) {
+        throw new StackAssertionError("ResponseBodyError when processing refresh token response", { e });
+      }
+      throw new StackAssertionError("Unexpected error when processing refresh token response", { e });
+    }
```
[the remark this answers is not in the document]
nams1570: I think this ends up adding more boilerplate to solve a non-existent problem. We aren't going to get another StackAssertionError here, and if we did that's something we would want to know about.

packages/stack-shared/src/interface/client-interface.ts
```diff
@@ -1015,37 +1025,53 @@ export class StackClientInterface {
...
+    try {
+      result = await oauth.processAuthorizationCodeResponse(as, client, response);
+    } catch (e) {
+      if (e instanceof oauth.ResponseBodyError) {
+        if ((e.cause as any).code === "MULTI_FACTOR_AUTHENTICATION_REQUIRED") {
+          throw new KnownErrors.MultiFactorAuthenticationRequired((e.cause as any).details.attempt_code);
+        }
+        // TODO Handle OAuth 2.0 response body error
+        throw new StackAssertionError("Outer OAuth error during authorization code response", { cause: e.cause });
       }
```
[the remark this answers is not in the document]
nams1570: I think this ends up adding more boilerplate to solve a non-existent problem. We aren't going to get another StackAssertionError here, and if we did that's something we would want to know about.

packages/stack-shared/src/interface/client-interface.ts
```diff
@@ -1015,37 +1025,53 @@ export class StackClientInterface {
...
-      // TODO Handle OAuth 2.0 response body error
-      throw new StackAssertionError("Outer OAuth error during authorization code response", { result });
+      throw new StackAssertionError("Unexpected error when processing authorization code response", { cause: e });
```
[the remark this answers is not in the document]
nams1570: Not an issue. ResponseBodyError has the cause attribute and we already check for e being a ResponseBodyError. This issue will not happen

packages/stack-shared/src/interface/client-interface.ts
```diff
@@ -1015,37 +1025,53 @@ export class StackClientInterface {
...
+    let params: URLSearchParams;
+    try {
+      params = await this._networkRetryException(
+        async () => oauth.validateAuthResponse(as, client, options.oauthParams, options.state),
+      );
+    } catch (e) {
+      if (e instanceof oauth.AuthorizationResponseError) {
+        throw new StackAssertionError("Authorization response error when validating outer OAuth response", { e });
+      }
+      throw new StackAssertionError("Unexpected error when validating outer OAuth response", { e });
```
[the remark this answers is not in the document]
nams1570: This is a good catch. There's no real need to wrap `validateAuthResponse` in `_networkRetryException`, and that function needs some reworking anyway. Will fixup the commit and push a fix.

packages/stack-shared/src/interface/client-interface.ts
```diff
@@ -153,22 +154,28 @@ export class StackClientInterface {
...
-      token_endpoint: this.getApiUrl() + '/auth/oauth/token',
+      token_endpoint: tokenEndpoint,
     };
     const client: oauth.Client = {
       client_id: this.projectId,
       client_secret: this.options.publishableClientKey,
       token_endpoint_auth_method: 'client_secret_post',
     };
 
+    const clientAuthentication = oauth.ClientSecretPost(this.options.publishableClientKey);
```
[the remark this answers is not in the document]
nams1570: Minor, but good catch. Fixed in latest push

packages/stack-shared/src/interface/client-interface.ts
```diff
@@ -190,17 +196,20 @@ export class StackClientInterface {
...
+      }
 
-    return AccessToken.createIfValid(result.access_token) ?? throwErr("Access token in fetchNewAccessToken is invalid, looks like the backend is returning an invalid token!", { result });
+      return AccessToken.createIfValid(result.access_token) ?? throwErr("Access token in fetchNewAccessToken is invalid, looks like the backend is returning an invalid token!", { result });
+    } catch (e){
+      if (e instanceof oauth.ResponseBodyError) {
+        throw new StackAssertionError("ResponseBodyError when processing refresh token response", { e });
+      }
+      throw new StackAssertionError("Unexpected error when processing refresh token response", { e });
+    }
```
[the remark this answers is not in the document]
nams1570: good catch, handled in new push

packages/stack-shared/src/interface/client-interface.ts
```diff
@@ -1015,37 +1026,51 @@ export class StackClientInterface {
...
-      token_endpoint_auth_method: 'client_secret_post',
     };
-    const params = await this._networkRetryException(
-      async () => oauth.validateAuthResponse(as, client, options.oauthParams, options.state),
-    );
-    if (oauth.isOAuth2Error(params)) {
-      throw new StackAssertionError("Error validating outer OAuth response", { params }); // Handle OAuth 2.0 redirect error
+    const clientAuthentication = oauth.ClientSecretPost(this.options.publishableClientKey);
+    // Allow insecure HTTP requests only in test environment (for localhost testing)
+    const allowInsecure = getNodeEnvironment() === 'test' && tokenEndpoint.startsWith('http://');
```
N2D4, line 1045: the `http://` seems unnecessary, as allowInsecure already implies that it's allowed (not required), right?

N2D4:
```suggestion
    const allowInsecure = getNodeEnvironment() === 'test'
```

packages/stack-shared/src/interface/client-interface.ts
```diff
@@ -190,17 +196,22 @@ export class StackClientInterface {
...
-      throw new StackAssertionError("OAuth error", { result });
+    let result: oauth.TokenEndpointResponse;
+    try {
+      result = await oauth.processRefreshTokenResponse(as, client, response);
+    } catch (e){
+      if (e instanceof oauth.ResponseBodyError) {
+        throw new StackAssertionError("ResponseBodyError when processing refresh token response", { cause: e });
+      }
+      throw new StackAssertionError("Unexpected error when processing refresh token response", { cause: e });
     }
```
N2D4: usually when we wrap stuff in StackAssertionError we do so so we can add additionall logging information in the extraData argument (or the error message), but neither of these seem particularly helpful here. is there a specific reason you did it here?

packages/stack-shared/src/interface/client-interface.ts
```diff
@@ -1015,37 +1026,51 @@ export class StackClientInterface {
...
+
+    let params: URLSearchParams;
+    try {
+      params = oauth.validateAuthResponse(as, client, options.oauthParams, options.state);
+    } catch (e) {
+      if (e instanceof oauth.AuthorizationResponseError) {
+        throw new StackAssertionError("Authorization response error when validating outer OAuth response", { cause: e });
+      }
+      throw new StackAssertionError("Unexpected error when validating outer OAuth response", { cause: e });
     }
```
N2D4, line 1060: as above

N2D4 (approved):

packages/stack-shared/src/interface/client-interface.ts
nams1570: So the ResponseBodyError actually contains more information in its body (as per the [doc](https://github.com/panva/oauth4webapi/blob/main/docs/classes/ResponseBodyError.md)). It can also be a wrapper on their end for other errors that occurred during the function processing. Consequently, i thought logging it would give us access to that extra info, like the wrapped errors.

packages/stack-shared/src/interface/client-interface.ts
nams1570: Similar comment to above! AuthorizationResponseError also can serve as a wrapper over other errors.

packages/stack-shared/src/interface/client-interface.ts
nams1570: `allowInsecure` is the field we pass to the oauth4webapi API to get them to process our request. I check for `http` explicitly in case we want to add tests that send out https requests in the future, because for those tests we may not want to allow insecure/ may want oauth4webapi to flag it. Happy to remove it if it isn't likely to happen though!

packages/stack-shared/src/interface/client-interface.ts
N2D4: to do that you'll actually have to pass it

```suggestion
        throw new StackAssertionError("ResponseBodyError when processing refresh token response", { cause: e, ... });
      }
      throw new StackAssertionError("Unexpected error when processing refresh token response", { cause: e, ... });
    }
```

packages/stack-shared/src/interface/client-interface.ts
N2D4: in which case would it make a difference?

N2D4: either way doesn't matter, let's keep it

N2D4 (approved):

packages/stack-shared/src/interface/client-interface.ts
nams1570: Changes made in the recent push

PR #1112, opened 2026-01-16
[Docs] Update docs to openrouter usage for ai chat

madster456:
## Summary by CodeRabbit

* **New Features**
  * Switched AI provider for improved streaming and model selection.
  * Updated chat UI: smoother near-bottom scrolling, refined layout, and visual enhancements.

* **Documentation**
  * Removed legacy Stack Auth setup guide from docs.

* **Chores**
  * Updated docs dependencies and environment configuration; added new API key variable and removed an old global key.

### DIFF
--- a/docs/src/components/chat/ai-chat.tsx
+++ b/docs/src/components/chat/ai-chat.tsx
@@ -386,7 +386,7 @@ export function AIChatDrawer() {
         response: response,
         metadata: {
           sessionId: sessionId,
-          model: 'gemini-2.0-flash',
+          model: 'anthropic/claude-4.5-sonnet',
           temperature: 0,
         }
       };


docs/src/app/api/chat/route.ts
```diff
@@ -135,14 +135,15 @@ Remember: You're here to help users succeed with Stack Auth. Be helpful but conc
 
   try {
     const result = streamText({
-      model: google('gemini-2.5-flash'),
+      model: openrouter('anthropic/claude-3.5-sonnet'),
```
N2D4: shouldn't this be a newer model? this is very old

madster456: We can pick whatever model we want here, I just chose something that wasnt google, and had low cost to test on. 

What model would you like to see here?

PR #1119, opened 2026-01-19
fix sign in bug on dev

BilalG1:
## Summary by CodeRabbit

* **Refactor**
  * Tightened OAuth environment handling: non-HTTPS token requests are now permitted only in development or test runtimes and only when the token endpoint uses plain HTTP. Public APIs and signatures are unchanged; this restricts when insecure token exchanges are allowed without affecting normal end-user behavior.

### DIFF
--- a/apps/backend/src/app/api/latest/check-feature-support/route.tsx
+++ b/apps/backend/src/app/api/latest/check-feature-support/route.tsx
@@ -22,7 +22,11 @@ export const POST = createSmartRouteHandler({
     body: yupString().defined(),
   }),
   handler: async (req) => {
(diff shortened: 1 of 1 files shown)


packages/stack-shared/src/interface/client-interface.ts
```diff
@@ -166,7 +165,7 @@ export class StackClientInterface {
     };
 
     const clientAuthentication = oauth.ClientSecretPost(this.options.publishableClientKey);
-    const allowInsecure = getNodeEnvironment() === 'test' && tokenEndpoint.startsWith('http://');
+    const allowInsecure = process.env.NODE_ENV !== 'production' && tokenEndpoint.startsWith('http://');
```
N2D4: some environments use different names for the prod environments, such as `vercel-prod`
```suggestion
    const allowInsecure = (process.env.NODE_ENV.includes("dev") || process.env.NODE_ENV === 'test') && tokenEndpoint.startsWith('http://');
```

PR #1122, opened 2026-01-19
fix: localhost not loading

nams1570:
### Summary of Changes

The recent push to main fixed the issue with localhost not loading, but only for next.js environments. We don't have a guarantee that the sdk will be loaded in an environment with access to `process`. So, we move away from using environment variables to check for test/dev environments. We allow it to be configured via the sdk and also check for localhost explicitly (sort of a heuristic for development env).

## Summary by CodeRabbit

* **New Features**
  * Added optional configuration option to enable insecure HTTP requests for testing and development purposes (defaults to disabled for production safety).

* **Tests**
  * Added comprehensive test coverage for HTTP request security validation across localhost and non-localhost scenarios.

### DIFF
--- a/apps/backend/src/app/api/latest/check-feature-support/route.tsx
+++ b/apps/backend/src/app/api/latest/check-feature-support/route.tsx
@@ -22,7 +22,7 @@ export const POST = createSmartRouteHandler({
     body: yupString().defined(),
   }),
   handler: async (req) => {
-    const featureName = req.body?.feature_name;
+    const featureName = req.body?.name as unknown;
     const expectedUnsupportedFeatures = ["rsc-handler-signIn"];
     if (!expectedUnsupportedFeatures.includes(featureName)) {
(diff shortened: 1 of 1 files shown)


packages/stack-shared/src/interface/client-interface.ts
```diff
@@ -38,6 +39,12 @@ export type ClientInterfaceOptions = {
   extraRequestHeaders: Record<string, string>,
   projectId: string,
   prepareRequest?: () => Promise<void>,
+  /**
+   * Allows HTTP (non-HTTPS) requests to non-localhost servers.
+   * WARNING: Only use this for testing environments. Never enable in production.
+   * @default false
+   */
+  dangerouslyAllowInsecureHttp?: boolean,
```
N2D4, line 47: it's not that dangerous. it would still require us to have an HTTP base url... which we'd never do in production

keep in mind, tokenEndpoint is controlled by us, not an attacker

```suggestion
```

PR #1124, opened 2026-01-20
[Fix]: Add better and more informative error handling to email-rendering engines

nams1570:
### Summary of Changes

We ran into an error on sentry when the vercel fallback and freestyle improvements were pushed to prod. Spiking into the error handling revealed that the errors could be more informative to enable easier debugging.

We improve the error handling and add extra test coverage to cover the error pathways through the code. Note that we do not test vercel sandbox itself nor the fallback mechanism-this is because a) these will be logged and tested in prod with the sanity test code, and b) creating a mock vercel sandbox instance the way we have a mock freestyle server would just slow down any tests that pass through the email rendering pipeline, all for something thats meant to just be a fallback. However, locally, we tested with scripts and real vercel sandbox test project to success. Note that we also tried running the existing email-rendering test suite with fake freestyle credentials and real vercel-sandbox credentials (to mimic the fallback) and they passed.

## Summary by CodeRabbit

* **New Features**
  * Email preview now defaults to user/project when not provided and merges preview variables from templates.

* **Bug Fixes**
  * Error responses are consistently JSON-formatted and include stack traces plus richer execution context for clearer troubleshooting.
  * Batch processing: a single failing template now fails the entire batch and surfaces the specific template error.

* **Tests**
  * Expanded tests covering bundling/runtime errors, validation failures, JSON error formatting, preview behavior, and batch-failure scenarios.


N2D4 (changes requested):
as discussed

apps/backend/src/lib/js-execution.tsx
```diff
@@ -75,17 +75,32 @@ function createVercelSandboxEngine(): JsEngine {
...
+              result = { status: 'error', error: { message: 'Execution of innerCode returned undefined' } };
+            }
+            resultAsJson = JSON.stringify(result);
+          } catch (e) {
+            if(e instanceof Error) {
+              result = { status: 'error', error: { message: e.message, stack: e.stack, cause: e.cause} };
+            } else {
+              result = { status: 'error', error: { message: String(e), stack: undefined, cause: undefined} };
+            }
+            resultAsJson = JSON.stringify(result);
```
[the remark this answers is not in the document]
nams1570: Decided to simplify runner script in line with how freestyle runs their code. So this comment is moot.

apps/backend/src/lib/email-rendering.tsx
```diff
@@ -97,11 +100,23 @@ async function bundleAndExecute<T>(
...
+          freestyleError: executeResult.error,
+          vercelError: vercelResult.error,
+          innerCode: bundle.data,
+          innerOptions: [
+            { nodeModules, engine: 'freestyle' },
+            { nodeModules, engine: 'vercel-sandbox' },
+          ],
+        },
+      ));
+      return Result.error(JSON.stringify(vercelResult.error));
```
N2D4, line 115: why are we capturing an error here? this just means that the email was incorrect, no? or what exactly is the actionable item for us here if this happens?

N2D4: like wouldn't this trigger if someone's email template has a bug?

nams1570: This lets us compare the freestyle and vercel execute results in event of a failure, to see if they differ

PR #1138, opened 2026-01-28
Sign up rules

N2D4:
> [!NOTE]
> **High Risk**
> Touches core sign-up/auth flows and user restriction semantics (including new DB constraints) and introduces dynamic rule evaluation/logging; misconfiguration or CEL/parser bugs could block sign-ups or incorrectly restrict users.
> 
> **Overview**
> Introduces **CEL-based sign-up rules** (config-driven) that are evaluated during password/OTP/OAuth sign-ups and anonymous upgrades; matching rules can reject sign-ups or mark users as admin-restricted, and triggers are logged for analytics.
> 
> Extends `ProjectUser` with `restrictedByAdmin` plus public/private restriction details, updates restriction computation/filtering, and exposes these fields via user CRUD (including validation + DB constraint enforcing consistency when unrestricted).
> 
> Adds a new dashboard **Sign-up Rules** page with a visual condition builder (CEL <-> visual tree), drag-reorder by priority, per-rule 48h sparkline analytics via a new hidden internal endpoint, and adds user-page UI to view/edit manual restrictions. Also refactors ClickHouse client initialization to require env vars (removing `isClickhouseConfigured` checks) and adjusts CI container startup wait time.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2141e6893e8c1b72303b805e9234f996010d0880. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Sign-up Rules: visual rule builder, in-project CRUD with drag-reorder, per-rule analytics, backend evaluation, and admin UI.
  * Admin user restrictions: dashboard controls, banners/status, public/private admin details surfaced in user views.

* **APIs & Schema**
  * Config and user schemas extended; new SignUpRejected error and sign-up rule types added.

* **Tests**
  * Extensive unit and E2E coverage for rules, parser, evaluator, analytics, and restricted-user flows.

* **Docs**
  * Editorial guidance added to AGENTS.md.

* **Chores**
  * DB statement timeout, updated clean script, minor dependency additions.

### DIFF
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stackframe/stack-backend",
-  "version": "2.8.63",
+  "version": "2.8.64",
   "repository": "https://github.com/stack-auth/stack-auth",
   "private": true,
   "type": "module",
--- a/apps/backend/scripts/verify-data-integrity/index.ts
+++ b/apps/backend/scripts/verify-data-integrity/index.ts
@@ -79,6 +79,10 @@ async function main() {
   const shouldSkipNeon = flags.includes("--skip-neon");
   const recentFirst = flags.includes("--recent-first");
   const noBail = flags.includes("--no-bail");
+  const maxUsersPerProjectFlag = flags.find(f => f.startsWith("--max-users-per-project="));
+  const maxUsersPerProject = maxUsersPerProjectFlag
+    ? parseInt(maxUsersPerProjectFlag.split("=")[1], 10)
+    : Infinity;
 
   const { recurse, collectedErrors } = createRecurse({ noBail });
 
@@ -147,7 +151,9 @@ async function main() {
     console.warn("Using mock Stripe server (STACK_STRIPE_SECRET_KEY=sk_test_mockstripekey); skipping Stripe payout integrity checks.");
   }
 
-  const maxUsersPerProject = 100;
+  if (maxUsersPerProject !== Infinity) {
+    console.log(`Will check at most ${maxUsersPerProject} users per project.`);
+  }
 
   const endAt = Math.min(startAt + count, projects.length);
   for (let i = startAt; i < endAt; i++) {
@@ -157,16 +163,8 @@ async function main() {
         return;
       }
 
-      const [currentProject, users, projectPermissionDefinitions, teamPermissionDefinitions] = await Promise.all([
+      const [currentProject, projectPermissionDefinitions, teamPermissionDefinitions] = await Promise.all([
         expectStatusCode(200, `/api/v1/internal/projects/current`, {
-          method: "GET",
-          headers: {
-            "x-stack-project-id": projectId,
-            "x-stack-access-type": "admin",
-            "x-stack-development-override-key": getEnvVariable("STACK_SEED_INTERNAL_PROJECT_SUPER_SECRET_ADMIN_KEY"),
-          },
-        }),
-        expectStatusCode(200, `/api/v1/users?limit=${maxUsersPerProject}`, {
           method: "GET",
           headers: {
             "x-stack-project-id": projectId,
@@ -192,6 +190,30 @@ async function main() {
         }),
       ]);
       void currentProject;
+
+      // Fetch users with pagination
+      const PAGE_LIMIT = 1000;
+      const allUsers: any[] = [];
+      let cursor: string | undefined = undefined;
+      while (allUsers.length < maxUsersPerProject) {
+        const remainingToFetch = maxUsersPerProject - allUsers.length;
+        const limit = Math.min(PAGE_LIMIT, remainingToFetch);
+        const cursorParam: string = cursor ? `&cursor=${encodeURIComponent(cursor)}` : "";
+        const usersPage = await expectStatusCode(200, `/api/v1/users?limit=${limit}${cursorParam}`, {
+          method: "GET",
+          headers: {
+            "x-stack-project-id": projectId,
+            "x-stack-access-type": "admin",
+            "x-stack-development-override-key": getEnvVariable("STACK_SEED_INTERNAL_PROJECT_SUPER_SECRET_ADMIN_KEY"),
+          },
+        });
+        allUsers.push(...usersPage.items);
+        if (!usersPage.pagination?.next_cursor) {
+          break;
+        }
+        cursor = usersPage.pagination.next_cursor;
+      }
+      const users = { items: allUsers.slice(0, maxUsersPerProject) };
 
       const tenancy = await getSoleTenancyFromProjectBranch(projectId, DEFAULT_BRANCH_ID, true);
       const paymentsConfig = tenancy ? (tenancy.config as OrganizationRenderedConfig).payments : undefined;
--- a/apps/backend/src/app/health/email/route.tsx
+++ b/apps/backend/src/app/health/email/route.tsx
@@ -1,4 +1,5 @@
 import { createSmartRouteHandler } from "@/route-handlers/smart-route-handler";
+import { traceSpan } from "@/utils/telemetry";
 import { yupNumber, yupObject, yupString, yupTuple } from "@stackframe/stack-shared/dist/schema-fields";
 import { generateSecureRandomString } from "@stackframe/stack-shared/dist/utils/crypto";
 import { getEnvVariable, getNodeEnvironment } from "@stackframe/stack-shared/dist/utils/env";
@@ -62,29 +63,31 @@ const fetchFromResend = async (): Promise<{ data: ResendEmail[] }> => {
 };
 
 const performSignUp = async (email: string, password: string) => {
-  const apiBaseUrl = getEnvVariable("NEXT_PUBLIC_STACK_API_URL");
-  const response = await fetch(`${apiBaseUrl}/api/v1/auth/password/sign-up`, {
-    method: "POST",
-    headers: {
-      "Content-Type": "application/json",
-      "X-Stack-Access-Type": "client",
-      "X-Stack-Publishable-Client-Key": getEnvVariable("STACK_EMAIL_MONITOR_PUBLISHABLE_CLIENT_KEY"),
-      "X-Stack-Project-Id": getEnvVariable("STACK_EMAIL_MONITOR_PROJECT_ID"),
-    },
-    body: JSON.stringify({
-      email,
-      password,
-      verification_callback_url: getEnvVariable("STACK_EMAIL_MONITOR_VERIFICATION_CALLBACK_URL"),
-    }),
-  });
-
-  const responseBody = await response.text();
-
-  if (!response.ok) {
-    throw new StackAssertionError(`Sign-up failed: ${response.status} - ${responseBody}`, {
-      responseBody,
+  await traceSpan("performing sign-up", async () => {
+    const apiBaseUrl = getEnvVariable("NEXT_PUBLIC_STACK_API_URL");
+    const response = await fetch(`${apiBaseUrl}/api/v1/auth/password/sign-up`, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        "X-Stack-Access-Type": "client",
+        "X-Stack-Publishable-Client-Key": getEnvVariable("STACK_EMAIL_MONITOR_PUBLISHABLE_CLIENT_KEY"),
+        "X-Stack-Project-Id": getEnvVariable("STACK_EMAIL_MONITOR_PROJECT_ID"),
+      },
+      body: JSON.stringify({
+        email,
+        password,
+        verification_callback_url: getEnvVariable("STACK_EMAIL_MONITOR_VERIFICATION_CALLBACK_URL"),
+      }),
     });
-  }
+
+    const responseBody = await response.text();
+
+    if (!response.ok) {
+      throw new StackAssertionError(`Sign-up failed: ${response.status} - ${responseBody}`, {
+        responseBody,
+      });
+    }
+  });
 };
 
 const isExpectedVerificationEmail = (email: ResendEmail, testEmail: string): boolean => {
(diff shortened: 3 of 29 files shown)


apps/backend/src/lib/sign-up-rules.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (matches) {
+      const actionConfig = rule.action;
+      const actionType = actionConfig.type;
+      const action: SignUpRuleAction = {
+        type: actionType,
+        message: actionConfig.message,
+      };
+
+      // log asynchronously
+      runAsynchronouslyAndWaitUntil(logRuleTrigger(tenancy, ruleId, context, action));
```
nams1570, line 85: Echoing one of the other comments, if `logRuleTrigger`isn't exported (read: internal to this file), and the one call of it here doesn't pass/need a `userId`, should we consider nixing that argument?

apps/backend/src/lib/sign-up-rules.ts
```diff
@@ -0,0 +1,109 @@
...
+ * - Passkey signup
+ * - Anonymous user conversion (when anonymous user adds email/auth method)
+ *
+ * This function should NOT be called when creating anonymous users.
+ *
+ * @param tenancy - The tenancy to evaluate rules for
+ * @param context - The signup context with email, authMethod, etc.
+ * @returns The rule result with action to take
+ */
+export async function evaluateSignUpRules(
```
nams1570, line 53: nit: could we add an explicit return type to this function?

apps/backend/src/lib/users.tsx
```diff
@@ -1,9 +1,92 @@
...
+ *
+ * This function evaluates sign-up rules before creating/upgrading the user.
+ * Use this for all signup paths:
+ * - Password signup
+ * - OTP signup
+ * - OAuth signup
+ * - Passkey signup
+ * - Anonymous user conversion
+ *
+ * Do NOT use this for creating anonymous users (use createOrUpgradeAnonymousUser directly).
```
nams1570: I think this comment is outdated. Should we update this to say `use createOrUpgradeAnonymousUserWithoutRules`?

packages/stack-shared/src/config/schema.ts
```diff
@@ -562,6 +581,18 @@ const organizationConfigDefaults = {
...
       }),
     },
+    signUpRules: (key: string) => ({
+      enabled: false,
+      displayName: undefined,
+      priority: 0,
+      condition: undefined,
+      action: {
+        type: 'allow',
+        metadata: undefined,
```
nams1570: Should this metadata field be removed?it doesn't appear in the `yupObject` defined above, nor is it used anywhere.

apps/backend/prisma/migrations/20260201400001_add_restricted_by_admin_constraint/migration.sql
nams1570, line 1: Discussion: should it be possible to have `restrictedByAdmin` set to true without a `restrictedByAdminReason` or `restrictedByAdminPrivateDetails`? If we enforce this constraint as well, every restricted user is accounted for and project admins can come back and see whether it's worth lifting the restrictions later.

apps/backend/src/app/api/latest/users/crud.tsx
```diff
@@ -101,7 +101,8 @@ export function computeRestrictedStatus<T extends OnboardingConfig>(
   isAnonymous: boolean,
```
nams1570, line 102: <p class='graphite__hidden'><i>[Re: line +100]</i></p>
the comment right before this function definition is outdated. The comment states "A user is "restricted" if they've signed up but haven't completed onboarding requirements.", but a user can also be restricted by the admin now.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/stack-auth/stack-auth/1138?utm_source=unchanged-line-comment">Graphite</a>.</p>

apps/backend/src/app/api/latest/users/crud.tsx
```diff
@@ -101,7 +101,8 @@ export function computeRestrictedStatus<T extends OnboardingConfig>(
...
   primaryEmailVerified: boolean,
   config: T,
-): { isRestricted: false, restrictedReason: null } | { isRestricted: true, restrictedReason: { type: "anonymous" | "email_not_verified" } } {
+  restrictedByAdmin?: boolean,
+): { isRestricted: false, restrictedReason: null } | { isRestricted: true, restrictedReason: { type: "anonymous" | "email_not_verified" | "restricted_by_administrator" } } {
```
nams1570, line 106: nit: is it worth it to uplift this into a type?

apps/backend/src/app/api/latest/users/crud.tsx
```diff
@@ -101,7 +101,8 @@ export function computeRestrictedStatus<T extends OnboardingConfig>(
...
   // note: when you implement this function, make sure to also update the filter in the list users endpoint
```
nams1570, line 107: bug: As far as I can see, the list users endpoint wasnt updated (`onList` handler in `/users` route file). The `shouldFilterRestrictedByEmail` check doesn't actually check any of the new restricted conditions (this is also evinced by the comments in the crud handler file: "(restricted = signed up but email not verified)" which indicate that they're operating off of outdated logic).

apps/backend/src/app/api/latest/users/crud.tsx
```diff
@@ -1059,6 +1095,30 @@ export const usersCrudHandlers = createLazyProxy(() => createCrudHandlers(usersC
...
+        ? undefined
+        : (data.restricted_by_admin_private_details || null);
+
+      // Compute effective restricted flag considering existing value for PATCH updates
+      const effectiveRestrictedByAdmin = data.restricted_by_admin ?? oldUser.restrictedByAdmin;
+
+      if (!effectiveRestrictedByAdmin) {
+        // User is not (or will not be) restricted - reason/details must not be provided
+        if (restrictedByAdminReason != null) {
+          throw new StatusError(StatusError.BadRequest, "restricted_by_admin_reason requires restricted_by_admin=true");
```
nams1570, line 1116: nit: could this validation logic be lifted into a helper? considering it's duplicated between this and the preceding function?

apps/backend/src/app/api/latest/internal/sign-up-rules-stats/route.tsx
nams1570, line 1: Discussion: Every query to this endpoint queries clickhouse again. Is it worth caching the clickhouse responses?

Also, there is no limit on how many rules can be created. And each rule gets a 48 hour bucket associated with it when this endpoint is hit. In practice, our users will probably not make too many rules, but is this something we still want to try and ameliorate? Should we consider running a GROUP BY in the clickhouse query?

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx
```diff
@@ -0,0 +1,860 @@
...
+  const [enabled, setEnabled] = useState(rule?.enabled ?? true);
+  const [isSaving, setIsSaving] = useState(false);
+
+  // Parse existing condition or create empty group
+  const initialConditionTree = useMemo((): RuleNode => {
+    if (rule?.condition) {
+      const parsed = parseCelToVisualTree(rule.condition);
+      if (parsed) return parsed;
+    }
+    const group = createEmptyGroup('and');
```
Comment (nams1570) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx:145]: If we reach here, that means an existing condition that was created couldn't be parsed right? Should we notify the user somehow that this happened? Otherwise, silently, the condition will be wiped out.

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx
```diff
@@ -0,0 +1,860 @@
...
+      >
+        {/* Sparkline and trigger count */}
+        <div className="hidden sm:flex items-center mr-1">
+          <RuleSparkline
+            data={analytics?.hourlyCounts ?? []}
+            totalCount={analytics?.totalCount ?? 0}
+            isLoading={isAnalyticsLoading}
+          />
+        </div>
+        <Button
```
Comment (nams1570) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx:396]: should the edit button be disabled while saving?

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx
```diff
@@ -0,0 +1,860 @@
...
+function useSignUpRulesAnalytics() {
+  const stackAdminApp = useAdminApp();
+  const [analytics, setAnalytics] = useState<Map<string, RuleAnalytics>>(new Map());
+  const [isLoading, setIsLoading] = useState(true);
+
+  React.useEffect(() => {
+    let cancelled = false;
+    setIsLoading(true);
+
+    const fetchAnalytics = async () => {
```
Comment (nams1570) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx:511]: should we give users a retry button for fetching analytics? In case of transient errors? This is non blocking because users can always refresh.

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx
```diff
@@ -0,0 +1,860 @@
...
+                    ))}
+                  </SortableContext>
+                </DndContext>
+              </div>
+            </div>
+          )}
+
+          {/* Normal rules list (when no pending order) */}
+          {!hasOrderChanges && signUpRules.length > 0 && (
+            <DndContext
```
Comment (nams1570) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx:806]: isn't this duplicated code with 766-796?

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx
```diff
@@ -0,0 +1,860 @@
...
+      if (parsed) return parsed;
+    }
+    const group = createEmptyGroup('and');
+    group.children = [createEmptyCondition()];
+    return group;
+  }, [rule?.condition]);
+
+  const [conditionTree, setConditionTree] = useState<RuleNode>(initialConditionTree);
+
+  const handleSave = async () => {
```
Comment (nams1570) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx:155]: Is it worth it to do frontend CEL/ regex validation?

nams1570 (commented):
Overall logic is pretty solid, but there are some gaps in behavior I'm curious about. Test coverage looks good.

apps/backend/src/lib/sign-up-rules.ts
N2D4: yes, although do note this is controversial https://www.youtube.com/watch?v=I6V2FkW1ozQ

apps/backend/src/app/api/latest/users/crud.tsx
N2D4: is it used somewhere else? i don't think it is (but ai wrote this code lol so i'm not sure)

apps/backend/src/app/api/latest/users/crud.tsx
N2D4: excellent catch!

apps/backend/src/app/api/latest/users/crud.tsx
nams1570: `computeRestrictedStatus` is only used in this file I believe, though it's exported for some reason

apps/backend/src/app/api/latest/users/crud.tsx
N2D4: the entire create/update logic here is kinda duplicated in this endpoint, we should refactor the entire endpoint at some point using the new CudHandlers that @BilalG1 made, but for now i think we can keep it like this

apps/backend/src/app/api/latest/internal/sign-up-rules-stats/route.tsx
N2D4: > Discussion: Every query to this endpoint queries clickhouse again. Is it worth caching the clickhouse responses?

almost certainly not unless we have reason to believe not doing so would cause trouble, it would be a premature optimization

> Also, there is no limit on how many rules can be created. And each rule gets a 48 hour bucket associated with it when this endpoint is hit. In practice, our users will probably not make too many rules, but is this something we still want to try and ameliorate? Should we consider running a GROUP BY in the clickhouse query?

good thinking but i think we can also worry about this when it happens, it's relatively low-cost if it fails and we'll see on sentry if someone hits ratelimits (so we can fix it then)

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx
Comment (N2D4) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx:396]: probably but the UX is pretty bad in general right now, it's just one prompt, will probably have to rework it at some point

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx
Comment (N2D4) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx:511]: we should just not have transient errors (our clients should be prepared to deal with network etc. errors)

Comment (N2D4) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx:511]: generally better to have retry of course just in case (also so users don't get frustrated) but probably not worth worrying about the ux here i'd say... it's hard to put a retry button onto the small space it's in

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx
Comment (N2D4) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx:806]: yes but it's maintained by AI anyways haha

apps/backend/src/app/api/latest/users/crud.tsx
N2D4: i mean the type. why did you want to uplift it? just to separate things?

N2D4: i personally like having things inline if they're never reused

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx
Comment (N2D4) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sign-up-rules/page-client.tsx:145]: probably captureError because at least on the frontend we should never let this happen

apps/backend/src/app/api/latest/users/crud.tsx
nams1570: yes, sorry, to clarify I thought it would read better. But I agree that it's valid to have it inline especially if it isn't used in another file.

apps/backend/prisma/migrations/20260201400001_add_restricted_by_admin_constraint/migration.sql
N2D4: that's how i initially had it (making the internal details required), but that doesn't fit every use case. will make it so it's required in the frontend at least tho

nams1570: Yeah I think that's a decent compromise. Worth noting that any attempt to mutate the state via the SDK won't run through said validation though.

nams1570 (approved):
LGTM!

apps/backend/prisma/migrations/20260201400001_add_restricted_by_admin_constraint/migration.sql
N2D4: i think that's good though, it's a bit weird if i can't change the "restrictedByAdmin" property through SDK (or only change it to one value)

PR #1141, opened 2026-01-28
fix analytics queries

BilalG1:
## Summary by CodeRabbit

* **Improvements**
  * Added automatic retry with stricter result validation and clearer error handling for query profiling to improve reliability.

* **Chores**
  * CI workflows updated to use larger runner instances (upgraded runner size across relevant jobs).

### DIFF
--- a/.github/workflows/restart-dev-and-test-with-custom-base-port.yaml
+++ b/.github/workflows/restart-dev-and-test-with-custom-base-port.yaml
@@ -19,7 +19,6 @@ jobs:
     runs-on: ubicloud-standard-16
     env:
       NEXT_PUBLIC_STACK_PORT_PREFIX: "69"
-      CI: "1"
 
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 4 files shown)


apps/backend/src/lib/clickhouse.tsx
```diff
@@ -48,30 +48,44 @@ export const getQueryTimingStats = async (client: ClickHouseClient, queryId: str
...
-    },
-    format: "JSON",
-  });
+  const queryProfile = async () => {
+    const profile = await client.query({
+      query: `
+      SELECT
+        ProfileEvents['CPUTimeMicroseconds'] / 1000 AS cpu_time_ms,
+        ProfileEvents['RealTimeMicroseconds'] / 1000 AS wall_clock_time_ms
+      FROM clusterAllReplicas('default', system.query_log)
```
N2D4: good for now, but in the future we'll need a different solution

```suggestion
      // TODO: In the future, we'll need a different solution to make sure we always find the query log for the query that executed so we don't have to query all replicas
      FROM clusterAllReplicas('default', system.query_log)
```

apps/e2e/tests/backend/endpoints/api/v1/analytics-events.test.ts
```diff
@@ -91,7 +91,7 @@ it("cannot read events from other projects", async ({ expect }) => {
             "branch_id": "main",
             "event_type": "$token-refresh",
             "project_id": "<stripped UUID>",
-            "team_id": null,
+            "team_id": "",
```
N2D4: you may have to restart deps, this should definitely be null. empty strings are not good sentinel values

apps/backend/src/lib/clickhouse.tsx
N2D4: do we even need this? i guess we can remove it since time doesn't matter much for now

N2D4 (approved):

PR #1211, opened 2026-02-18
analytics: use anonymous users

BilalG1:


packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -443,14 +443,30 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+        try {
+          await this._signUpAnonymously();
+          const retrySession = await this._getSession();
+          const retryTokens = await retrySession.getOrFetchLikelyValidTokens(20_000, 75_000);
+          return retryTokens?.accessToken.token ?? null;
+        } catch {
+          return null;
+        }
+      }
+      return null;
```
N2D4: can't this just be `(await (await getUser({ or: "anomyous" })).getTokens()).accessToken`?  
  
although we should probably check whether analytics are enabled first — so we don't unnecessarily create an anonymous user for projects that don't have analytics enabled

PR #1220, opened 2026-02-24
Transactions rework + ledger playground

N2D4:
---

> [!NOTE]
> **High Risk**
> Touches core payments accounting/refund logic and changes transaction-entry semantics plus DB schema, so regressions could impact billing state, entitlement calculations, and Stripe sync behavior.
> 
> **Overview**
> Reworks payments to be *ledger-driven*: introduces `lib/payments/ledger` (transaction builder/pipeline + balance algorithm) and updates APIs to fetch transactions/owned products/item quantities from this shared ledger logic instead of bespoke per-table transaction builders.
> 
> Adds persistence for subscription lifecycle/default products: `Subscription` gains `endedAt` + `billingCycleAnchor`, and a new `DefaultProductsSnapshot` table is introduced and used to snapshot include-by-default products for correct historical calculations.
> 
> Adds internal-only tools: a hidden `ledger-playground` endpoint to run ledger functions over live/mock data with time-sliced snapshots, and a `resync-subscriptions` endpoint to re-pull subscription state from Stripe across projects. Also standardizes transaction entry `type` strings to kebab-case and updates cancellation/refund paths to set/use `endedAt` and delegate to `refundTransaction`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 05fe175846ae63a2eaa2b3c7ac99ff4193b1582d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added ledger playground interface for visualizing and debugging payment transactions
  * Added subscription resync capability to update Stripe account data
  * Enhanced subscription lifecycle tracking with end dates and billing cycle anchors
  * Implemented default product snapshots for improved multi-tenant product management
  * Improved item quantity and ownership calculations for customers

* **Chores**
  * Standardized transaction entry naming conventions
  * Refactored internal payment ledger computation system

### DIFF
--- a/apps/backend/src/lib/payments/ledger/implementation.tsx
+++ b/apps/backend/src/lib/payments/ledger/implementation.tsx
@@ -186,7 +186,7 @@ export async function getOwnedProductsForCustomer(options: {
           stripeSubscriptionId: subMeta?.stripeSubscriptionId ?? null,
           currentPeriodEnd: subMeta?.currentPeriodEnd ?? null,
           cancelAtPeriodEnd: subMeta?.cancelAtPeriodEnd ?? false,
-          isCancelable: entry.product_id !== null,
+          isCancelable: entry.subscription_id !== undefined,
         } : null,
       });
     }


apps/backend/src/lib/payments/ledger/transactions/list.ts
```diff
@@ -0,0 +1,82 @@
...
+  ) {}
+
+  getFirstCursor() { return ""; }
+  getLastCursor() { return ""; }
+
+  async next(opts: {
+    after: string,
+    limit: number,
+    filter: FullTransactionFilter,
+    orderBy: TransactionOrderBy,
```
nams1570: Why do we pass orderBy when it isn't used here?

apps/backend/src/lib/payments/ledger/transactions/list.ts
```diff
@@ -0,0 +1,82 @@
...
+    limitPrecision: "exact" | "approximate",
```
nams1570: Same with limitPrecision-why pass it when it's not being used? As far as I can see, this class doesn't inherit from any other class so it doesn't have a contract to fulfill

apps/backend/src/lib/payments/ledger/transactions/list.ts
```diff
@@ -0,0 +1,82 @@
...
+  }) {
+    const filterKey = JSON.stringify(opts.filter);
+    const allPromise = this.cache.get(filterKey) ?? buildTransactions(this.prisma, this.tenancyId, opts.filter);
+    this.cache.set(filterKey, allPromise);
+    const all = await allPromise;
+    const start = opts.after ? Math.max(0, all.findIndex((tx) => tx.id === opts.after) + 1) : 0;
+    const page = all.slice(start, start + opts.limit);
+    const cursor = page.length > 0 ? page[page.length - 1].id : opts.after;
+    return {
+      items: page.map((item) => ({ item, prevCursor: item.id, nextCursor: item.id })),
```
nams1570: Why return prevCursor and nextCursor for each item along with the item itself? I went through the codebase, and every use of `getTransactionsPaginatedList` or `BuiltTransactionsList` only uses  `entry.item` and never `entry.prevCursor` or `entry.nextCursor` (see `apps/backend/src/lib/payments/ledger/implementation.tsx`, `apps/backend/src/app/api/latest/internal/payments/transactions/route.tsx`.

nams1570: If it's to match the abstract class `PaginatedList`s shape, though we don't inherit from it here, I think the per-entry `prevCursor` and `nextCursor` of that class are also redundant.

apps/backend/src/lib/payments/ledger/transactions/queue.ts
```diff
@@ -0,0 +1,48 @@
...
+  push(value: T) {
+    this.heap.push({ seq: this.seq++, value });
+    this.bubbleUp(this.heap.length - 1);
+  }
+
+  pop(): T | null {
+    if (this.heap.length === 0) return null;
+    const root = this.heap[0];
+    const tail = this.heap.pop();
+    if (!tail) return root.value;
```
nams1570: nit: isn't this dead code? the check on line 11 ensures that the heap has at least 1 element. pop returns undefined only when the list is empty.

apps/backend/src/lib/payments/ledger/transactions/queue.ts
```diff
@@ -0,0 +1,48 @@
+export class EventQueue<T extends { at: number }> {
```
nams1570, line 2: Discussion: The current iteration is not thread-safe, but that should be ok for this PR. Currently in this PR, the eventqueue heap is only used in one area, a new one is initialized in buildTransactions every time it's run. This means that the heap is never used in a place that requires concurrent-safe operations.

apps/backend/src/lib/payments/ledger/transactions/list.ts
nams1570: especially since the buildTransactions code already sorts by `created_at` at the end

apps/backend/src/lib/payments/ledger/transactions/list.ts
```diff
@@ -0,0 +1,82 @@
...
+  };
+  const seedEvents = await fetchSeedEvents(prisma, tenancyId, filter);
+  for (const event of seedEvents) state.queue.push(event);
+
+  while (true) {
+    const event = state.queue.pop();
+    if (!event) break;
+    if (event.at > state.nowMillis) break;
+    processEvent(state, event);
+  }
```
nams1570, line 26: We've discussed this in person too, but since we build all transactions from t=0, the output array and the heap can end up being very very large. There's potential for a route that calls this function to timeout waiting for it, with potential for OOM in really extreme cases. We should figure out how to transition this to a more scalable solution sooner over later

apps/backend/src/lib/payments/ledger/transactions/helpers-core.ts
```diff
@@ -0,0 +1,147 @@
...
+  const [value, unit] = interval;
+  const date = new Date(baseMillis);
+  if (unit === "minute") date.setMinutes(date.getMinutes() + value);
+  else if (unit === "hour") date.setHours(date.getHours() + value);
+  else if (unit === "day") date.setDate(date.getDate() + value);
+  else if (unit === "week") date.setDate(date.getDate() + value * 7);
+  else if (unit === "month") date.setMonth(date.getMonth() + value);
+  else date.setFullYear(date.getFullYear() + value);
+  return date.getTime();
+}
```
nams1570, line 36: Using `TZ=America/Los_Angeles`, these cases happen:

- `2025-03-09 00:30` + `[1, "day"]` -> delta is **23 hours** (spring DST jump)
- `2025-11-02 00:30` + `[1, "day"]` -> delta is **25 hours** (fall DST rollback)
- `2025-01-31 12:00` + `[1, "month"]` -> lands on **Mar 3** (not Feb 28)
- `2025-03-31 12:00` + `[1, "month"]` -> lands on **May 1**
- Dec 31 + 1 month -> **Jan 31** (expected)

- Dec 31 + 2 months -> **Mar 3** (surprising rollover through February)

- Jan 30 + 1 month -> **Mar 2**

- Aug 31 + 1 month -> **Oct 1**

- Leap day Feb 29 + 1 year -> **Mar 1** next year

**  
**This is using the addInterval function defined here. So there's some odd behavior around daylight savings, and with months it counts 30 days rather than going to the effective next month end.   
  
Is there logic that could break? Or does this not matter?

apps/backend/src/lib/payments/ledger/transactions/processor.ts
```diff
@@ -0,0 +1,558 @@
...
+    }));
+    state.queue.push({
+      ...event,
+      at: addInterval(event.at, event.repeat),
+      items: nextItems,
+    });
+    return;
+  }
+
+  if (event.kind === "subscription-end-event") {
```
nams1570, line 438: Bug: This is for an edge case. Lets say a user of ours sets things up so that if you cancel your subscription within the first 5 days you get a refund. Their customer does so. The endedAt would be set at the time the subscription is cancelled right? but if a refund takes longer to process then the refundedAt would be set to a later date than the endedAt (endedAt < refundedAt). If this were to happen, then the subscription-end-event would be processed first and `state.activePurchases.delete(sourceKey);` would be run. Then, when the subscription-refund-event runs this block of code would be hit:

```typescript
 const source = state.activePurchases.get(sourceKey);
 if (!source) return;
```

In practice, this means the refund event would fail. In our codebase, refundedAt is set when this route `/internal/payments/transactions/refund/route.tsx ` is hit, and the cancel and refund actions aren't coupled together enough to deal with this case. Not that coupling is a good idea, just pointing out that it doesn't seem like we handle this.

apps/backend/src/lib/payments/ledger/transactions/processor.ts
```diff
@@ -0,0 +1,558 @@
...
+        if (nextQuantity > 0 && nextConfig) {
+          nextItems.set(itemId, {
+            quantity: nextQuantity,
+            repeat: nextConfig.repeat,
+            expiresWhenRepeated: nextConfig.expires === "when-repeated",
+            sourceTxId: txId,
+            grants: currentGrants,
+          });
+        }
+      }
```
nams1570, line 183: Potential bug: This case is also run when nextQuantity and prevQuantity are the same (i.e there's a new snapshot that doesn't change item quantity, just edits some metadata). The effect here is that the sourceTxId is reset to the new metadata-change-only txn.  
  
Lets say we have two default product snapshots- A and B. A occurs some time before B and actually changes the item quantities. B doesnt change any item quantities, it is pure metadata. A runs and queues up some default-product-item-grant-repeat-event events and so does B.  
  
When the default-product-item-grant-repeat-event event is processed, A's events are skipped because of this line of code:  
`if (itemState.sourceTxId !== event.causedByTxId) continue;`.  
So the timeline for repeated item grants/expiries now changes to B's timeline, even though A was where the items were initially added. This may be ok, but I feel like if we want to change the scheduling it should be an explicit field, not done by default.

apps/backend/src/lib/payments/ledger/implementation.tsx
```diff
@@ -0,0 +1,398 @@
...
+
+function normalizeDefaultSnapshot(snapshot: Record<string, any>): Record<string, any> {
+  const normalized: Record<string, any> = {};
+  for (const [productId, product] of Object.entries(snapshot)) {
+    normalized[productId] = normalizeDefaultSnapshotProduct(product as Record<string, unknown>);
+  }
+  return normalized;
+}
+
+function getCurrentDefaultProducts(transactions: Transaction[]): { snapshot: Record<string, any>, createdAtMillis: number } {
```
nams1570: Discussion: one thing to note about this function is that it only returns the first snapshot it finds. It relies implicitly on the txns being ordered before they're passed to this function. This may be fine, but it's an assumption we should probably comment on at least

apps/backend/src/lib/payments/ledger/implementation.tsx
```diff
@@ -0,0 +1,398 @@
...
+      const effectiveQuantity = Math.max(0, typedEntry.quantity - revoked);
+      if (effectiveQuantity <= 0) continue;
+
+      const isSubscription = typedEntry.subscription_id != null;
+      const isOneTime = typedEntry.one_time_purchase_id != null;
+      const subMeta = isSubscription ? subscriptionMetadata.get(typedEntry.subscription_id) : null;
+
+      ownedProducts.push({
+        id: typedEntry.product_id,
+        type: isSubscription ? "subscription" : isOneTime ? "one_time" : "include-by-default",
```
nams1570: We have explicit logic for including default products already later in this func. Why do we have/need fallback logic to include-by-default?

nams1570: As far as I can tell, default product related events never cause transactions with `product-grant` txn entries. They add only `default-product-item-grant / default-product-item-change / default-product-item-expire/ default-products-change` events.  
  
Only `subscription-start-event` and `one-time-purchase` events add/use `product-grant` txn entries. If you somehow have a product-grant entry that isnt a one time purchase or subscription, falling back to include-by-default would mask false/errorneous/ impossible states imo.

apps/backend/src/lib/payments/ledger/implementation.tsx
```diff
@@ -0,0 +1,398 @@
...
+  // Build revocation map — keyed by transaction ID only, since entry index
+  // is unreliable (subscription-start entry order varies depending on whether
+  // money_transfer is present)
+  const revokedQuantities = new Map<string, number>();
+  for (const tx of transactions) {
+    for (const entry of tx.entries) {
+      const typedEntry = entry as any;
+      if (typedEntry.type === "product-revocation") {
+        const key = typedEntry.adjusted_transaction_id;
+        revokedQuantities.set(key, (revokedQuantities.get(key) ?? 0) + typedEntry.quantity);
```
nams1570: Bug: `revokedQuantities` should probably be keyed to txn entries, not to txns. No conceptual reason why a txn can't have multiple product_grant entries. This would cause issues with the `effectiveQuantity` calculations later

apps/backend/src/lib/payments/ledger/implementation.tsx
```diff
@@ -0,0 +1,398 @@
...
+  const { snapshot: defaultProducts, createdAtMillis: defaultsCreatedAtMillis } = getCurrentDefaultProducts(transactions);
+  const productLines = options.tenancy.config.payments.productLines;
+
+  for (const [productId, product] of Object.entries(defaultProducts)) {
+    const productLineId = (product as any).product_line_id;
+
+    if (productLineId && productLineId in productLines) {
+      const hasConflict = ownedProducts.some((owned) =>
+        owned.product.product_line_id === productLineId &&
+        owned.type !== "include-by-default"
```
nams1570, line 211: How would we end up in a position where there's a default product in ownedProduct? As I mentioned in the above comment, product-grants are not associated with any default products, so im not sure when this would be triggered. Even if that somehow happened, wouldn't this check actually cause double-counting of a default product?

apps/backend/src/lib/payments/ledger/implementation.tsx
```diff
@@ -0,0 +1,398 @@
...
+  };
+
+  // Process item-level entries chronologically.
+  for (let ti = 0; ti < transactions.length; ti++) {
+    const tx = transactions[ti];
+    for (let ei = 0; ei < tx.entries.length; ei++) {
+      const entry = tx.entries[ei];
+      const typedEntry = entry as any;
+      if (typedEntry.type === "product-grant") {
+        paidGrantsByTx.set(tx.id, { productId: typedEntry.product_id ?? null, product: typedEntry.product, quantity: typedEntry.quantity });
```
nams1570: Potential bug: I think this shouldn't be keyed by tx id no? Otherwise a tx with multiple product grants would overwrite the field repeatedly

apps/backend/src/lib/payments/ledger/implementation.tsx
```diff
@@ -0,0 +1,398 @@
...
+  itemId: string,
+  customerId: string,
+  customerType: "user" | "team" | "custom",
+  now?: Date,
+}) {
+  await ensureDefaultProductsSnapshotUpToDate(options.prisma, options.tenancy);
+  const allTransactions = await getAllTransactionsForCustomer(options.prisma, options.tenancy, options.customerType, options.customerId);
+
+  const now = options.now ?? new Date();
+  const transactions = allTransactions.filter((tx) => tx.effective_at_millis <= now.getTime()).sort((a, b) => a.effective_at_millis - b.effective_at_millis);
```
nams1570: Discussion: is it possible for two txns to have the same `effective_at_millis`? if so, what is the behavior? Would the sorting not being deterministic cause issues, perhaps with which default snapshot is being used for future transaction parsing?

apps/backend/src/lib/payments/ledger/implementation.tsx
```diff
@@ -0,0 +1,398 @@
...
+        }
+      } else if (typedEntry.type === "default-products-change") {
+        defaultSnapshot = normalizeDefaultSnapshot(typedEntry.snapshot as Record<string, any>);
+      } else if (typedEntry.type === "item-quantity-change") {
+        if (typedEntry.item_id !== options.itemId) continue;
+        const expiresAtMillis: number | null | undefined = typedEntry.expires_at_millis;
+        const ledgerTransaction = {
+          amount: typedEntry.quantity,
+          grantTime: new Date(tx.effective_at_millis),
+          expirationTime: expiresAtMillis ? new Date(expiresAtMillis) : new Date(8640000000000000),
```
nams1570: Can expiresAtMillis be set to 0?

apps/backend/src/lib/payments/ledger/refund.ts
```diff
@@ -0,0 +1,220 @@
...
+) {
+  const subscription = await prisma.subscription.findUnique({
+    where: { tenancyId_id: { tenancyId: tenancy.id, id: subscriptionId } },
+  });
+  if (!subscription) throw new KnownErrors.SubscriptionInvoiceNotFound(subscriptionId);
+  if (subscription.refundedAt) throw new KnownErrors.SubscriptionAlreadyRefunded(subscriptionId);
+
+  const product = subscription.product as InferType<typeof productSchema>;
+  const includedItemCount = Object.keys(product.includedItems).length;
+  const subProductGrantIndex = 2;
```
nams1570, line 109: Why is this hardcoded? Do we know it's always this?

nams1570: Maybe switch to a helper function that gets the right product grant index? Otherwise isn't it too coupled

apps/backend/src/lib/payments/ledger/refund.ts
```diff
@@ -0,0 +1,220 @@
...
+  prisma: Awaited<ReturnType<typeof getPrismaClientForTenancy>>,
+  tenancy: Tenancy,
+  subscriptionId: string,
+  refundEntries: RefundEntrySelection[],
+) {
+  const subscription = await prisma.subscription.findUnique({
+    where: { tenancyId_id: { tenancyId: tenancy.id, id: subscriptionId } },
+  });
+  if (!subscription) throw new KnownErrors.SubscriptionInvoiceNotFound(subscriptionId);
+  if (subscription.refundedAt) throw new KnownErrors.SubscriptionAlreadyRefunded(subscriptionId);
```
nams1570, line 105: Wouldn't this block multiple partial refunds? Every refund that occurs will set the refundedAt if the db call went through. This would block any future refund requests

apps/backend/src/lib/payments/ledger/refund.ts
nams1570, line 1: We should probably think about how we want to handle multiple partial refunds and try to work that into this file

apps/backend/src/lib/payments/ledger/transactions/queue.ts
N2D4: javascript is a single-threaded language. there are no threads, and as such everything is thread-safe (minus some sharedarraybuffer shenanigans but we're not doing any of that)

apps/backend/src/app/api/latest/payments/payment-method/[customer_type]/[customer_id]/set-default/route.ts
```diff
@@ -1,4 +1,4 @@
-import { ensureClientCanAccessCustomer, ensureStripeCustomerForCustomer, getDefaultCardPaymentMethodSummary } from "@/lib/payments";
+import { ensureClientCanAccessCustomer, ensureStripeCustomerForCustomer, getDefaultCardPaymentMethodSummary } from "@/lib/payments/index";
```
Comment (nams1570) [inline apps/backend/src/app/api/latest/payments/payment-method/[customer_type]/[customer_id]/set-default/route.ts:1]: nit: why do we need to import from index here? Isn't the point of a module that you can import from it directly?

apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/[product_id]/route.ts
```diff
@@ -1,13 +1,14 @@
-import { ensureProductIdOrInlineProduct, getOwnedProductsForCustomer } from "@/lib/payments";
+import { SubscriptionStatus } from "@/generated/prisma/client";
+import { getOwnedProductsForCustomer } from "@/lib/payments";
+import { ensureProductIdOrInlineProduct } from "@/lib/payments/index";
```
Comment (nams1570) [inline apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/[product_id]/route.ts:3]: nit: same question here and in other places about why we import from index

apps/backend/src/app/api/latest/team-invitations/accept/verification-code-handler.tsx
```diff
@@ -75,6 +75,8 @@ export const teamInvitationCodeHandler = createVerificationCodeHandler({
     }
     const prisma = await getPrismaClientForTenancy(tenancy);
 
+    // HACK: Hardcode team member limit for internal project
```
nams1570, line 78: nit: are these comments meant to stay unimplemented for now?

apps/backend/src/lib/payments/implementation.tsx
```diff
@@ -744,23 +485,25 @@ export async function grantProductToCustomer(options: {
     quantity,
   });
 
-  if (conflictingProductLineSubscriptions.length > 0) {
-    const conflicting = conflictingProductLineSubscriptions[0];
-    if (conflicting.stripeSubscriptionId) {
+  const conflictingSubscriptions = purchaseContext.conflictingProducts.filter((p) => p.type === "subscription");
+  if (conflictingSubscriptions.length > 0) {
+    const conflicting = conflictingSubscriptions[0];
```
nams1570, line 490: Why pick just the first conflicting subscription here?

apps/backend/src/lib/payments/implementation.tsx
```diff
@@ -593,6 +345,7 @@ export function productToInlineProduct(product: ProductWithMetadata): yup.InferT
   return {
```
nams1570, line 345: <p class='graphite__hidden'><i>[Re: line +296]</i></p>
Are we sure this function is concurrent-safe? If two requests invoke this function at the same time, you can end up with two `stripe.customers.create` calls. I don't know if it throws on multiple calls with the same metadata.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/stack-auth/stack-auth/1220?utm_source=unchanged-line-comment">Graphite</a>.</p>

apps/backend/src/lib/payments/ledger/algo.ts
```diff
@@ -0,0 +1,47 @@
...
+    timeSet.add(grantTime);
+  }
+  const times = Array.from(timeSet.values()).sort((a, b) => a - b);
+  if (times.length === 0) {
+    return 0;
+  }
+
+  let grantedSum = 0;
+  let expiredSum = 0;
+  let usedSum = 0;
```
nams1570: What's the point in this var again? Doesn't seem to be used anywhere

nams1570 (commented):
Identified a few potential bugs. Broadly, we should think about what happens when a stripe call succeeds while a db call fails and vice versa. How do we keep them in sync?  
Also, can we trust stripe calls to throw on duplicate requests always? For the refund.create call, I know that it throws if a refund has been asked for a charge which has already been refunded to the original extent for example, but I'm not sure about other stripe calls.  
There's an interesting page on ensuring idempotency with stripe operations that may be worth taking a look at: https://docs.stripe.com/api/idempotent_requests?api-version=2024-11-20.acacia

apps/backend/src/lib/payments/ledger/algo.ts
nams1570: usedSum I mean

apps/backend/src/lib/payments/ledger/implementation.tsx
N2D4: the getOwnedProductsForCustomer function?

nams1570: No, `getCurrentDefaultProducts`

apps/backend/src/lib/payments/ledger/implementation.tsx
nams1570: TIEBREAK BY ID

apps/backend/src/lib/payments/ledger/implementation.tsx
nams1570: change to == null

apps/backend/src/lib/payments/ledger/refund.ts
nams1570: Add a comment here to explain why it's always 2. Just used for assertions for now so it's ok

apps/backend/src/lib/payments/implementation.tsx
nams1570: we should cancel all the conflicting subscriptions from the purchase context. Purchase context is "hey if i buy this, what would the conflicting subscriptions be?"

apps/backend/src/lib/payments/implementation.tsx
nams1570: try to find a fix if it isn't too hacky

nams1570:
> Identified a few potential bugs. Broadly, we should think about what happens when a stripe call succeeds while a db call fails and vice versa. How do we keep them in sync? Also, can we trust stripe calls to throw on duplicate requests always? For the refund.create call, I know that it throws if a refund has been asked for a charge which has already been refunded to the original extent for example, but I'm not sure about other stripe calls. There's an interesting page on ensuring idempotency with stripe operations that may be worth taking a look at: https://docs.stripe.com/api/idempotent_requests?api-version=2024-11-20.acacia

Generally we should log a sentry error when stripe succeeds and db fails and vice versa, that way it can be manually fixed.

apps/backend/src/lib/payments/ledger/implementation.tsx
nams1570: Handled by failing loud in future commit

apps/backend/src/lib/payments/ledger/implementation.tsx
nams1570: There's actually an interesting edge case here. If the user has multiple default products within the same product line, we don;t handle that well. Tried to fix this in a followup commit

apps/backend/src/lib/payments/ledger/implementation.tsx
nams1570: Yeah, this was a bug. Cleaned up in followup commits.

apps/backend/src/app/api/latest/internal/payments/transactions/route.tsx
```diff
@@ -33,180 +24,29 @@ export const GET = createSmartRouteHandler({
...
-      if (r.source === "subscription") lastSubId = r.id;
-      if (r.source === "item_quantity_change") lastIqcId = r.id;
-      if (r.source === "one_time") lastOtpId = r.id;
-      if (r.source === "subscription-invoice") lastSiId = r.id;
-    }
-
-    const nextCursor = page.length === limit
-      ? [lastSubId, lastIqcId, lastOtpId, lastSiId].join('|')
-      : null;
+    const start = query.cursor ? Math.max(0, all.findIndex((tx) => tx.id === query.cursor) + 1) : 0;
```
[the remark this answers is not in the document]
nams1570: check latest commits

apps/backend/src/lib/payments/ledger/implementation.tsx
```diff
@@ -0,0 +1,372 @@
...
+  for (const [productId, product] of Object.entries(snapshot)) {
+    normalized[productId] = normalizeDefaultSnapshotProduct(product as Record<string, unknown>);
+  }
+  return normalized;
+}
+
+function getCurrentDefaultProducts(transactions: Transaction[]): { snapshot: Record<string, any>, effectiveAtMillis: number } {
+  const sortedTransactions = [...transactions].sort((a, b) => a.effective_at_millis - b.effective_at_millis);
+
+  for (const tx of transactions) {
```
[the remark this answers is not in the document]
nams1570: check latest commits

apps/backend/src/lib/payments/ledger/implementation.tsx
```diff
@@ -0,0 +1,372 @@
...
+  };
+
+  // Process item-level entries chronologically.
+  for (let ti = 0; ti < transactions.length; ti++) {
+    const tx = transactions[ti];
+    for (let ei = 0; ei < tx.entries.length; ei++) {
+      const entry = tx.entries[ei];
+      const typedEntry = entry as any;
+      if (typedEntry.type === "product-grant") {
+        paidGrantsByTx.set(tx.id, { productId: typedEntry.product_id ?? null, product: typedEntry.product, quantity: typedEntry.quantity });
```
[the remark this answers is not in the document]
nams1570: check latest commits

apps/backend/src/app/api/latest/internal/payments/transactions/route.tsx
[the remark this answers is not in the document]
nams1570: check latest commits

apps/backend/src/lib/payments/ledger/implementation.tsx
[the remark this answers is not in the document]
nams1570: check latest commits

apps/backend/src/lib/payments/ledger/implementation.tsx
```diff
@@ -0,0 +1,372 @@
...
+    for (const entry of tx.entries) {
+      if ((entry.type as string) === "default-products-change") {
+        return {
+          snapshot: normalizeDefaultSnapshot((entry as any).snapshot as Record<string, any>),
+          effectiveAtMillis: tx.effective_at_millis,
+        };
+      }
+    }
+  }
+  return { snapshot: {}, effectiveAtMillis: 0 };
```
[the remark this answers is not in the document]
nams1570: check latest commits

apps/backend/src/lib/payments/ledger/implementation.tsx
```diff
@@ -0,0 +1,380 @@
...
+  };
+
+  // Process item-level entries chronologically.
+  for (let ti = 0; ti < transactions.length; ti++) {
+    const tx = transactions[ti];
+    for (let ei = 0; ei < tx.entries.length; ei++) {
+      const entry = tx.entries[ei];
+      const typedEntry = entry as any;
+      if (typedEntry.type === "product-grant") {
+        paidGrantsByTx.set(tx.id, { productId: typedEntry.product_id ?? null, product: typedEntry.product, quantity: typedEntry.quantity });
```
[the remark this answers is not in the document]
nams1570: check latest commits

PR #1330, opened 2026-04-13
local emulator build improvements

BilalG1:
… V8 --jitless

2.6 GB to 1.3 GB final image

Flip arm64 matrix back to ubicloud-standard-8 so both arches share one runner fleet. Cross-arch TCG on an amd64 host previously SIGTRAP'd in migrations because V8's JIT emitted arm64 instructions that QEMU's cross-arch translator mis-handled; pair the existing -cpu cortex-a72 fallback with NODE_OPTIONS=--jitless on the migration docker exec to force V8 to stay on the interpreter. Does not affect amd64 migrations (KVM, no TCG).

## Summary by CodeRabbit

* **Chores**
  * Optimized emulator images with binary stripping, compression, and preservation of standalone runtime dependencies.
  * Improved multi-architecture build matrix, added optional KVM detection/fallback, and gated certain emulator runtime steps for arm64.
  * Enhanced build scripts to generate and include env files and persist richer logs and artifacts.

* **New Features**
  * Centralized provision entrypoint to streamline install → migrations → slimming sequence.

* **Tests**
  * Added a fast QEMU serial boot test for architecture validation.


N2D4 (approved):
HYPE

PR #1350, opened 2026-04-18
[Refactor] [Fix] Remove default prod creation

nams1570:
With the new bulldozer rework we dont support default products anymore. Users are encouraged to currently manually handle granting products to their end users.

We block api requests and new product creations that attempt to set no price, and we remove any options to set include-by-default. We also migrate users' existing product snapshots in `Subscriptions`, `OneTimePurchases`, and `ProductVersions` to have no price set if it's an include-by-default product. This will make it so that next time a user goes onto their products page, they will be informed that the pricing is invalid and it is no longer delivered by default.

Note, however, that these products will still be providing items and the like to the users who have them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Migrated legacy product snapshots so missing included-items no longer break readers.
  * Removed deprecated "include-by-default" pricing sentinel; pricing now requires explicit price entries and write validation rejects the old sentinel.

* **Chores**
  * Simplified dashboard pricing flows: create/edit/save now use explicit prices and surface an alert when a formerly implicit free plan needs an explicit $0 price.
  * Config overrides and stored data are auto-normalized to explicit price objects.

* **Tests**
  * Updated and added tests covering migration, validation, and switching behavior for explicit prices.


mantrakp04:
@greptile-ai review

nams1570:
@greptileai please do a full rereview and refresh your summary. Note the PR is now based off of dev, so some of the switch logic to handle certain bugs was removed when I cherry picked things.

nams1570:
@greptileai just made some changes, please do a proper rereview

nams1570:
@greptileai please rereview

nams1570:
@greptileai some more changes, please do a proper rereview. Also take a look at the logic of the migration- there should be no include-by-default products anymore, but every product should have a price.

N2D4 (approved):

nams1570:
@greptileai do a proper rereview, and try to spot issues with the migration file or if it's still possible to set include-by-default as the price or if any users will be left in an improper state

PR #1356, opened 2026-04-20
fix(email-queue): recover stuck sending without duplicate retry

mantrakp04:
## Summary

Email outbox rows can get stuck in `SENDING` if a worker dies after setting `startedSendingAt` but before finishing or unclaiming. This change adds `recoverEmailsStuckInSending`, which runs each email queue step and marks rows past the stuck timeout as **terminal server errors** with delivery status unknown, **without** scheduling an automatic retry (to avoid duplicate sends if the provider already accepted the message).

## Changes

- **`recoverEmailsStuckInSending`**: updates stuck rows with `finishedSendingAt`, `canHaveDeliveryInfo: false`, and server error fields; emits Sentry via `captureError` when any rows are recovered.
- **Tests**: `email-queue-step.test.tsx` covers recovery of old `startedSendingAt`, no-op for recent sends, and idempotency (second pass does not re-queue).

## Test plan

- [ ] `pnpm` / vitest for `apps/backend/src/lib/email-queue-step.test.tsx` (requires dev DB like other integration tests in this package)

## Summary by CodeRabbit

* **Bug Fixes**
  * Email reliability: messages that remained stuck in sending are now automatically marked as terminal failures, assigned standardized error details, cleared from retry scheduling, prevented from receiving delivery info, and recovery emits an alert only when actual work occurs. Recovery is safe to run repeatedly (idempotent).

* **Tests**
  * Added integration tests validating recovery behavior, proper field updates, and idempotency.

### DIFF
--- a/apps/backend/src/lib/email-queue-step.tsx
+++ b/apps/backend/src/lib/email-queue-step.tsx
@@ -175,6 +175,7 @@ async function recoverEmailsStuckInSending(): Promise<void> {
       sendServerErrorExternalDetails: externalDetails,
       sendServerErrorInternalMessage: `Email was stuck in sending (startedSendingAt <= ${stuckCutoff.toISOString()}) and was recovered by the email queue step. Not retried to avoid duplicate delivery.`,
       sendServerErrorInternalDetails: internalDetails,
+      nextSendRetryAt: null,
       shouldUpdateSequenceId: true,
     },
     select: { id: true, tenancyId: true, startedSendingAt: true, to: true, sendAttemptErrors: true },
--- a/apps/backend/src/lib/email-queue-step.tsx
+++ b/apps/backend/src/lib/email-queue-step.tsx
@@ -97,7 +97,7 @@ export const runEmailQueueStep = withTraceSpan("runEmailQueueStep", async () => 
 
   const sendPlan = await withTraceSpan("runEmailQueueStep-prepareSendPlan", prepareSendPlan)(deltaSeconds);
   await withTraceSpan("runEmailQueueStep-processSendPlan", processSendPlan)(sendPlan);
-  await withTraceSpan("runEmailQueueStep-logEmailsStuckInSending", logEmailsStuckInSending)();
+  await withTraceSpan("runEmailQueueStep-recoverEmailsStuckInSending", recoverEmailsStuckInSending)();
   const sendEnd = performance.now();
 
   if (sendPlan.length > 0 || queuedCount > 0 || pendingRender.length > 0) {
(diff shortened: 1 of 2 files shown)


mantrakp04:
@greptile-ai review

apps/backend/src/lib/email-queue-step.tsx
```diff
@@ -134,25 +134,68 @@ async function retryEmailsStuckInRendering(): Promise<void> {
...
+ * (either marking the send as finished or unclaiming it for retry). We deliberately do NOT
+ * retry automatically: the upstream send may have actually been accepted by the SMTP/provider
+ * before the worker died, and retrying would risk a duplicate delivery. Instead, we mark the
+ * row as a terminal server error with delivery status unknown, and emit a Sentry signal so a
+ * human can investigate if desired.
+ */
+async function recoverEmailsStuckInSending(): Promise<void> {
+  const stuckCutoff = new Date(Date.now() - STUCK_EMAIL_TIMEOUT_MS);
+  const recoveredAt = new Date();
+  const externalMessage = "Email sending did not complete in time. The email may or may not have been delivered; we are not retrying automatically to avoid sending a duplicate.";
```
nams1570: External message is what the user can see on the email outbox/ sent pages. I don't think the user needs to know about our retry logic.

nams1570: Consider reworking the message to make it more user friendly

apps/backend/src/lib/email-queue-step.tsx
```diff
@@ -134,25 +134,68 @@ async function retryEmailsStuckInRendering(): Promise<void> {
...
+      // status is effectively unknown, and treating it as "no delivery info available" is the
+      // honest representation.
+      canHaveDeliveryInfo: false,
+      sendServerErrorExternalMessage: externalMessage,
+      sendServerErrorExternalDetails: externalDetails,
+      sendServerErrorInternalMessage: `Email was stuck in sending (startedSendingAt <= ${stuckCutoff.toISOString()}) and was recovered by the email queue step. Not retried to avoid duplicate delivery.`,
+      sendServerErrorInternalDetails: internalDetails,
+      shouldUpdateSequenceId: true,
+    },
+    select: { id: true, tenancyId: true, startedSendingAt: true, to: true, sendAttemptErrors: true },
```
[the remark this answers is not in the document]
nams1570: yeah this is a good shout from greptile, this is exposing PII.

apps/backend/src/lib/email-queue-step.tsx
```diff
@@ -134,25 +134,68 @@ async function retryEmailsStuckInRendering(): Promise<void> {
   }
 }
 
-async function logEmailsStuckInSending(): Promise<void> {
-  const res = await globalPrismaClient.emailOutbox.findMany({
+/**
+ * Recover rows that have been stuck in SENDING for longer than STUCK_EMAIL_TIMEOUT_MS.
+ *
+ * These rows indicate a worker died after setting `startedSendingAt` but before cleaning up
```
nams1570: I'm not sure this is entirely true. Yes worker death is a cause, but there could be other causes too. The core issue is that claimEmailsForSending and processTenancyBatch check for startedSendingAt being null before they choose to handle a particular email. Some other causes could be lowLevelEmail... hanging or any of the prisma calls inside processTenancyBatch throwing an exception

nams1570 (changes requested):
Bots are right, you should set nextSendRetryAt to null in the unrecoverable case. This is also done implicitly in queueReadyEmails though maybe we should consider explicitly doing so.

apps/backend/src/lib/email-queue-step.tsx
```diff
@@ -134,25 +134,73 @@ async function retryEmailsStuckInRendering(): Promise<void> {
...
+/**
+ * Recover rows that have been stuck in SENDING for longer than STUCK_EMAIL_TIMEOUT_MS.
+ *
+ * These rows can happen when sending hangs or fails after `startedSendingAt` was set but before
+ * the send attempt was finalized. We deliberately do NOT retry automatically: the upstream send
+ * may have actually been accepted by the SMTP/provider, and retrying would risk a duplicate
+ * delivery. Instead, we mark the row as a terminal server error with delivery status unknown,
+ * and emit a Sentry signal so a human can investigate if desired.
+ */
+async function recoverEmailsStuckInSending(additionalWhere?: Prisma.EmailOutboxWhereInput): Promise<void> {
```
nams1570: nit: to be clear, you're not recovering the emails here you're marking them as failed. I'd suggest renaming this and also changing the text of the internal messages to reflect that.

nams1570 (commented):
just a nit for a rename, after you make that change it should be good to go

nams1570 (approved):

PR #1363, opened 2026-04-21
Fix null-unsafe payments config validation for partial overrides

mantrakp04:
## Summary
- Make the `branchPaymentsSchema` custom validator tolerant of partial override objects
- Avoid crashing when `payments.products` or `payments.productLines` are absent during validation
- Add regression tests for partial configs plus the existing missing-line and customer-type mismatch cases

## Testing
- Added Vitest coverage for partial payments configs and validation failures
- Lint passed for the touched schema files
- Typecheck passed for `packages/stack-shared`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation robustness with stricter type-safety checks for payment-related data configurations.
  * Enhanced error messages for clearer feedback on validation failures.

* **Tests**
  * Added comprehensive test coverage for edge cases including missing configurations and type mismatches.


mantrakp04:
@greptile-ai review

packages/stack-shared/src/config/schema.ts
```diff
@@ -181,15 +181,21 @@ export const branchPaymentsSchema = yupObject({
...
-    for (const [productId, product] of Object.entries(value.products)) {
-      if (!product.productLineId) continue;
-      const productLine = getOrUndefined(value.productLines, product.productLineId);
+    const products = value.products;
+    if (!isObjectLike(products)) return true;
+
+    const productLines = value.productLines;
+    for (const [productId, product] of Object.entries(products)) {
+      if (!isObjectLike(product)) continue;
+      if (product.productLineId == null) continue;
```
nams1570: Discussion: this is a stricter check than what we had before. Is there a reason you switched it? This change would mean empty product line ids are not skipped.

nams1570 (commented):
Test-wise, why not just append to the unit tests in schema.ts? Also, is there a reason behind not using inline snapshots? that's more in line with how the codebase handles it.
Also, it would be nice to have them run abortEarly = false tests if possible since by default that's what smart-request runs.

BilalG1 (approved):

nams1570 (approved):

PR #1371, opened 2026-04-22
fix(internal-tool): continue dev startup when spacetime publish fails

mantrakp04:
## Summary
- pre-dev.mjs now warns and exits 0 when the local SpacetimeDB publish fails, instead of aborting `next dev`
- Lets contributors without a running local SpacetimeDB server still start the internal-tool dev server
- Updates the header comment to reflect the new behavior

## Test plan
- [ ] Run `pnpm dev` in `apps/internal-tool` with no SpacetimeDB server running — dev server should still start, with a warning
- [ ] Run with SpacetimeDB server running and `spacetime` CLI installed — publish still runs and dev proceeds
- [ ] Run without `spacetime` CLI installed — existing warn-and-continue path still works

## Summary by CodeRabbit

* **Chores**
  * Updated local publishing configuration to derive server settings from environment variables for improved flexibility and easier customization.

### DIFF
--- a/apps/internal-tool/scripts/spacetime-publish.mjs
+++ b/apps/internal-tool/scripts/spacetime-publish.mjs
@@ -6,8 +6,27 @@ import { spawnSync } from "node:child_process";
 
 const target = process.argv[2]; // "local" or "prod"
 
+/** HTTP API for 'spacetime publish' (matches docker/dependencies/docker.compose.yaml host port ...39). */
+function localPublishServerUrl() {
+  if (process.env.STACK_SPACETIME_PUBLISH_URL) {
+    return process.env.STACK_SPACETIME_PUBLISH_URL;
+  }
+  const prefix = process.env.NEXT_PUBLIC_STACK_PORT_PREFIX ?? "81";
+  return `http://127.0.0.1:${prefix}39`;
+}
+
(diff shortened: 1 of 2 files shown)


apps/internal-tool/scripts/spacetime-publish.mjs
```diff
@@ -6,8 +6,27 @@ import { spawnSync } from "node:child_process";
...
+  if (process.env.STACK_SPACETIME_PUBLISH_URL) {
+    return process.env.STACK_SPACETIME_PUBLISH_URL;
+  }
+  const prefix = process.env.NEXT_PUBLIC_STACK_PORT_PREFIX ?? "81";
+  return `http://127.0.0.1:${prefix}39`;
+}
+
 const configs = {
-  local: ["publish", "stack-auth-llm", "--server", "local", "-p", "spacetimedb", "--yes", "--no-config", "--delete-data=on-conflict"],
+  local: [
```
nams1570, line 19: for context, if you run `spacetime server list` on your CLI, local is a nickname for your default local spacetime server. It seems to default to 3000, though that being said this is a good fix because we shouldnt rely on it.

nams1570 (approved):

PR #1372, opened 2026-04-22
fix: refresh-token P2025 race with concurrent sign-out

mantrakp04:
## Summary
- Fixes Sentry [STACK-BACKEND-146](https://stackframe-pw.sentry.io/issues/7377768662/): `PrismaClientKnownRequestError` P2025 on `projectUserRefreshToken.update()` during token refresh.
- Root cause: `generateAccessTokenFromRefreshTokenIfValid` (`apps/backend/src/lib/tokens.tsx`) reads the refresh-token row upstream, then issues `.update(...)` on it (and on `projectUser`) inside a `Promise.all`. If a concurrent sign-out (`DELETE /auth/sessions/current`), session revoke, password change, or user deletion removes the row between the read and the update, Prisma throws P2025 and the refresh endpoint 500s.

## Changes
- `apps/backend/src/lib/tokens.tsx` — swap the two `.update(...)`s for `.updateMany(...)` so a missing row is a no-op, then re-check the refresh token still exists; return `null` if it doesn't. The refresh route already maps `null` -> `KnownErrors.RefreshTokenNotFoundOrExpired` (401), which is the correct user-facing behavior for a just-revoked session.
- `apps/backend/src/oauth/model.tsx` — in `generateAccessToken`, replace the "ultra-rare race condition" `throwErr` fallback with `throw new KnownErrors.RefreshTokenNotFoundOrExpired()` so concurrent sign-out during an OAuth `refresh_token` grant returns a clean 401 instead of 500.
- `apps/e2e/tests/backend/endpoints/api/v1/auth/sessions/current/refresh-race.test.ts` — new regression test that fires `POST /auth/sessions/current/refresh` and `DELETE /auth/sessions/current` concurrently with the same refresh token. Before the fix it 500s on the first iteration; after, it passes in ~12s.

## Test plan
- [x] New regression test passes locally.
- [x] Existing `auth/sessions/**` + `auth/oauth/token.test.ts` still pass (27 tests, 3 todo, 0 failed).
- [ ] CI green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refresh flows now detect a revoked or removed refresh token during concurrent operations and stop cleanly, preventing issuance of an access token from stale data.
  * A specific refresh-token-not-found/expired error is returned instead of a generic failure when refresh cannot proceed.

* **Tests**
  * Added E2E tests exercising concurrent refresh vs sign-out to prevent race-condition crashes and validate safe handling of competing requests.

### DIFF
--- a/apps/backend/src/lib/tokens.tsx
+++ b/apps/backend/src/lib/tokens.tsx
@@ -250,9 +250,10 @@ export async function generateAccessTokenFromRefreshTokenIfValid(options: Refres
 
   // updateMany (instead of update) so a concurrent sign-out / session revocation
   // that deletes the row between the caller's read and this write does not
-  // surface as a P2025 500. We re-check existence below and return null if the
-  // token was deleted, which the refresh route maps to RefreshTokenNotFoundOrExpired.
-  await Promise.all([
+  // surface as a P2025 500. count === 0 on the refresh-token update means the
+  // row was deleted mid-flight; return null so the refresh route maps it to
+  // RefreshTokenNotFoundOrExpired.
+  const [, refreshTokenUpdate] = await Promise.all([
     prisma.projectUser.updateMany({
       where: {
         tenancyId: options.tenancy.id,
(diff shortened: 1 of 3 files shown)


mantrakp04:
@greptile-ai review

nams1570 (approved):
To clarify, this doesn't fix the race it just prevents it from crashing. The idea here is that `updateMany` vs `update` returns a 0 instead of crashing with a P2025 when the TOCTOU is hit (viz. the refresh token was deleted after it was read or user was deleted after you start the song and dance) and this lets you handle it with a better user facing error.

Bot suggestions of using a txn to "properly solve it" are prolly overkill and transaction semantics would add more issues than they solve, so I think this is good to go.

PR #1374, opened 2026-04-23
fix(init-prompt): require StackProvider for all frameworks

BilalG1:
## Summary
- The init prompt marked the `StackProvider` step as *React only* and placed it after the `StackHandler` step. Following it on a Next.js project produced a layout with no provider, so `StackHandler` crashed at runtime with `useStackApp must be used within a StackProvider`.
- Make the provider step unconditional and move it ahead of the handler step so the dependency order matches the instruction order. Also quote the exact error message so the model won't skip it.

## Test plan
- [ ] Run `npx @stackframe/stack-cli init` (or the web flow) against a fresh Next.js app and confirm `/handler/[...stack]` renders without the `useStackApp` error.
- [ ] Re-run against a Vite/React app to confirm the reordered instructions still produce a working setup.

## Summary by CodeRabbit

* **Documentation**
  * Added explicit global MCP config file path guidance for several coding agents.
  * Documented required provider configuration across supported frameworks.
  * Clarified where to place provider wrappers in root app layouts (including Next.js app/layout).
  * Reordered setup steps to surface the required "Wrap your app in a Stack provider" step and updated step numbering.

### DIFF
--- a/packages/stack-shared/src/helpers/init-prompt.ts
+++ b/packages/stack-shared/src/helpers/init-prompt.ts
@@ -65,47 +65,9 @@ export const stackServerApp = new StackServerApp({
 });
 \`\`\`
 
-### 3) Create the Stack handler (if available in framework)
+### 3) Wrap your app in a Stack provider
 
-This sets up pages for sign in, sign up, password reset, etc.
-
-\`\`\`tsx
-import { StackHandler } from "@stackframe/stack"; // Next.js
-// import { StackHandler } from "@stackframe/react"; // React
-
-export default function Handler() {
-  return <StackHandler fullPage />;
-}
-\`\`\`
-
-### 4) Create a Suspense boundary
-
-Suspense is necessary for many stack auth hooks such as useUser to function. Add a loading component with a custom loading indicator for the current project. Don't add if one already exists
-
-For example:
-\`\`\`tsx
-//src/loading.tsx
-
-export default function Loading() {
-  return <p>Loading...</p>
-}
-\`\`\`
-
-### 5) Link environment variables
-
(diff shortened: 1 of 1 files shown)


packages/stack-shared/src/helpers/init-prompt.ts
```diff
@@ -65,7 +69,28 @@ export const stackServerApp = new StackServerApp({
...
+return (
+  <body>
+    <StackProvider app={stackClientApp}>
+      <StackTheme>{children}</StackTheme>
+    </StackProvider>
+  </body>
+);
+\`\`\`
+
+### 4) Create the Stack handler (if available in framework)
```
nams1570, line 93: nit: Does this still apply? Looks like you're pulling it for all frameworks

nams1570 (approved):

packages/stack-shared/src/helpers/init-prompt.ts
```diff
@@ -65,7 +69,28 @@ export const stackServerApp = new StackServerApp({
 });
 \`\`\`
 
-### 3) Create the Stack handler (if available in framework)
+### 3) Wrap your app in a Stack provider
+
+Required for all frameworks (including Next.js). \`StackHandler\`, \`useUser\`, and \`useStackApp\` all depend on it — without it you will get "useStackApp must be used within a StackProvider" at runtime. In Next.js, add it to the root \`app/layout.tsx\` around \`{children}\`. In React/Vite, wrap your root component.
+
+\`\`\`tsx
+import { StackProvider, StackTheme } from "@stackframe/stack";
```
nams1570: Are these imports right for react/vite?

PR #1379, opened 2026-04-23
cli add back init options

BilalG1:
## Summary by CodeRabbit

* **New Features**
  * Added a "create-cloud" mode to the CLI init flow.
  * New interactive project creation flow that can prompt for display name and select/create a team-backed project.

* **Behavior Changes**
  * Init now resolves mode from flags, config, or interactive prompts; prompts to choose linking vs creating when inputs are missing.
  * Non-interactive runs now error when required inputs are absent; cloud linking offers auto-create in interactive mode.

* **Refactor**
  * Centralized auth, project-creation, and env key writing for clearer, safer linking and creation flows.

### DIFF
--- a/packages/stack-cli/src/commands/init.ts
+++ b/packages/stack-cli/src/commands/init.ts
@@ -10,11 +10,12 @@ import { writeConfigValue } from "../lib/config.js";
 import { CliError, AuthError } from "../lib/errors.js";
 import { isNonInteractiveEnv } from "../lib/interactive.js";
 import { createInitPrompt } from "../lib/init-prompt.js";
+import { createProjectInteractively } from "../lib/create-project.js";
 import { runClaudeAgent } from "../lib/claude-agent.js";
 import { detectImportPackageFromDir, renderConfigFileContent } from "@stackframe/stack-shared/dist/config-rendering";
 
 type InitOptions = {
-  mode?: "create" | "link-config" | "link-cloud",
+  mode?: "create" | "create-cloud" | "link-config" | "link-cloud",
   apps?: string,
   configFile?: string,
   selectProjectId?: string,
@@ -26,7 +27,7 @@ export function registerInitCommand(program: Command) {
   program
     .command("init")
     .description("Initialize Stack Auth in your project")
-    .option("--mode <mode>", "Mode: create, link-config, or link-cloud (skips interactive prompts)")
+    .option("--mode <mode>", "Mode: create, create-cloud, link-config, or link-cloud (skips interactive prompts)")
     .option("--apps <apps>", "Comma-separated app IDs to enable (for create mode)")
     .option("--config-file <path>", "Path to existing config file (for link-config mode)")
     .option("--select-project-id <id>", "Project ID to link (for link-cloud mode)")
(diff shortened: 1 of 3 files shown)


packages/stack-cli/src/commands/init.ts
```diff
@@ -226,7 +219,70 @@ async function handleLinkFromCloud(flags: Record<string, unknown>, opts: InitOpt
...
+        name: `${p.displayName} (${p.id})`,
+        value: p.id,
+      })),
+    });
+  }
+
+  const project = projects.find((p) => p.id === projectId)!;
+  await writeProjectKeysToEnv(project, outputDir);
   return {};
 }
```
nams1570, line 338: <p class='graphite__hidden'><i>[Re: line +357]</i></p>
Doesn't this silently overwrite the config file? Is this expected or should we prompt the user like we do for .env updates?  
I feel like if I was a user I would want to know if you were going to do this
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/stack-auth/stack-auth/1379?utm_source=unchanged-line-comment">Graphite</a>.</p>

packages/stack-cli/src/commands/init.ts
```diff
@@ -26,7 +28,7 @@ export function registerInitCommand(program: Command) {
   program
     .command("init")
     .description("Initialize Stack Auth in your project")
-    .option("--mode <mode>", "Mode: create, link-config, or link-cloud (skips interactive prompts)")
+    .option("--mode <mode>", "Mode: create, create-cloud, link-config, or link-cloud (skips interactive prompts)")
```
nams1570, line 31: request: This isn't a good fit for this block but i couldnt find a better spot. I think you need to have option validation. It's theoretically possible to pass create-cloud with --select-project-id which makes no sense, or to pass --apps with create-cloud which does nothing

packages/stack-cli/src/commands/init.ts
```diff
@@ -57,24 +59,36 @@ async function runInit(program: Command, opts: InitOptions) {
...
+        { name: "Stack Auth Cloud", value: "hosted" as const },
+        { name: "Local (requires local emulator installation, ~1.3gb storage required)", value: "local" as const },
+      ],
+    });
+    mode = location === "local" ? "create" : "create-cloud";
+  }
 
   let configPath: string | undefined;
 
   if (mode === "link" || mode === "link-config" || mode === "link-cloud") {
```
nams1570: isnt this mode === "link" check dead code? we dont support it anymore right?

nams1570: Also by this point wouldn't we have dealt with the other two possible link modes?

nams1570: So this would be dead code?

packages/stack-cli/src/commands/init.ts
```diff
@@ -57,24 +59,36 @@ async function runInit(program: Command, opts: InitOptions) {
...
+    mode = opts.mode;
+  } else if (opts.selectProjectId) {
+    mode = "link-cloud";
+  } else if (opts.configFile) {
+    mode = "link-config";
+  } else {
+    console.log("Creating a new Stack Auth project.\n");
+    const location = await select({
+      message: "Where would you like to create the project?",
+      choices: [
```
nams1570: So vercel bot is right here, why are we no longer supporting linking to existing project from here?

packages/stack-cli/src/commands/init.ts
```diff
@@ -226,7 +219,70 @@ async function handleLinkFromCloud(flags: Record<string, unknown>, opts: InitOpt
     fs.writeFileSync(envPath, envLines + "\n");
```
nams1570, line 269: bug: if the path isnt valid here, we would have ended up potentially creating a new project for nothing

nams1570: writeFileSync would throw an ENOENT

nams1570 (changes requested):

packages/stack-cli/src/commands/init.ts
nams1570: this was supposed to be for lines 356-357

PR #1381, opened 2026-04-24
[codex] Fix Neon malformed Basic auth validation

mantrakp04:
## What changed

This fixes Sentry issue [STACK-BACKEND-1A3](https://stackframe-pw.sentry.io/issues/7436639623/?project=4507442898272256&query=is%3Aunresolved&referrer=issue-stream&seerDrawer=true).

A request with this malformed header:

```http
Authorization: Basic
```

used to crash the Neon auth validator with a `StackAssertionError`, which turned a bad client request into a 500.

The fix makes `neonAuthorizationHeaderSchema` only validate Neon client credentials after the Basic auth header successfully decodes. If decoding fails, the Neon-specific validator returns `true` and lets `basicAuthorizationHeaderSchema` produce the intended 400 schema error: `Authorization header must be in the format "Basic <base64>"`.

## Reviewer walkthrough

There are two checks chained together:

1. `basicAuthorizationHeaderSchema` checks that the header is structurally valid Basic auth.
2. `neonAuthorizationHeaderSchema` checks that the decoded `client_id:client_secret` matches a configured Neon client.

Yup may still run the second check after the first one has failed, because route validation collects errors with `abortEarly: false`. The old code assumed the first check had already passed and called `throwErr(...)` when decoding returned `null`. This PR changes that path to return `true`, because the format error is already owned by the first check.

## Tests

- `pnpm -C packages/stack-shared exec vitest run --maxWorkers=1 --minWorkers=1 src/schema-fields.ts`
- `pnpm -C apps/e2e exec vitest run --maxWorkers=1 --minWorkers=1 tests/backend/endpoints/api/v1/integrations/neon/projects/transfer.test.ts -t "malformed"`
- `pnpm -C packages/stack-shared lint`
- `pnpm -C packages/stack-shared typecheck`
- `pnpm -C apps/e2e lint`
- `pnpm -C apps/e2e typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authorization header validation so malformed Basic credentials yield the standard, user-facing validation error instead of a provider-specific response.

* **Tests**
  * Added end-to-end and unit tests covering authorization header validation and edge cases to ensure consistent 400/SCHEMA_ERROR responses and expected error details.

### DIFF
--- a/apps/e2e/tests/backend/endpoints/api/v1/integrations/neon/projects/transfer.test.ts
+++ b/apps/e2e/tests/backend/endpoints/api/v1/integrations/neon/projects/transfer.test.ts
@@ -216,6 +216,7 @@ it("should fail if the neon client details are missing", async ({ expect }) => {
 });
 
 it("should fail if the neon client authorization header is malformed", async ({ expect }) => {
+  // This project ID is arbitrary because malformed Basic auth is rejected before any project lookup runs.
   const projectId = "73782539-cf39-486b-a9f8-9b2893f79ef2";
   const response = await niceBackendFetch(urlString`/api/v1/integrations/neon/projects/transfer?project_id=${projectId}`, {
     method: "GET",
--- a/apps/e2e/tests/backend/endpoints/api/v1/integrations/neon/projects/transfer.test.ts
+++ b/apps/e2e/tests/backend/endpoints/api/v1/integrations/neon/projects/transfer.test.ts
@@ -215,6 +215,38 @@ it("should fail if the neon client details are missing", async ({ expect }) => {
   `);
 });
 
+it("should fail if the neon client authorization header is malformed", async ({ expect }) => {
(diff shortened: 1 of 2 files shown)


nams1570 (approved):
This is an interesting bug. It stems from the fact that the `createSmartRouteHandler` has `abortEarly` set to false, meaning all yup validations run instead of the latter ones not running if the former failed. So the assumption in the throwErr was invalid...
I think it makes sense for `smart-request` to keep `abortEarly` false generally, we get more information on a failure that way. 
The fix is correct as far as I can tell, but your unit test doesn't actually cover the scenario you just fixed. By default the validate runs with abortEarly = True, but our scenario necessitates abortEarly=False. I believe the e2e test does which is good but the unit test doesn't. I'll leave it up to you to decide whether to get rid of the unit test or fix it.
We do run `decodeBasicAuthorizationHeader` three times for this one route though. Wondering if there's room for a refactor there. but i wouldn't block merging on it.

PR #1390, opened 2026-04-27
Add stack doctor

aadesh18:
## Summary

Adds `stack doctor` — a read-only companion to `stack fix`. No AI, no network, no edits. Detects the framework, runs a fixed list of checks, prints a pass/fail report, and exits 0 if everything's green or 1 if any check fails.

Stacked on top of #1387 (the fix-command PR) so the diff here is just the doctor work.

## Usage

```
stack doctor [--output-dir <dir>] [--framework <next|react|js>] [--json]
```

- `--output-dir`: project root to inspect (defaults to cwd).
- `--framework`: override auto-detection.
- `--json`: machine-readable report instead of the human format.

Exit code is 0 on all-pass-or-warn, 1 on any failure or unsupported project.

## Pre-checks

- `package.json` exists in the target directory. Missing → exit 1.
- `package.json` is valid JSON. Malformed → exit 1 with the parse error.
- `package.json` declares at least one dependency.
- A framework is recognized:
    - `next` in deps **AND** `app/` or `src/app/` exists → Next.js.
    - `next` with only `pages/` → exit 1, "pages router not yet supported".
    - `react` or `react-dom` in deps → React.
    - Anything else with deps → JS catch-all (covers SvelteKit, Astro, Remix, Vue, Node services).

### Next.js (7 checks)

1. `@stackframe/stack` is installed (deps or devDeps).
2. Stack client app instance file exists (`stack/client.ts(x)`, prefixed with `src/` if the project uses `src/app/`).
3. Stack server app instance file exists (`stack/server.ts(x)`, same prefix rule).
4. Handler route exists (`app/handler/[...stack]/page.{tsx,ts,jsx,js}`, with the same prefix).
5. Root layout (`app/layout.{tsx,jsx,ts,js}`) imports `StackProvider` from `@stackframe/stack` AND renders `<StackProvider>` in JSX. Distinguishes:
    - Pass: import + JSX usage.
    - Warn: import present, JSX missing.
    - Fail: JSX present, import missing.
    - Fail: layout file missing entirely.
6. Required env vars: `NEXT_PUBLIC_STACK_PROJECT_ID` (fail), `NEXT_PUBLIC_STACK_PUBLISHABLE_CLIENT_KEY` (warn — only required when the dashboard's "require publishable client keys" toggle is on), `STACK_SECRET_SERVER_KEY` (fail).
7. `stack.config.ts` validity — only runs if the file exists. Loads it via `jiti`, requires a `config` named export that is a plain object (mirroring `stack config push` validation). Skipped silently if no config file is present.

### React (4 checks)

1. `@stackframe/react` is installed.
2. Stack client app instance file exists (`stack/client.ts(x)`, prefixed with `src/` if the project has `src/`).
3. Required env vars: `VITE_STACK_PROJECT_ID` (fail), `VITE_STACK_PUBLISHABLE_CLIENT_KEY` (warn). No server secret check — React-only path is client-only.
4. `stack.config.ts` validity (same as Next).

### JS / Node catch-all (4 checks)

Covers SvelteKit, Astro, Remix, Vue, Nest, plain Node services.

1. `@stackframe/js` is installed.
2. Stack app instance file exists — either `stack/client.ts` or `stack/server.ts` (with `src/` prefix if applicable). Either-or, since backend services often only ship `server.ts` and frontend-only SvelteKit only `client.ts`.
3. Required env vars: `STACK_PROJECT_ID` or `PUBLIC_STACK_PROJECT_ID` (fail), `STACK_PUBLISHABLE_CLIENT_KEY` or `PUBLIC_STACK_PUBLISHABLE_CLIENT_KEY` (warn), `STACK_SECRET_SERVER_KEY` (fail). The `PUBLIC_*` aliases exist because SvelteKit and Astro require that prefix to expose vars to client code.
4. `stack.config.ts` validity (same as Next).

## Env var lookup (shared by all frameworks)

For each required name (or alias group), check in this order:

1. `.env.local` at project root.
2. `.env`.
3. `process.env`.

A var is "present" if the resolved value is non-empty after trimming. Values are **never printed** — only presence/absence — so the secret server key cannot leak through doctor output.

For alias groups (`STACK_PROJECT_ID` / `PUBLIC_STACK_PROJECT_ID`), if any alias is present the logical var is satisfied.

## Severity rules

- `fail` counts toward exit 1.
- `warn` does not count toward exit 1.
- `skip` (returning null from a check) produces no row at all — used by the optional config-file check.

##

### DIFF
--- a/apps/e2e/tests/general/cli.test.ts
+++ b/apps/e2e/tests/general/cli.test.ts
@@ -565,3 +565,416 @@ describe("Stack CLI — Emulator", () => {
     expect(stdout).toContain("--repo");
   });
 });
+
+// Doctor CLI tests — no backend required. Each test builds a fixture project
+// in a temp dir and runs `stack doctor --output-dir <dir> --json`.
+describe("Stack CLI — Doctor", () => {
+  let doctorTmpRoot: string;
+
+  beforeAll(() => {
+    if (!fs.existsSync(CLI_BIN)) {
+      throw new Error("CLI not built. Run `pnpm --filter @stackframe/stack-cli run build` first.");
+    }
+    doctorTmpRoot = fs.mkdtempSync(path.join(os.tmpdir(), "stack-cli-doctor-test-"));
+  });
+
+  afterAll(() => {
+    if (doctorTmpRoot && fs.existsSync(doctorTmpRoot)) {
+      fs.rmSync(doctorTmpRoot, { recursive: true });
+    }
+  });
+
+  function runDoctor(
+    args: string[],
+    envOverrides?: Record<string, string>,
+  ): Promise<{ stdout: string, stderr: string, exitCode: number | null }> {
+    const env: Record<string, string> = {
+      PATH: process.env.PATH ?? "",
+      HOME: process.env.HOME ?? "",
+      CI: "1",
+      ...envOverrides,
+    };
+    return new Promise((resolve) => {
+      execFile("node", [CLI_BIN, ...args], {
+        env,
+        timeout: 30_000,
+      }, (error, stdout, stderr) => {
+        resolve({
+          stdout: stdout.toString(),
+          stderr: stderr.toString(),
+          exitCode: error ? (error as any).code ?? 1 : 0,
+        });
+      });
+    });
+  }
+
+  function makeProject(subdir: string, files: Record<string, string>): string {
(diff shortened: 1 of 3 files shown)


aadesh18:
@greptile review again

aadesh18:
@greptile review - i fixed those issues

BilalG1 (approved):
we can probably add a few tests for this in cli.test.ts but otherwise looks good

PR #1395, opened 2026-04-29
feat: mock SAML 2.0 IdP for testing

BilalG1:
First of 3 stacked PRs adding SAML 2.0 SSO. **Stacked above:** [#PR_BACKEND] (backend feature) → [#PR_CLIENT] (client + tests).

## What this PR adds

- **`apps/mock-saml-idp/`** — standalone Bun/Express service that acts as a multi-tenant SAML 2.0 Identity Provider for e2e tests and local dev. Listens on port 8115. Each tenant has its own RSA keypair + self-signed cert generated at startup, so one mock service can back many SamlConnection rows.
- **CI workflow** — starts \`mock-saml-idp\` alongside \`mock-oauth-server\` in \`.github/workflows/e2e-api-tests.yaml\`.
- **Root scripts** — \`pnpm start:mock-saml-idp\` + \`dev:basic\` filter inclusion.
- **Snapshot serializer** — strips SAML query params (\`SAMLRequest\`, \`SAMLResponse\`, \`RelayState\`), the \`stack-saml-inner-\` CSRF cookie prefix, and SAML xs:ID identifiers + IssueInstant/NotBefore/NotOnOrAfter timestamps so e2e snapshots don't reroll.
- **Seed-dummy-data** — \`STACK_SEED_ENABLE_SAML=true\` block that pre-creates acme + globex SAML connections on the dummy project, fetching the IdP cert from the running mock at seed time. Dormant unless the env var is set; safe to ship before the backend SAML routes exist.

## Why a separate service (and not just a stub in the test file)

The mock uses **\`samlify\`** deliberately because the upcoming backend SAML wrapper uses **\`@node-saml/node-saml\`**. Different libraries on each side mean a bug in either library's signature canonicalization surfaces as a test failure rather than canceling out (the backend would otherwise be its own test oracle, which is no oracle at all).

## Endpoints

- \`GET  /idp/:tenant/metadata\` — IdP metadata XML
- \`GET  /idp/:tenant/sso\` — AuthnRequest receiver, renders a login form
- \`POST /idp/:tenant/login\` — builds + auto-POSTs a signed SAMLResponse
- \`POST /idp/:tenant/test-controls\` — queues misbehaviors for the next assertion: \`bad-signature\`, \`expired\`, \`not-yet-valid\`, \`wrong-audience\`, \`wrong-in-response-to\`, \`missing-name-id\`, \`missing-email\`, \`replay\`, \`sign-with-tenant\`. The backend's negative e2e tests trigger these to ensure the validator actually rejects bad assertions.
- \`GET  /idp\` — introspection (lists tenants + queued misbehaviors)

## Test plan

- [x] \`pnpm --filter @stackframe/mock-saml-idp typecheck\` passes
- [x] \`pnpm --filter @stackframe/mock-saml-idp lint\` passes
- [x] \`pnpm --filter @stackframe/mock-saml-idp start\` boots and serves valid metadata for both tenants with distinct certs
- [x] CI \`background-action\` step starts the mock and \`http://localhost:8115/idp\` becomes ready

The mock IdP is independently useful — it can back SAML testing for any future feature, not just the SSO work in the next PR.

## Release Notes

* **New Features**
  * Added a mock SAML 2.0 Identity Provider service supporting multiple tenants, per-tenant key generation, and configurable assertion scenarios.

* **Tests**
  * Enhanced E2E test infrastructure with improved snapshot serialization, including better SAML authentication data redaction and parameter handling.

* **Chores**
  * Integrated mock SAML IdP into the development launchpad for convenient access.


mantrakp04:
@greptile-ai review

apps/mock-saml-idp/src/index.ts
```diff
@@ -0,0 +1,445 @@
...
+ * uses `@node-saml/node-saml`. Different libraries on each side means a bug
+ * in either library's signature canonicalization surfaces as a test failure
+ * instead of being masked by both sides agreeing.
+ */
+import express from 'express';
+import handlebars from 'handlebars';
+import forge from 'node-forge';
+import * as samlify from 'samlify';
+
+const stackPortPrefix = process.env.NEXT_PUBLIC_STACK_PORT_PREFIX ?? "81";
```
mantrakp04, line 20: bug: this uses suffix `15`, but that suffix is already taken by `examples/supabase` (`next dev --port ${NEXT_PUBLIC_STACK_PORT_PREFIX:-81}15`) and the dev launchpad still labels suffix 15 as Supabase.

Concrete failure mode: with the default prefix, both services want `8115`; with `NEXT_PUBLIC_STACK_PORT_PREFIX=91`, both want `9115`. In `pnpm dev:full` or any workflow that starts both, whichever service starts second fails to bind the port. Can we pick an unused suffix for the mock SAML service and update the launchpad/app config consistently?

apps/backend/src/lib/seed-dummy-data.ts
```diff
@@ -1939,6 +1939,69 @@ async function seedBulkSignupsAndActivity(options: {
...
 
+/**
+ * Pre-creates two SAML connections (acme + globex) on the dummy project that
+ * point at the local mock SAML IdP. Gated on STACK_SEED_ENABLE_SAML='true'.
+ * Fetches the mock IdP's metadata at seed time so the seeded cert matches
+ * the cert the mock generated at startup — the mock currently regenerates
+ * keys per restart, so re-seed if you restart the mock.
+ */
+async function seedSamlConnections(projectId: string): Promise<void> {
+  const mockUrl = getEnvVariable("STACK_MOCK_SAML_URL", "http://localhost:8115");
```
mantrakp04: bug: this default URL ignores `NEXT_PUBLIC_STACK_PORT_PREFIX`, but the mock IdP itself derives its port from that prefix.

So if I run with `NEXT_PUBLIC_STACK_PORT_PREFIX=91`, the mock starts on `http://localhost:9115`, but seed still fetches `http://localhost:8115/idp/acme/metadata`. That either fails, or worse, hits a stale service from a different dev environment. The default should be derived from the same prefix/suffix as the mock, with `STACK_MOCK_SAML_URL` only as the explicit override.

apps/backend/src/lib/seed-dummy-data.ts
```diff
@@ -1939,6 +1939,69 @@ async function seedBulkSignupsAndActivity(options: {
...
+    }),
+  );
+
+  // Set the entire connection entry as a single value, not as deep
+  // dot-keys — config normalization with onDotIntoNonObject="ignore"
+  // drops dot-keys that try to navigate into a record entry that
+  // doesn't yet exist (same convention as auth.oauth.providers).
+  const overlay: Parameters<typeof overrideEnvironmentConfigOverride>[0]["environmentConfigOverrideOverride"] = {};
+  for (const f of fetched) {
+    overlay[`auth.saml.connections.${f.slug}`] = {
```
mantrakp04: BUG: as far as I can tell, this writes `auth.saml.connections.*`, but this branch's config schema does not define `auth.saml` yet. I only see the existing `auth.password` / `otp` / `passkey` / `oauth` sections in `packages/stack-shared/src/config/schema.ts`.

That means `STACK_SEED_ENABLE_SAML=true` can successfully fetch metadata and write overrides, but the resulting config is invalid / dropped during normalization, so the dummy project still doesn't get usable SAML connections. If this PR depends on the backend SAML/schema PR, can we either stack it on that branch or keep the seed path out until the schema exists on the base?

apps/mock-saml-idp/src/index.ts
```diff
@@ -0,0 +1,445 @@
...
+async function buildAssertion(
+  tenant: TenantState,
+  parsed: ParsedRequest,
+  user: { email: string, displayName: string },
+): Promise<{ samlResponseB64: string, acsUrl: string }> {
+  const misbehavior = tenant.nextMisbehavior;
+  tenant.nextMisbehavior = { kind: 'none' };
+
+  // `replay`: re-emit the cached response.
+  if (misbehavior.kind === 'replay') {
```
mantrakp04: Potential bug: the replay cache stores `relayState`, but the replay path only returns the old `SAMLResponse` + `acsUrl`; later the login route renders the form with the current request's `relayState`.

So this doesn't replay the previous POST exactly. It tests "old SAMLResponse with new RelayState", which may be a different failure mode than a real replay attack depending on how the SP binds RelayState/request state. If the intent is a true replay, return/use the cached RelayState too. If the intent is specifically mismatched old response + fresh state, I think the misbehavior should be named/scoped that way so tests don't assert the wrong thing.

apps/backend/package.json
```diff
@@ -61,6 +61,7 @@
     "@aws-sdk/client-s3": "^3.855.0",
     "@clickhouse/client": "^1.14.0",
     "@node-oauth/oauth2-server": "^5.1.0",
+    "@node-saml/node-saml": "^5.0.0",
```
mantrakp04: why is this dependency added here? I don't see any backend import of `@node-saml/node-saml` in this diff, and the mock IdP comment says the backend wrapper uses it, which doesn't seem true on this base.

If the backend SAML wrapper is coming in a stacked PR, I think this dependency/comment should live there instead. Otherwise this PR increases backend dependency surface without the owning code being present, and the comment in the mock is misleading for anyone reviewing this branch by itself.

apps/mock-saml-idp/src/index.ts
```diff
@@ -0,0 +1,445 @@
...
+
+// ---------- assertion building -------------------------------------------
+
+const ASSERTION_LIFETIME_MS = 5 * 60 * 1000;
+
+function isoNow(offsetMs = 0): string {
+  return new Date(Date.now() + offsetMs).toISOString();
+}
+
+async function buildAssertion(
```
mantrakp04: Suggestion: `buildAssertion` is carrying a lot of different responsibilities right now: consuming mutable tenant state, implementing replay behavior, choosing the signing tenant, constructing the SP, computing assertion fields, doing XML replacement/mutation, signing, and updating replay cache.

This makes the core test primitive hard to reason about, especially because several misbehaviors are stateful or security-sensitive. Can we split this into smaller helpers along the lines of `consumeNextMisbehavior`, `resolveSigningTenant`, `buildAssertionFields`, `renderLoginResponseXml`, and `cacheReplayableResponse`? That would also make it much easier to add focused tests for the edge cases.

apps/mock-saml-idp/src/index.ts
mantrakp04: also pls update the "pnpm kms" script to include this port

mantrakp04 (changes requested):

mantrakp04 (approved):

BilalG1:

PR #1396, opened 2026-04-29
feat(backend): SAML SSO backend (schema, helpers, routes, admin CRUD)

BilalG1:
Second of 3 stacked PRs adding SAML 2.0 SSO. **Stacked on:** #1395 (mock IdP). **Stacked above:** #1397 (client + tests).

> Note: target branch is \`pr/saml-mock-idp\`. Will be retargeted to \`dev\` once #1395 merges.

## What this PR adds

The backend half of SAML SSO. Purely server-side — client SDK methods, dashboard page, demo page, and e2e tests live in the next PR (which depends on this one).

### Database

- **3 new Prisma models** in \`apps/backend/prisma/schema.prisma\` + migration \`20260429000000_add_saml_sso_models\`:
  - **\`ProjectUserSamlAccount\`** — per-user SAML account, mirrors \`ProjectUserOAuthAccount\`. Unique \`(tenancyId, samlConnectionId, nameId)\` enforces multi-tenant connection isolation at the DB level — the same NameID arriving via a different connection is treated as a separate identity.
  - **\`SamlAuthMethod\`** — connects an \`AuthMethod\` to a \`ProjectUserSamlAccount\`.
  - **\`SamlOuterInfo\`** — in-flight AuthnRequest state, keyed by AuthnRequest ID (TEXT, since SAML xs:IDs aren't UUIDs).

### Connection config storage

Per-connection config (entity ID, IdP cert, ACS URL, attribute mapping, domain) is stored as **JSON in \`tenancy.config.auth.saml.connections\`** — same pattern as \`auth.oauth.providers\`. No dedicated SamlConnection Prisma model. Connection-isolation invariant lives at the per-user table level.

### Schema

- \`packages/stack-shared/src/config/schema.ts\` — adds \`auth.saml.{accountMergeStrategy, connections}\` at branch level (non-sensitive fields) + extends connections at environment level with IdP-side fields (entity ID, SSO URL, X.509 cert, attribute mapping). Defaults block adds a SAML record with a per-key default function. Fuzzer test config updated.

### Helpers

- **\`apps/backend/src/lib/external-auth.tsx\`** — extracts the email-merge strategy switch out of \`oauth.tsx\` into a shared helper so both protocols use the same contact-channel lookup and link_method/raise_error/allow_duplicates logic.
- **\`apps/backend/src/lib/saml-account.tsx\`** — SAML-side parallel of OAuth's findExisting/link/create user-linking helpers, scoped by \`(tenancyId, samlConnectionId, nameId)\`.

### Protocol wrapper

- **\`apps/backend/src/saml/saml.tsx\`** — wraps \`@node-saml/node-saml\`. \`buildSamlClient\`, \`buildAuthnRequestUrl\` (returns URL + extracted requestId for replay protection), \`parseAndVerifyAssertion\` (signature + audience + clock skew + InResponseTo), \`getSpMetadataXml\`. Defines \`SamlConnectionConfig\` locally.
- **\`apps/backend/src/saml/metadata-parser.tsx\`** — pulls entity ID, SSO URL, and signing X.509 out of pasted IdP metadata XML using xmldom + xpath.
- **\`apps/backend/src/saml/discovery.tsx\`** — email-domain → connection lookup for the upcoming \`signInWithSso({ email })\` SDK flow.

### Routes

Under \`apps/backend/src/app/api/latest/auth/saml/\`:

- **\`GET /auth/saml/login/[connection_id]\`** — receives the same OAuth client params as \`/auth/oauth/authorize\`, builds an AuthnRequest, persists OAuth context + AuthnRequest ID in \`SamlOuterInfo\`, sets a CSRF cookie, redirects to the IdP. Honors \`stack_response_mode=json\`.
- **\`POST /auth/saml/acs/[connection_id]\`** — receives the IdP's POST. Parses InResponseTo, looks up \`SamlOuterInfo\` to recover tenancy, validates CSRF cookie, runs node-saml's full \`validatePostResponseAsync\` (signature + audience + clock skew + InResponseTo). Defense-in-depth re-checks InResponseTo and **cross-connection mismatch** (an assertion sent to the wrong ACS endpoint is rejected — see e2e test #10 in the next PR).
- **\`GET /auth/saml/metadata/[connection_id]\`** — public-fetchable SP metadata XML for the customer's IdP admin to paste.
- **\`GET /auth/saml/discover\`** — email-domain → connection lookup.

ACS hands off to \`oauthServer.authorize()\` so Stack Auth issues a customer-facing OAuth code (Stack Auth itself acts as an OAuth2 provider to the customer's SDK — same pattern as the existing OAuth callback handler at \`oauth/callback/[provider_id]\`).

### Admin CRUD

- \`POST/GET/DELETE /api/v1/saml-connections\` + \`GET /api/v1/saml-connections/[id]\` — admin-only thin REST wrappers around the JSON-config storage.

## Out-of-scope (planned follow-ups)

- IdP-initiated SSO
- Signed AuthnRequests, encrypted assertions
- Single Logout (SLO)
- SCIM 2.0 user provisioning
- Group/role attribute mapping → team membership sync
- Automatic IdP metadata refresh from a URL

## Test plan

- [x] \`pnpm --filter @stackframe/backend typecheck\` passes
- [x] \`pnpm --filter @stackframe/backend lint\` passes
- [x] \`pnpm --filter @stackframe/stack-shared typecheck\` passes
- [x] Migration applies cleanly (\`prisma migrate dev\`)
- [x] Codegen succeeds (\`pnpm --filter @stackframe/backend codegen-prisma\`)
- [ ] Routes exercised end-to-end via the e2e suite in the next PR (round-trip test against mock IdP, plus failure matrix for bad-signature / expired / wrong-audience / replay / cross-connection forgery)

### DIFF
--- a/apps/backend/prisma/schema.prisma
+++ b/apps/backend/prisma/schema.prisma
@@ -645,9 +645,9 @@ model OAuthOuterInfo {
 // ProjectUserOAuthAccount. samlConnectionId is the string key under
 // tenancy.config.auth.saml.connections.
 model ProjectUserSamlAccount {
-  id               String  @default(uuid()) @db.Uuid
-  tenancyId        String  @db.Uuid
-  projectUserId    String? @db.Uuid
+  id               String @default(uuid()) @db.Uuid
+  tenancyId        String @db.Uuid
+  projectUserId    String @db.Uuid
   samlConnectionId String
   nameId           String
 
@@ -659,10 +659,11 @@ model ProjectUserSamlAccount {
   email        String?
   nameIdFormat String?
 
-  // Before the SAML account is connected to a user (e.g. mid-flight), the
-  // projectUser is null. Once the ACS callback creates/links the user, this
-  // is set.
-  projectUser ProjectUser? @relation(fields: [tenancyId, projectUserId], references: [tenancyId, projectUserId], onDelete: Cascade)
+  // SAML accounts always reference a real user — the ACS handler creates
+  // (or links) the user before persisting the account row, so there is no
+  // intermediate "unlinked" state. This invariant lets findExistingSamlAccount
+  // assume projectUserId is non-null without a runtime fallback.
+  projectUser ProjectUser @relation(fields: [tenancyId, projectUserId], references: [tenancyId, projectUserId], onDelete: Cascade)
 
   // if allowSignIn is true, samlAuthMethod must be set
   samlAuthMethod SamlAuthMethod?
--- a/apps/backend/prisma/schema.prisma
+++ b/apps/backend/prisma/schema.prisma
@@ -308,6 +308,7 @@ model ProjectUser {
   signUpRiskScoreFreeTrialAbuse Int @default(0) @db.SmallInt
 
   projectUserOAuthAccounts ProjectUserOAuthAccount[]
+  projectUserSamlAccounts  ProjectUserSamlAccount[]
   teamMembers              TeamMember[]
   contactChannels          ContactChannel[]
   authMethods              AuthMethod[]
@@ -317,6 +318,7 @@ model ProjectUser {
   passkeyAuthMethod          PasskeyAuthMethod[]
   otpAuthMethod              OtpAuthMethod[]
   oauthAuthMethod            OAuthAuthMethod[]
+  samlAuthMethod             SamlAuthMethod[]
   projectApiKey              ProjectApiKey[]
   directPermissions          ProjectUserDirectPermission[]
   Project                    Project?                      @relation(fields: [projectId], references: [id])
@@ -486,6 +488,7 @@ model AuthMethod {
   passwordAuthMethod PasswordAuthMethod?
   passkeyAuthMethod  PasskeyAuthMethod?
   oauthAuthMethod    OAuthAuthMethod?
+  samlAuthMethod     SamlAuthMethod?
 
   projectUser ProjectUser @relation(fields: [tenancyId, projectUserId], references: [tenancyId, projectUserId], onDelete: Cascade)
 
@@ -625,6 +628,89 @@ model OAuthOuterInfo {
   info       Json
   innerState String   @unique
   expiresAt  DateTime
+
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+}
+
+// =================== SAML SSO ===================
+//
+// SAML connection *config* (entity ID, IdP cert, ACS URL, attribute mapping,
+// domain, etc.) is stored as JSON in tenancy.config.auth.saml.connections —
+// matching how OAuth provider config (clientId/clientSecret) is stored. The
+// Prisma models below cover only per-user account records and the in-flight
+// AuthnRequest temp store.
+
+// One row per (tenancy, samlConnectionId, NameID). Mirrors
+// ProjectUserOAuthAccount. samlConnectionId is the string key under
+// tenancy.config.auth.saml.connections.
+model ProjectUserSamlAccount {
+  id               String  @default(uuid()) @db.Uuid
+  tenancyId        String  @db.Uuid
+  projectUserId    String? @db.Uuid
+  samlConnectionId String
+  nameId           String
+
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  // Snapshot from the most recent assertion. Used to distinguish accounts
+  // and to populate primary contact channel on first sign-in.
+  email        String?
+  nameIdFormat String?
+
+  // Before the SAML account is connected to a user (e.g. mid-flight), the
+  // projectUser is null. Once the ACS callback creates/links the user, this
+  // is set.
+  projectUser ProjectUser? @relation(fields: [tenancyId, projectUserId], references: [tenancyId, projectUserId], onDelete: Cascade)
+
+  // if allowSignIn is true, samlAuthMethod must be set
+  samlAuthMethod SamlAuthMethod?
+  allowSignIn    Boolean         @default(true)
+
+  @@id([tenancyId, id])
+  // A NameID is unique per (project, connection). The same NameID from a
+  // different connection is a distinct identity — this is what enforces
+  // multi-tenant connection isolation at the DB level.
+  @@unique([tenancyId, samlConnectionId, nameId])
+  // Composite for FK relation from SamlAuthMethod.samlAccount.
+  @@unique([tenancyId, samlConnectionId, projectUserId, nameId])
+  @@index([tenancyId, projectUserId])
+  @@index([tenancyId, samlConnectionId])
+}
+
+// Connects a project user to a SamlAuthMethod via a SAML account.
+// Mirrors OAuthAuthMethod.
+model SamlAuthMethod {
+  tenancyId        String @db.Uuid
+  authMethodId     String @db.Uuid
+  samlConnectionId String
+  nameId           String
+  projectUserId    String @db.Uuid
+
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  authMethod  AuthMethod             @relation(fields: [tenancyId, authMethodId], references: [tenancyId, id], onDelete: Cascade)
+  samlAccount ProjectUserSamlAccount @relation(fields: [tenancyId, samlConnectionId, projectUserId, nameId], references: [tenancyId, samlConnectionId, projectUserId, nameId])
+  projectUser ProjectUser            @relation(fields: [tenancyId, projectUserId], references: [tenancyId, projectUserId], onDelete: Cascade)
+
+  @@id([tenancyId, authMethodId])
+  // A user has at most one SAML auth method per connection.
+  @@unique([tenancyId, projectUserId, samlConnectionId])
+  @@unique([tenancyId, samlConnectionId, projectUserId, nameId])
+}
+
+// In-flight AuthnRequest state. Mirrors OAuthOuterInfo. Keyed by the
+// AuthnRequest ID so the ACS handler can look up the original context
+// when the IdP POSTs the assertion back via the browser.
+model SamlOuterInfo {
+  // SAML AuthnRequest IDs are XML IDs (xs:ID). They start with a non-numeric
+  // character per spec, so we store as String — not Uuid.
+  id String @id
+
+  info      Json
+  expiresAt DateTime
 
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
--- a/apps/backend/src/app/api/latest/auth/saml/acs/[connection_id]/route.tsx
+++ b/apps/backend/src/app/api/latest/auth/saml/acs/[connection_id]/route.tsx
@@ -21,6 +21,7 @@ import { createSamlUserAndAccount, findExistingSamlAccount, handleSamlEmailMerge
 import { Tenancy, getTenancy } from "@/lib/tenancies";
 import { oauthServer } from "@/oauth";
 import { getPrismaClientForTenancy, globalPrismaClient } from "@/prisma-client";
+import { Prisma } from "@/generated/prisma/client";
 import { createSmartRouteHandler } from "@/route-handlers/smart-route-handler";
 import { buildSamlClient, extractInResponseTo, parseAndVerifyAssertion, SamlConnectionConfig } from "@/saml/saml";
 import { InvalidClientError, InvalidScopeError, Request as OAuthRequest, Response as OAuthResponse } from "@node-oauth/oauth2-server";
@@ -28,6 +29,7 @@ import { KnownError, KnownErrors } from "@stackframe/stack-shared";
 import { yupMixed, yupNumber, yupObject, yupString } from "@stackframe/stack-shared/dist/schema-fields";
 import { getEnvVariable } from "@stackframe/stack-shared/dist/utils/env";
 import { StackAssertionError, StatusError, captureError } from "@stackframe/stack-shared/dist/utils/errors";
+import { has } from "@stackframe/stack-shared/dist/utils/objects";
 import { deindent } from "@stackframe/stack-shared/dist/utils/strings";
 import { cookies } from "next/headers";
 import { redirect } from "next/navigation";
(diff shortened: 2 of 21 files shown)


apps/backend/src/app/api/latest/auth/saml/login/[connection_id]/route.tsx
```diff
@@ -0,0 +1,182 @@
...
+      id: params.connection_id,
+      displayName: connectionRaw.displayName,
+      idpEntityId: connectionRaw.idpEntityId,
+      idpSsoUrl: connectionRaw.idpSsoUrl,
+      idpCertificate: connectionRaw.idpCertificate,
+      domain: connectionRaw.domain,
+      attributeMapping: connectionRaw.attributeMapping,
+    };
+
+    const baseUrl = new URL(query.redirect_uri).origin;
```
Comment (mantrakp04) [inline apps/backend/src/app/api/latest/auth/saml/login/[connection_id]/route.tsx]: BUG: walkthrough on the SAML SP origin here.

1. In SAML, Stack Auth is the Service Provider, so the SP entity ID and ACS URL need to describe Stack Auth's public SAML endpoints.
2. The metadata route builds those URLs from the API request origin, so the IdP is told to send assertions back to something like `https://api.stack-auth.com/api/v1/auth/saml/acs/...`.
3. This login route builds the AuthnRequest with `new URL(query.redirect_uri).origin`, where `redirect_uri` is the customer's OAuth callback URL, eg `https://app.customer.com/callback`.
4. The ACS route later rebuilds the SAML client from that same stored `redirectUri` origin.
5. That means the IdP can be configured from metadata for Stack Auth's API origin, but login/ACS validation expects the assertion audience/recipient for the customer's app origin.

I think a normal customer callback URL can therefore produce issuer/audience/ACS mismatches. We should derive the SP origin from one canonical Stack Auth public API origin and use it consistently for metadata, AuthnRequest creation, and ACS validation.

apps/backend/src/app/api/latest/saml-connections/route.tsx
```diff
@@ -0,0 +1,177 @@
...
+      }).optional(),
+    }).defined(),
+  }),
+  response: yupObject({
+    statusCode: yupNumber().oneOf([200]).defined(),
+    bodyType: yupString().oneOf(["json"]).defined(),
+    body: samlConnectionResponseShape,
+  }),
+  async handler({ auth, body }) {
+    const overlay: Record<string, unknown> = {
```
mantrakp04: BUG: walkthrough on first-time connection create.

1. This endpoint says it creates or updates a SAML connection.
2. For a brand-new connection, the parent config entry `auth.saml.connections.<id>` does not exist yet.
3. The overlay only writes child dot-paths like `auth.saml.connections.<id>.displayName`.
4. The config code already treats dotting into a missing record entry as invalid/silently ignored during rendering; see the existing warnings for `auth.oauth.providers.google.clientId` and similar paths.
5. So this can return 200 with the response body from the request, while the rendered config still has no usable SAML connection.

I think create needs to write the whole record entry at `auth.saml.connections.<id>` when it does not exist yet. Child dot-path updates are fine after the parent entry exists.

apps/backend/src/saml/saml.tsx
```diff
@@ -0,0 +1,209 @@
...
+  const emailAttr = connection.attributeMapping?.email ?? "email";
+  const displayNameAttr = connection.attributeMapping?.displayName ?? "displayName";
+
+  return {
+    nameId: profile.nameID,
+    nameIdFormat: profile.nameIDFormat,
+    // node-saml types inResponseTo as a loose object; coerce to string defensively.
+    inResponseTo: profile.inResponseTo ? String(profile.inResponseTo) : null,
+    attributes,
+    email: pickFirst(attributes[emailAttr]) ?? profile.nameID,
```
mantrakp04: BUG: walkthrough on using NameID as email.

1. SAML responses have a `NameID`, but that value is not always an email address.
2. Many IdPs use persistent/opaque identifiers as NameID, like `00u123...` or another stable subject ID.
3. This fallback makes `email` become `profile.nameID` whenever the mapped email attribute is missing.
4. ACS later only checks that `assertion.email` is truthy, then passes it into user creation/merge as a verified email.
5. In that case an opaque NameID can become the user's primary email or email merge key.

I think we should reject assertions without a real mapped email attribute, unless the NameID format/value explicitly proves it is an email address. Otherwise this turns an IdP subject identifier into verified email identity.

apps/backend/src/app/api/latest/auth/saml/discover/route.tsx
```diff
@@ -0,0 +1,76 @@
...
+  response: yupObject({
+    statusCode: yupNumber().oneOf([200]).defined(),
+    bodyType: yupString().oneOf(["json"]).defined(),
+    body: yupObject({
+      connection_id: yupString().defined(),
+      display_name: yupString().defined(),
+    }).defined(),
+  }),
+  async handler({ query }) {
+    const tenancy = await getSoleTenancyFromProjectBranch(query.project_id, DEFAULT_BRANCH_ID, true);
```
mantrakp04: Potential bug: walkthrough on branch selection in discovery.

1. Admin CRUD writes SAML config into `auth.tenancy.branchId`, so SAML connections are branch-scoped.
2. Login resolves tenancy from `client_id`, which includes project + branch.
3. Discovery only accepts `project_id` and then hardcodes `DEFAULT_BRANCH_ID`.
4. So if a non-default branch has a SAML connection, discovery will not see it.
5. Worse, discovery can return a default-branch connection even though the eventual login request is for a different branch.

Should discovery accept enough information to resolve the same branch that login will use? Otherwise `signInWithSso({ email })` can pick a different config than the one used by the actual SAML login.

apps/backend/src/app/api/latest/auth/saml/metadata/[connection_id]/route.tsx
```diff
@@ -0,0 +1,68 @@
...
+      project_id: yupString().defined(),
+    }).defined(),
+  }),
+  response: yupObject({
+    statusCode: yupNumber().oneOf([200]).defined(),
+    bodyType: yupString().oneOf(["text"]).defined(),
+    body: yupString().defined(),
+  }),
+  async handler({ params, query }, fullReq) {
+    const tenancy = await getSoleTenancyFromProjectBranch(query.project_id, DEFAULT_BRANCH_ID, true);
```
Comment (mantrakp04) [inline apps/backend/src/app/api/latest/auth/saml/metadata/[connection_id]/route.tsx]: Potential bug: same branch-scoping walkthrough as discovery, but for metadata.

1. SAML connection config is stored under a tenancy/branch.
2. Admin CRUD writes to the current admin tenancy's branch.
3. This metadata route only takes `project_id` and hardcodes `DEFAULT_BRANCH_ID`.
4. An IdP admin trying to configure SAML for a non-default branch can get metadata for the wrong branch, or a 404 even though the connection exists on their branch.
5. Later, login resolves from `client_id` and may use a different branch/config than the metadata that was pasted into the IdP.

I think metadata needs an explicit branch identity, or some other way to guarantee it is returning metadata for the same tenancy that login will use.

apps/backend/src/saml/metadata-parser.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  // Prefer HTTP-Redirect binding (what we send AuthnRequests over), fall back
+  // to HTTP-POST. Most IdPs publish both.
+  const ssoUrl =
+    (SELECT(
+      "string(/md:EntityDescriptor/md:IDPSSODescriptor/md:SingleSignOnService[@Binding='urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect']/@Location)",
+      doc,
+    ) as string).trim()
+    || (SELECT(
```
mantrakp04: Potential bug: walkthrough on POST vs Redirect binding.

1. IdP metadata can advertise different SSO bindings.
2. HTTP-Redirect means the browser goes to the IdP with `SAMLRequest` in the URL query.
3. HTTP-POST means the browser posts a form body to the IdP.
4. This parser prefers Redirect, but falls back to a POST binding URL when Redirect is missing.
5. The login path still calls `buildAuthnRequestUrl()`, which produces a Redirect-style GET URL with query params.
6. So for POST-only IdPs, we can store a POST endpoint and later send it a Redirect-style request it may not support.

I think V1 should either require HTTP-Redirect metadata, or actually implement POST-binding AuthnRequest generation before accepting POST-only metadata.

apps/backend/src/app/api/latest/auth/saml/acs/[connection_id]/route.tsx
```diff
@@ -0,0 +1,314 @@
...
+    const tenancy = await getTenancy(outerInfo.tenancyId);
+    if (!tenancy) {
+      throw new StackAssertionError("Tenancy from SamlOuterInfo not found", { tenancyId: outerInfo.tenancyId });
+    }
+    const prisma = await getPrismaClientForTenancy(tenancy);
+
+    if (!(params.connection_id in tenancy.config.auth.saml.connections)) {
+      throw new StatusError(StatusError.NotFound, `SAML connection ${params.connection_id} not found`);
+    }
+    const connectionRaw = tenancy.config.auth.saml.connections[params.connection_id];
```
Comment (mantrakp04) [inline apps/backend/src/app/api/latest/auth/saml/acs/[connection_id]/route.tsx:156]: Potential bug: walkthrough on disabling a connection mid-flow.

1. Login checks `connectionRaw.allowSignIn === false` and blocks disabled connections.
2. After login starts, the AuthnRequest state lives for 10 minutes.
3. During that window, an admin can disable the SAML connection.
4. The IdP can still post the already-started assertion back to ACS.
5. ACS loads the connection and checks that the IdP fields exist, but it does not re-check `allowSignIn`.
6. If validation succeeds, ACS can still issue the OAuth code.

Since ACS is the final trust point before sign-in completes, I think it should re-check `allowSignIn` there too. Otherwise disabling a connection has a 10-minute grace period for already-started flows.

apps/backend/src/lib/saml-account.tsx
```diff
@@ -0,0 +1,179 @@
...
+  params: {
+    tenancyId: string,
+    samlConnectionId: string,
+    nameId: string,
+    nameIdFormat: string | null,
+    email: string | null,
+    projectUserId: string,
+  },
+): Promise<{ samlAccountId: string }> {
+  const samlAccount = await prisma.projectUserSamlAccount.create({
```
mantrakp04: BUG: walkthrough on atomicity while linking SAML accounts.

1. `linkSamlAccountToUser()` first creates a `ProjectUserSamlAccount`.
2. It then separately creates an `AuthMethod` with nested `SamlAuthMethod`.
3. If the second write fails, for example because a unique constraint is hit, the first write has already committed.
4. That leaves a sign-in-enabled SAML account row without the matching auth-method state.
5. Later, `findExistingSamlAccount()` finds that row and treats it as a valid account.
6. ACS then relies on the account being fully linked and can hit inconsistent state.

This should be one transaction or one nested write pattern where the SAML account and auth method are created together, or neither is created. We should not be able to leave a sign-in-enabled orphaned SAML account.

mantrakp04 (changes requested):

mantrakp04:
@greptile-ai review

apps/backend/src/app/api/latest/auth/saml/login/[connection_id]/route.tsx
```diff
@@ -0,0 +1,186 @@
...
+      return {
+        statusCode: 200,
+        bodyType: "json",
+        body: { location: samlUrl },
+      };
+    }
+
+    // Browser-redirect mode: set a CSRF cookie keyed to the AuthnRequest ID.
+    // The ACS route checks this cookie before honoring the assertion.
+    (await cookies()).set(
```
Comment (mantrakp04) [inline apps/backend/src/app/api/latest/auth/saml/login/[connection_id]/route.tsx:183]: Potential bug: I might be missing something about how we expect the IdP POST to behave in browsers, but I think this cookie needs `sameSite: "none"` for the normal SAML browser flow.

Walkthrough: login sets `stack-saml-inner-*` here, then the user goes to the IdP, then the IdP auto-submits a cross-site POST to ACS. Since this cookie has no explicit SameSite, browsers default it to Lax, and Lax cookies are not sent on cross-site POSTs. ACS then deletes/consumes the `SamlOuterInfo` row before checking the cookie, so the user gets a failed flow and has to restart.

Can we validate this in a real browser test? If the flow does need this cookie, I think production should set `sameSite: "none"` + `secure: true`.

apps/backend/src/saml/saml.tsx
```diff
@@ -0,0 +1,215 @@
...
+function pemToBareCert(pemOrBare: string): string {
+  return pemOrBare
+    .replace(/-----BEGIN CERTIFICATE-----/g, "")
+    .replace(/-----END CERTIFICATE-----/g, "")
+    .replace(/\s+/g, "");
+}
+
+/** Construct a node-saml client for a given connection. Cheap to call per-request. */
+export function buildSamlClient(connection: SamlConnectionConfig, baseUrl: string): SAML {
+  return new SAML({
```
mantrakp04, line 55: Potential bug: `idpEntityId` is stored and parsed, but as far as I can tell it is never actually used during validation. We configure `entryPoint`, cert, SP issuer, callback, and audience, but not the expected IdP issuer.

Wouldn't a response signed by the configured cert but issued under a different IdP entity/app still pass as long as the audience matches? Maybe `node-saml` has issuer validation we can wire up here; otherwise I think ACS should compare the signed Response/Assertion issuer against `connection.idpEntityId`.

apps/backend/src/app/api/latest/auth/saml/acs/[connection_id]/route.tsx
```diff
@@ -0,0 +1,333 @@
...
+  || KnownErrors.SignUpRejected.isInstance(error)
+);
+
+export const POST = createSmartRouteHandler({
+  metadata: { hidden: true },
+  request: yupObject({
+    params: yupObject({
+      connection_id: yupString().defined(),
+    }).defined(),
+    body: yupMixed().defined(),
```
Comment (mantrakp04) [inline apps/backend/src/app/api/latest/auth/saml/acs/[connection_id]/route.tsx]: Potential bug: this public ACS endpoint accepts `yupMixed()` and then casts `body.SAMLResponse` manually below. That means malformed bodies like `SAMLResponse[]=x`, non-base64 values, or unexpectedly large values can reach `extractInResponseTo()` / XML parsing before we return a controlled 400.

Can we make the request schema explicit, roughly `{ SAMLResponse: yupString().defined(), RelayState: yupString().optional() }`, and wrap base64/XML parse failures as `StatusError.BadRequest`? This is an auth endpoint exposed to arbitrary IdP/browser POSTs, so I think it should fail loud but cleanly.

apps/backend/prisma/schema.prisma
```diff
@@ -630,6 +633,89 @@ model OAuthOuterInfo {
...
+// Prisma models below cover only per-user account records and the in-flight
+// AuthnRequest temp store.
+
+// One row per (tenancy, samlConnectionId, NameID). Mirrors
+// ProjectUserOAuthAccount. samlConnectionId is the string key under
+// tenancy.config.auth.saml.connections.
+model ProjectUserSamlAccount {
+  id               String  @default(uuid()) @db.Uuid
+  tenancyId        String  @db.Uuid
+  projectUserId    String? @db.Uuid
```
mantrakp04: Potential bug: this allows a `ProjectUserSamlAccount` with `projectUserId = null` while `allowSignIn` defaults to true below. The code path then treats an existing enabled SAML account as valid, and later has to throw if it has no user.

I think the DB should enforce the invariant described in the comment: if a SAML account is sign-in-enabled, it needs the linked user/auth-method state. Either make `projectUserId` required for persisted SAML accounts, or default unlinked rows to `allowSignIn = false` and add a check constraint so we can't persist a sign-in-enabled orphan.

mantrakp04 (changes requested):

mantrakp04 (approved):

BilalG1:

PR #1397, opened 2026-04-29
feat: SAML SSO client SDK, dashboard, demo, and e2e tests

BilalG1:
Third of 3 stacked PRs adding SAML 2.0 SSO. **Stacked on:** #1396 (backend) → #1395 (mock IdP).

> Note: target branch is `pr/saml-backend`. Will be retargeted to `dev` once #1396 merges.

## Screenshots

The dashboard SSO management page is brand new — there's no \"before\" page to compare to. The screenshots below show the page in its three meaningful states (empty, mid-create, after-add) plus dark mode and the delete confirmation.

### 1. Empty state — no connections configured yet

SAML SSO page: empty state

### 2. \"Add SAML connection\" dialog (clean, before fill)

Add SAML connection dialog

### 3. After adding an Acme connection — surfaces SP metadata URL + ACS URL for the IdP admin to copy

SAML SSO page with one connection

### 4. Same in dark mode

SAML SSO page with one connection (dark mode)

### 5. Delete confirmation — warns that linked user accounts remain in the DB

Delete SAML connection confirmation

> Two small dashboard fixes were caught while taking these screenshots and shipped in commit `fe8197ca8`:
> - Removed `yup.string().url()` from the IdP SSO URL field — it rejects `http://localhost` which broke the dev/test setup. Backend SAML wrapper validates the URL on use.
> - Switched `updateConfig({ pushable: true → false })` because SAML connection fields (cert, IdP URLs) live at the environment level, not the branch level that's pushed to GitHub. Same convention as the auth-methods page does for OAuth client secrets.

---

## What this PR adds

The user-facing surface and verification for the backend SAML feature in #1396.

### Client SDK

`packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts` adds three methods on `StackClientApp`:

- **`signInWithSaml({ connectionId, returnTo })`** — explicit connection selection. Mirrors `signInWithOAuth`.
- **`signInWithSso({ email, returnTo })`** — email-domain discovery via `/auth/saml/discover`, then redirects through the matched connection. Throws if no connection matches so callers can fall back.
- **`getSamlConnectionForEmail(email)`** — pure lookup (no redirect) so the customer's UI can show \"Sign in with Acme SSO\" branding before the user clicks.

Backed by `getSamlUrl` + `authorizeSaml` + `discoverSamlConnection` on `StackClientInterface` (mirrors `getOAuthUrl` + `authorizeOAuth`, without provider_scope or bot challenge — SAML originates from a corporate IdP, not a public form). Generated via `pnpm -w run generate-sdks`; propagates from `packages/template` into `packages/js`, `packages/react`, `packages/stack`.

### Dashboard

`apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sso/` — single page for managing SAML connections on the current project. Reads from `project.useConfig()` (same source as the e2e tests + seed script), writes via `useUpdateConfig`. PEM headers stripped from cert input automatically. Each connection card surfaces SP metadata URL + ACS URL for the customer's IT admin to copy.

V1 limitations (planned follow-ups):
- Edit by delete-and-recreate — no in-place edit dialog yet
- No \"paste IdP metadata XML\" auto-fill (the backend metadata-parser is ready; needs UI wiring)
- No separate detail page
- No sidebar nav link yet — page is reachable by direct URL only (`/projects/<id>/sso`)

### Demo

`examples/demo/src/app/saml-demo/page.tsx` — exercises both SDK flows:
1. Email-domain discovery: enter `alice@acme.test`, click \"Sign in via SSO\"
2. Direct connection ID: per-tenant buttons calling `signInWithSaml({ connectionId })`

Pairs with the seed-dummy-data block from #1395 (gated on `STACK_SEED_ENABLE_SAML=true`).

### E2E tests — 18 tests across 4 files, all passing

`apps/e2e/tests/backend/endpoints/api/v1/auth/saml/`:

| File | Cases |
|---|---|
| `discover.test.ts` | happy path, 404, case-insensitivity, unknown project, cross-project isolation (5) |
| `metadata.test.ts` | XML contains entityID + ACS URL, 404, 404 on incomplete config (3) |
| `login.test.ts` | JSON mode + browser-redirect mode + CSRF cookie + 404 + 403 + invalid client (5) |
| `round-trip.test.ts` | full SP-initiated flow against mock IdP: JIT user creation + wrong-audience + bad-signature (cross-tenant forgery) + expired + replay (5) |

#### Test integrity self-check (verified)

- `grep -r \"from.*backend/src/saml\" apps/e2e/tests/` → 0 matches (tests don't import backend internals)
- Mock IdP uses `samlify`, backend uses `@node-saml/node-saml` — different libraries, so signature/canonicalization bugs would surface as failures rather than canceling out
- Negative cases come from the mock deliberately misbehaving via `/test-controls`, never from injecting bad data into the backend's own validator

## Test plan

- [x] `pnpm --filter @stackframe/template typecheck` + `lint` pass
- [x] `pnpm -w run generate-sdks` runs cleanly; new methods appear in packages/js, packages/react, packages/stack
- [x] `pnpm --filter @stackframe/example-demo-app typecheck` + `lint` pass
- [x] `pnpm --filter @stackframe/dashboard typecheck` + `lint` pass
- [x] All 18 SAML e2e tests pass against live backend + mock IdP locally
- [x] Dashboard SSO page manually exercised against the running dev server (screenshots above)

### DIFF
--- a/apps/backend/src/app/api/latest/saml-connections/route.tsx
+++ b/apps/backend/src/app/api/latest/saml-connections/route.tsx
@@ -12,6 +12,7 @@
  */
 import { overrideEnvironmentConfigOverride, resetEnvironmentConfigOverrideKeys } from "@/lib/config";
 import { createSmartRouteHandler } from "@/route-handlers/smart-route-handler";
+import { KnownErrors } from "@stackframe/stack-shared/dist/known-errors";
 import { adaptSchema, adminAuthTypeSchema, yupArray, yupBoolean, yupNumber, yupObject, yupString } from "@stackframe/stack-shared/dist/schema-fields";
 import { StatusError } from "@stackframe/stack-shared/dist/utils/errors";
 
@@ -47,6 +48,9 @@ export const GET = createSmartRouteHandler({
     }).defined(),
   }),
   async handler({ auth }) {
+    if (!auth.tenancy.config.apps.installed["saml-sso"]?.enabled) {
+      throw new KnownErrors.SamlSsoNotEnabled();
+    }
     const connections = auth.tenancy.config.auth.saml.connections;
     type Conn = (typeof auth.tenancy.config.auth.saml.connections)[string];
     return {
@@ -98,6 +102,9 @@ export const POST = createSmartRouteHandler({
     body: samlConnectionResponseShape,
   }),
   async handler({ auth, body }) {
+    if (!auth.tenancy.config.apps.installed["saml-sso"]?.enabled) {
+      throw new KnownErrors.SamlSsoNotEnabled();
+    }
     const overlay: Record<string, unknown> = {
       [`auth.saml.connections.${body.id}.displayName`]: body.display_name,
       [`auth.saml.connections.${body.id}.allowSignIn`]: body.allow_sign_in,
@@ -160,6 +167,9 @@ export const DELETE = createSmartRouteHandler({
     }).defined(),
   }),
   async handler({ auth, body }) {
+    if (!auth.tenancy.config.apps.installed["saml-sso"]?.enabled) {
+      throw new KnownErrors.SamlSsoNotEnabled();
+    }
     if (!(body.id in auth.tenancy.config.auth.saml.connections)) {
       throw new StatusError(StatusError.NotFound, `SAML connection ${body.id} not found`);
     }
--- a/apps/backend/src/app/api/latest/saml-connections/route.tsx
+++ b/apps/backend/src/app/api/latest/saml-connections/route.tsx
@@ -105,22 +105,31 @@ export const POST = createSmartRouteHandler({
     if (!auth.tenancy.config.apps.installed["saml-sso"]?.enabled) {
       throw new KnownErrors.SamlSsoNotEnabled();
     }
-    const overlay: Record<string, unknown> = {
-      [`auth.saml.connections.${body.id}.displayName`]: body.display_name,
-      [`auth.saml.connections.${body.id}.allowSignIn`]: body.allow_sign_in,
-      [`auth.saml.connections.${body.id}.idpEntityId`]: body.idp_entity_id,
-      [`auth.saml.connections.${body.id}.idpSsoUrl`]: body.idp_sso_url,
-      [`auth.saml.connections.${body.id}.idpCertificate`]: body.idp_certificate,
+    // Write the whole connection entry as a single value. Per-field deep
+    // dot-keys (e.g. `auth.saml.connections.X.displayName`) get dropped
+    // during config normalization with onDotIntoNonObject="ignore" when
+    // the parent record entry doesn't yet exist — same convention as
+    // auth.oauth.providers (see auth-methods/page-client.tsx) and the
+    // dashboard SSO page-client (see commit 5fa9629de).
+    const connectionEntry: Record<string, unknown> = {
+      displayName: body.display_name,
+      allowSignIn: body.allow_sign_in,
+      idpEntityId: body.idp_entity_id,
+      idpSsoUrl: body.idp_sso_url,
+      idpCertificate: body.idp_certificate,
     };
     if (body.domain !== undefined) {
-      overlay[`auth.saml.connections.${body.id}.domain`] = body.domain;
+      connectionEntry.domain = body.domain;
     }
     if (body.attribute_mapping) {
-      overlay[`auth.saml.connections.${body.id}.attributeMapping`] = {
+      connectionEntry.attributeMapping = {
         email: body.attribute_mapping.email,
         displayName: body.attribute_mapping.display_name,
       };
     }
+    const overlay = {
+      [`auth.saml.connections.${body.id}`]: connectionEntry,
+    };
 
     await overrideEnvironmentConfigOverride({
       projectId: auth.tenancy.project.id,
--- a/apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sso/page-client.tsx
+++ b/apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sso/page-client.tsx
@@ -1,0 +1,201 @@
+"use client";
+
+import { SmartFormDialog } from "@/components/form-dialog";
+import { ActionDialog, Alert, Button, Card, CardContent, CardHeader, Typography } from "@/components/ui";
+import { useUpdateConfig } from "@/lib/config-update";
+import React, { useState } from "react";
+import * as yup from "yup";
+import { PageLayout } from "../page-layout";
+import { useAdminApp } from "../use-admin-app";
+
+/**
+ * Dashboard for managing SAML SSO connections on the current project.
+ *
+ * Connection config is stored at tenancy.config.auth.saml.connections —
+ * the same JSON-config the seed script and admin /saml-connections
+ * endpoints write. This page reads via project.useConfig() and writes
+ * via useUpdateConfig() with key paths.
+ *
+ * V1 scope: single-page list with create + delete dialogs. The
+ * paste-IdP-metadata helper that auto-fills idpEntityId/idpSsoUrl/
+ * idpCertificate from a single XML blob is a planned follow-up — for
+ * now connection fields are entered manually.
+ */
+export default function PageClient() {
+  const stackAdminApp = useAdminApp();
+  const project = stackAdminApp.useProject();
+  const config = project.useConfig();
+  const connections = config.auth.saml.connections;
+
+  const [createOpen, setCreateOpen] = useState(false);
+  const [deleteId, setDeleteId] = useState<string | null>(null);
+
+  const connectionEntries = Object.entries(connections);
+
+  return (
+    <PageLayout
+      title="SAML SSO"
+      description="Manage SAML 2.0 connections so corporate IdP users can sign in to your project."
+      actions={<Button onClick={() => setCreateOpen(true)}>Add SAML connection</Button>}
+    >
+      {connectionEntries.length === 0 && (
+        <Alert>
+          No SAML connections configured yet. Click <strong>Add SAML connection</strong> to wire
+          up your first IdP.
+        </Alert>
+      )}
+
+      <div className="grid gap-4">
+        {connectionEntries.map(([id, conn]) => (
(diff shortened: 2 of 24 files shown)


packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -2854,6 +2854,63 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
     await neverResolve();
   }
 
+  /**
+   * Sign in via a configured SAML connection (SP-initiated flow).
+   * Mirrors signInWithOAuth — redirects to /auth/saml/login/[connectionId],
+   * which round-trips through the customer's IdP and back via ACS.
+   */
+  async signInWithSaml(options: {
```
mantrakp04, line 2862: BUG: walkthrough on the SDK generation gap here.

1. This PR adds the SAML methods to `packages/template`.
2. The PR description says `pnpm -w run generate-sdks` propagated them into `packages/js`, `packages/react`, and `packages/stack`.
3. But the diff only changes `packages/template`, and `rg "signInWithSaml|signInWithSso|getSamlConnectionForEmail" packages/js packages/react packages/stack` finds nothing.
4. The demo imports `useStackApp` from `@stackframe/stack`, not from `packages/template`.
5. So the demo cast type-checks locally, but at runtime `app.signInWithSaml` / `signInWithSso` are not actually present on the package it imports.

I think this needs the generated package sources committed too, otherwise the public SDK surface described by this PR does not ship.

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sso/page-client.tsx
```diff
@@ -0,0 +1,220 @@
...
+                />
+                <DetailRow label="Sign-in enabled" value={conn.allowSignIn ? "yes" : "no"} />
+              </div>
+              <div className="mt-3 grid gap-1 text-sm border-t pt-3">
+                <Typography type="p" variant="secondary" className="text-xs">
+                  Paste these into your IdP&apos;s admin console:
+                </Typography>
+                <DetailRow
+                  label="SP metadata URL"
+                  value={`/api/v1/auth/saml/metadata/${id}?project_id=${project.id}`}
```
Comment (mantrakp04) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sso/page-client.tsx]: BUG: walkthrough on the URLs we ask IdP admins to copy.

1. This page renders the SP metadata URL as `/api/v1/auth/saml/metadata/...`.
2. The dashboard runs on the dashboard origin, while the SAML metadata/ACS routes live on the backend API origin.
3. There is no dashboard rewrite for `/api/v1/...`, and IdP consoles generally store exactly what the admin pastes.
4. So copying this relative URL from the dashboard can easily configure the IdP against the dashboard host instead of the Stack API host.
5. The ACS URL below has the same problem.

Can we build these from `NEXT_PUBLIC_STACK_API_URL` / browser API base URL and show absolute URLs? For SAML setup, the origin is part of the contract.

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sso/page-client.tsx
```diff
@@ -0,0 +1,220 @@
...
+        // config normalization when the parent record entry doesn't yet
+        // exist — same convention as auth.oauth.providers in the
+        // auth-methods page.
+        await updateConfig({
+          adminApp: stackAdminApp,
+          configUpdate: {
+            [`auth.saml.connections.${values.id}`]: {
+              displayName: values.displayName,
+              allowSignIn: values.allowSignIn,
+              domain: values.domain || undefined,
```
Comment (mantrakp04) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/sso/page-client.tsx]: Potential bug: walkthrough on domain normalization.

1. The form accepts `domain` as a free text string.
2. We save it directly as `values.domain || undefined`.
3. Discovery lowercases during matching, but it does not trim stored whitespace.
4. So `acme.test ` looks fine in the dashboard card, but `alice@acme.test` will not match it.
5. Same thing applies to uppercase/mixed input if we ever compare before lowercasing elsewhere.

I think we should normalize this on write, probably `trim().toLowerCase()` and convert empty string to `undefined`.

apps/backend/src/app/api/latest/saml-connections/route.tsx
```diff
@@ -98,22 +102,34 @@ export const POST = createSmartRouteHandler({
...
+      connectionEntry.domain = body.domain;
     }
     if (body.attribute_mapping) {
-      overlay[`auth.saml.connections.${body.id}.attributeMapping`] = {
+      connectionEntry.attributeMapping = {
         email: body.attribute_mapping.email,
         displayName: body.attribute_mapping.display_name,
       };
     }
+    const overlay = {
```
mantrakp04: BUG: walkthrough on the admin upsert semantics.

1. Switching to whole-entry writes fixes first-time create, because deep dot paths into a missing record entry get dropped.
2. But this endpoint is documented as create-or-update / idempotent upsert.
3. `domain` and `attribute_mapping` are optional request fields.
4. On an update to an existing connection, this now writes a whole new object that only includes those fields if the request included them.
5. So a caller updating the IdP cert or display name without resending `domain` silently deletes the existing discovery domain. Same for attribute mapping.

I think create and update need different behavior: whole-entry write for create, but merge/preserve existing optional fields for updates unless the request explicitly clears them.

apps/e2e/tests/backend/endpoints/api/v1/auth/saml/round-trip.test.ts
```diff
@@ -0,0 +1,250 @@
...
+}
+
+function loginQuery() {
+  const projectKeys = backendContext.value.projectKeys;
+  if (projectKeys === "no-project") throw new Error("No project keys");
+  const branchId = backendContext.value.currentBranchId;
+  return {
+    client_id: !branchId ? projectKeys.projectId : `${projectKeys.projectId}#${branchId}`,
+    client_secret: projectKeys.publishableClientKey ?? "",
+    redirect_uri: "http://localhost:8101/handler/oauth-callback",
```
mantrakp04: Potential bug: walkthrough on the custom-port coverage.

1. This test hardcodes `http://localhost:8101/handler/oauth-callback`.
2. The custom-base-port workflow runs dashboard on `6701`.
3. Because `allow_localhost` accepts localhost broadly, the SAML flow can still return an OAuth code even when the callback URL points at the wrong dev port.
4. The happy-path assertion only checks that some `code` exists, not that the redirect went to the active dashboard origin.
5. `login.test.ts` has the same hardcoded `8101` redirect URI.

Can we use `localhostUrl("01", "/handler/oauth-callback")` / `localRedirectUrl` and assert the final callback origin? Otherwise the custom-port job can pass while this flow is wired to the default-port dashboard.

apps/e2e/tests/backend/endpoints/api/v1/auth/saml/round-trip.test.ts
```diff
@@ -0,0 +1,250 @@
...
+  await niceFetch(`${MOCK_SAML_BASE}/idp/${tenantSlug}/test-controls`, {
+    method: "POST",
+    body: JSON.stringify(body),
+    headers: { "content-type": "application/json" },
+  });
+}
+
+// ---------- tests ----------
+
+it("full SP-initiated round trip: new user JIT-created", async ({ expect }) => {
```
mantrakp04, line 168: Suggestion: walkthrough on what this happy path proves.

1. The test name says “new user JIT-created”.
2. The assertion only checks that ACS returns a redirect with an OAuth `code`.
3. That proves the authorization flow completed, but it does not directly prove the SAML account/user link was persisted correctly.
4. We already had review concerns around partial SAML account linking and orphaned account state.
5. A regression there could be easier to catch if this test also exchanged/used the code or queried the resulting user/account state.

Could we add one post-ACS assertion that verifies the created user/email or SAML account association exists? That would make the round-trip test cover the durable side effect, not only the redirect.

apps/e2e/tests/backend/endpoints/api/v1/auth/saml/round-trip.test.ts
mantrakp04: i think the fallback test might fail cuz of this

mantrakp04 (changes requested):

mantrakp04:
@greptile-ai review

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
BilalG1: packages/js, react, stack are not commited, only generated

mantrakp04 (approved):

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -2854,6 +2854,66 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
...
+
+    const location = await this._interface.authorizeSaml({
+      connectionId: options.connectionId,
+      redirectUrl: constructRedirectUrl(this.urls.oauthCallback, "redirectUrl"),
+      errorRedirectUrl: constructRedirectUrl(this.urls.error, "errorRedirectUrl"),
+      afterCallbackRedirectUrl,
+      codeChallenge,
+      state,
+    });
+    await this._redirectTo({ url: location });
```
mantrakp04, line 2887: Potential bug: validate that `_redirectTo()` is a no-op when `redirectMethod === "none"`, so this SAML flow can hang forever in that mode.

Walkthrough: an app constructs `StackClientApp({ redirectMethod: "none" })`, calls `signInWithSaml()` (or `signInWithSso()`, which delegates here), `authorizeSaml()` returns a location, `_redirectTo({ url: location })` just returns, and then we hit `await neverResolve()`. No navigation happened and the public SDK call never returns control, so the caller/button can stay stuck forever.

Can we either make `_redirectTo()` report whether navigation actually started and only `neverResolve()` in that case, or explicitly reject this flow for `redirectMethod: "none"` with a clear error / separate URL-returning API?

mantrakp04 (approved):

BilalG1:

PR #1401, opened 2026-04-30
[Docs Mintlify] - Updates and new additions

madster456:
## Summary

Refreshes the docs around Stack Auth setup, CLI workflows, local development, the local emulator, known SDK errors, self-hosting, and the public showcase. This also wires the new docs into Mintlify navigation and normalizes `sharp` dependency resolution for docs/image tooling.

Base: `dev` -> Head: `docs-mintlify/updates`  
Scope: 17 files, +1154 / -435

## What's New

- Adds a dedicated **Stack CLI** guide covering install, auth, init modes, project commands, config pull/push, `stack exec`, and emulator commands.
- Adds a full **Local Emulator** guide for QEMU requirements, ports, default credentials, config-file backed projects, image pulls, state, and troubleshooting.
- Reworks **Local Development** around two supported workflows: cloud-backed local dev and emulator-backed local dev, including app env vars, local config files, CI usage, and common failure modes.
- Rewrites **Self-host** around the supported `stackauth/server` Docker deployment path, including Postgres, ClickHouse, cron scheduling, seeded admin access, reverse proxy setup, SDK env vars, email, webhooks, S3 storage, upgrades, and common issues.
- Adds a **Known Errors** reference for public SDK-exposed known errors, runtime `errorCode` values, and REST API handling.
- Clarifies **CLI App Authentication** so users can distinguish authenticating their own CLI app from using the official `stack` command.
- Updates the JWT guide to remove the missing inline viewer reference and recommend an external JWT viewer.
- Adds showcase cards for Browser Use and Overworld with supporting images and styles.
- Pins `sharp` to `0.34.5` through pnpm overrides and lockfile cleanup.

## Review Notes

- The self-host guide was audited against the current Docker entrypoint, server env templates, seed script, ClickHouse migration behavior, cron endpoints, and SDK API URL env resolution.
- The Docker image starts the backend and dashboard, but not production schedulers, so the new cron section is called out explicitly.
- Managed Domain email setup is documented as operator-managed because it depends on server-side Resend/DNSimple credentials; self-hosters are directed toward Custom SMTP or their own Resend API key.
- `self-host-old.mdx` is kept as a legacy reference file and is not added to navigation.
- `emulator run` documentation now matches CLI behavior: it stops the emulator only when it started that emulator instance.

## Test Plan

- [x] Reviewed all files changed by `origin/dev...HEAD`.
- [x] Ran `git diff --check origin/dev...HEAD`.
- [x] Checked IDE diagnostics for the changed docs/CLI files.
- [ ] Preview Mintlify docs locally and click through new navigation entries.
- [ ] Verify showcase cards and images in light and dark themes.
- [ ] Smoke-test the copied self-host commands in a non-production Docker environment.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive Stack CLI, Local Emulator, Known Errors, and Local Development guides
  * Restructured Self-Hosting guide for production deployments and expanded authentication docs
  * Updated site navigation to include new guide pages

* **New Features**
  * Added visual showcase section with responsive cards and hover/zoom interactions (and supporting styles)


madster456:
Added a root `pnpm.overrides.sharp` entry so all transitive `sharp` consumers resolve to the same safe stable line (`^0.34.5`).

Without the override, Mintlify pulls an older `sharp`/`libvips` path while the repo also has a newer one, so running `docs-mintlify` can load two native `libvips` builds in the same process:

```text
Class GNotificationCenterDelegate is implemented in both
@img/sharp-libvips-darwin-arm64@1.0.4/.../libvips-cpp.42.dylib
and
@img/sharp-libvips-darwin-arm64@1.2.4/.../libvips-cpp.8.17.3.dylib

This may cause spurious casting failures and mysterious crashes.
One of the duplicates must be removed or renamed.

File: docs-mintlify/guides/other/self-host.mdx [hunk 1]
```diff
@@ -20,149 +20,290 @@ If you are unsure whether you should self-host, here are some things to consider
...
+At minimum, set:
+
+```env title="stack-auth.env"
+NEXT_PUBLIC_STACK_API_URL=https://auth-api.example.com
+NEXT_PUBLIC_STACK_DASHBOARD_URL=https://auth.example.com
+
+STACK_DATABASE_CONNECTION_STRING=postgresql://postgres:password@stack-auth-postgres:5432/stackframe
+STACK_SERVER_SECRET=replace-with-a-32-byte-base64url-secret
+CRON_SECRET=replace-with-a-long-random-secret
+
```
mantrakp04, line 91: Bug: The smoke-test ClickHouse container above creates `CLICKHOUSE_DB=analytics`, but this env block never sets `STACK_CLICKHOUSE_DATABASE=analytics`. As far as I can tell the backend will target its default database instead of the one the example created.

docs-mintlify/guides/going-further/local-emulator.mdx
```diff
@@ -0,0 +1,220 @@
...
+
+- QEMU for your architecture
+- `qemu-img`
+- `socat`
+- `curl`
+- `nc`
+- `lsof`
+- `openssl`
+
+`zstd` is optional but recommended for compressed snapshot handling.
```
mantrakp04: `zstd` does not seem optional for the default fresh-start path. `emulator start` captures a snapshot, and the capture path calls `zstd`, so a user following this without installing it can fail instead of just losing compressed snapshot handling.

docs-mintlify/guides/other/self-host-old.mdx
```diff
@@ -0,0 +1,168 @@
+---
```
mantrakp04: accident? this adds an orphan copy of the old self-host docs that is not in `docs.json`, and it still contains stale local-development/build instructions. I think we should delete it unless this is intentionally linked by some archive/redirect flow.

mantrakp04 (approved):

docs-mintlify/guides/other/self-host-old.mdx
madster456: Was left for archival purposes, my intention was to remove before merge. Just mental reference while building out the new one.

PR #1410, opened 2026-05-05
fix(oauth): stop forcing prompt=consent on non-Google providers

alvgaona:
The base OAuth provider hardcoded `prompt=consent` and `access_type=offline` on every authorize URL. Both are Google-specific:

- `prompt=consent` caused GitHub, Microsoft, Discord, GitLab, Apple, and X to re-show their authorization screen on **every** sign-in instead of only on first authorization or scope change.
- `access_type=offline` is a Google-only param that was silently ignored by every other provider.

This PR moves both params into Google's `authorizationExtraParams` so Google's behavior is unchanged (refresh token still reliably issued via `prompt=consent` + `access_type=offline`). Other providers now follow standard OAuth behavior: consent shown only on first authorization, new scopes, or after the user revokes the grant on the provider side.

### Behavioral change per provider

| Provider | Before | After |
|---|---|---|
| GitHub, Microsoft, Discord, GitLab, Apple, X | Consent every sign-in | Consent only on first auth / new scope / after revoke |
| Spotify, Twitch, Facebook, LinkedIn, Bitbucket | Param ignored | Param ignored (no change) |
| Google | `prompt=consent` + `access_type=offline` | Same — moved into Google's own params |

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed OAuth authorization flow to use provider-specific parameter configurations instead of applying parameters globally, improving compatibility across different OAuth providers and ensuring proper authorization handling for each provider.

### DIFF
--- a/apps/backend/src/oauth/providers/base.tsx
+++ b/apps/backend/src/oauth/providers/base.tsx
@@ -204,8 +204,6 @@ export abstract class OAuthBaseProvider {
       }),
       state: options.state,
       response_type: "code",
-      access_type: "offline",
-      prompt: "consent",
       ...this.authorizationExtraParams,
     });
   }
--- a/apps/backend/src/oauth/providers/google.tsx
+++ b/apps/backend/src/oauth/providers/google.tsx
@@ -24,6 +24,7 @@ export class GoogleProvider extends OAuthBaseProvider {
       baseScope: "https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/userinfo.profile",
       authorizationExtraParams: {
         prompt: "consent",
+        access_type: "offline",
         include_granted_scopes: "true",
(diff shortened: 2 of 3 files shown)


mantrakp04 (approved):

N2D4:
This is intended behavior. Is there a concrete use case you're looking for here? Maybe we can make this configurable on a per-provider level.

alvgaona:
> This is intended behavior. Is there a concrete use case you're looking for here? Maybe we can make this configurable on a per-provider level.

@N2D4 

I find it a bit weird that we have to authorize the app every time we sign in when it was already authorized previously. Say I sign in, then sign out, then I sign in again I wouldn't need to authorize the app again, right? Unless, the permissions were revoked directly.

I might be missing something though.

PR #1418, opened 2026-05-06
Custom Dashboards Versioning fix

aadesh18:
This PR fixes the versioning error that we ran into for custom dashboards. Now if the latest version of the packages does not work, we fall back to the version that is one patch below the latest version. We log this into sentry. If the fall back doesn't work either, we log that into sentry as well and show the user an error message.

Apart from that, I also made changes to ensure dashboards with older versions of the dashboard-ui-component package would still work. Each dashboard now stores the version it was created with, as a comment at the top of its source code, and we use that version when loading the dashboard. When a dashboard gets edited via the AI chat, we re-stamp it with the latest version of the package so it stays up to date.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling and reporting for dashboard load failures; host surfaces structured dependency errors for faster diagnostics.
  * Added automatic fallback loading for missing resources to reduce load failures.
  * Fixed page height calculation so pages align correctly with the viewport.

* **New Features**
  * Generated and editor-provided dashboard code is now stamped with the app version for clearer provenance.

* **UI/UX Improvements**
  * Clearer, more informative error messages when custom dashboard loading encounters issues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/hexclave/stack-auth/pull/1418)


aadesh18:
@greptile review again

aadesh18:
@greptile review again

N2D4 (approved):

PR #1419, opened 2026-05-06
stack-cli: support self-hosted URLs and tighten CLI auth polling

mantrakp04:
## Summary
- **Self-hosted CLI**: read `STACK_API_URL` / `STACK_DASHBOARD_URL` from env in `stack-cli` so the published CLI can talk to self-hosted Stack Auth installs without a custom build. The existing `STACK_CLI_PUBLISHABLE_CLIENT_KEY` override is kept as-is.
- **Docker example**: surface the three CLI-relevant vars in `docker/server/.env.example` so self-host operators see them.
- **Tighter polling-code TTL**: default `2h -> 2min`, max `24h -> 15min` for the CLI auth polling code. The code is only valid while a user is actively waiting in `stack login`, so a tight window limits the blast radius of a leaked code.
- **Raw-SQL poll handler**: convert `apps/backend/src/app/api/latest/auth/cli/poll/route.tsx` from `prisma.cliAuthAttempt.*` to raw SQL targeted at the tenancy source-of-truth schema, matching the pattern already used by the initiate handler in `apps/backend/src/app/api/latest/auth/cli/route.tsx`.

## Test plan
- [ ] `pnpm typecheck`
- [ ] `pnpm lint`
- [ ] `pnpm test run` (focus on CLI-auth tests if any)
- [ ] Manual: `stack login` against a local backend
  - polling code now expires after ~2 minutes by default
  - `waiting` / `success` / `used` / `expired` branches still return correct status codes and bodies
- [ ] Manual: published `stack-cli` against a self-hosted backend with `STACK_API_URL` / `STACK_DASHBOARD_URL` set, end-to-end login

## Summary by CodeRabbit

* **Improvements**
  * More robust CLI authentication polling with atomic database updates to prevent races; returns explicit statuses (waiting/expired/used/success) and provides the refresh token on success.

* **Changes**
  * Default CLI auth token TTL reduced to 2 minutes and capped at 15 minutes.
  * Anonymous refresh token is considered present only when not null; null expiry is treated as not-expired.


docker/server/.env.example
```diff
@@ -2,6 +2,9 @@
 
 NEXT_PUBLIC_STACK_API_URL=http://localhost:8102
 NEXT_PUBLIC_STACK_DASHBOARD_URL=http://localhost:8101
+STACK_API_URL=http://localhost:8102
+STACK_DASHBOARD_URL=http://localhost:8101
+STACK_CLI_PUBLISHABLE_CLIENT_KEY=
```
BilalG1: don't think we need these, self-hosters would pass in these when running cli instead

apps/backend/src/app/api/latest/auth/cli/route.tsx
```diff
@@ -25,7 +25,7 @@ export const POST = createSmartRouteHandler({
       tenancy: adaptSchema.defined(),
     }).defined(),
     body: yupObject({
-      expires_in_millis: yupNumber().max(1000 * 60 * 60 * 24).default(1000 * 60 * 120), // Default: 2 hours, max: 24 hours
+      expires_in_millis: yupNumber().max(1000 * 60 * 60 * 24).default(1000 * 60 * 2), // Default: 2 minutes, max: 24 hours
```
BilalG1: 2 min might be too short, apparently 10 or 15 min is common

BilalG1 (approved):

mantrakp04:
@greptile-ai I've addressed all the review feedback:

- **TOCTOU race in `cli/poll/route.tsx`** (P1, also flagged by Copilot): replaced the SELECT-then-UPDATE with a single atomic `UPDATE ... WHERE "usedAt" IS NULL RETURNING "refreshToken"`, so concurrent polls can't both claim the same refresh token.
- **`DEFAULT_API_URL` / `DEFAULT_DASHBOARD_URL` double-lookup** (P2): reverted the constants to plain string literals; env-var resolution stays in `resolveApiUrl()` / `resolveDashboardUrl()` as the single source of truth.
- **Bilal's notes**: removed the new lines from `docker/server/.env.example` (self-hosters pass these to the CLI), and bumped the default `expires_in_millis` from 2 min to 10 min.
- **CodeRabbit nits**: switched the truthy checks on `anon_refresh_token`, `refreshTokenObj.expiresAt`, `cliAuth.usedAt`, and `cliAuth.refreshToken` to explicit `!= null` / `=== null`.

Please take another look.

PR #1429, opened 2026-05-13
payments: rework refund flow to three-knob API

BilalG1:
## Summary
- Replaces per-entry refund schema with a flat `{ amount_usd, revoke_product, end_subscription? }` shape; refund state is now derived from bulldozer ledger rows (`refund:<sourceTxnId>:<uuid>`) instead of the legacy `refundedAt` column, enabling multiple partial refunds up to the remaining cap.
- Adds `invoice_id` for refunding any subscription invoice (start or renewal), Stripe idempotency keys derived from `(tenancyId, sourceTxnId, amount, prior_refunded)` so retries dedupe but intentional partials don't collide, and a legacy backstop that rejects pre-rework `refundedAt` purchases.
- Dashboard refund dialog rebuilt around the three toggles (revoke→end coupling cascades into the UI); refund rows surface in the listing as `type: "refund"` with `adjusted_by` linkage handling both new and legacy formats.

## Implements
[STA2-52 — Build in refund logic for payments](https://linear.app/stack-auth/issue/STA2-52/build-in-refund-logic-for-payments)

## Documented limitations (planned follow-up work)
These are called out in code comments and intentionally deferred to a follow-up PR:
- **Cap-check race under concurrent refunds.** Bulldozer's embedded `BEGIN/COMMIT` prevents an outer Prisma tx from scoping the writes, so two concurrent refunds can both pass the cap check. Needs a bulldozer-aware mutex or pending-refund-intent pattern. In practice refunds are admin-only and rare, so the race window is small.
- **Stripe + DB non-atomicity on the DB-success → response-loss path.** The Stripe idempotency key is keyed on `(tenancyId, sourceTxnId, amount, priorRefunded)`, so a retry after Stripe-success → DB-fail self-heals (Stripe dedupes; the next attempt writes the bulldozer row). The hole is the reverse direction: if the bulldozer row commits but the response is lost, a retry sees a higher `priorRefunded` and generates a fresh key — Stripe would issue a second real refund. No out-of-band reconciliation today.
- **Dashboard can't reach the `invoice_id` path.** Refund actions are only enabled on `purchase` rows and the submit call never passes `invoice_id`, so admins refunding a renewal must use the API directly. Follow-up: enable the action on `subscription-renewal` rows and thread `invoice_id` through.

## Architectural note
`active-subscription-end` and `item-quantity-expire` entries are **not** emitted on the refund row itself. They're produced by the derived sub-end transaction (`transactions.ts:158-228`) once Prisma `subscription.endedAt` is updated, keeping the `expiresWhen` / `when-repeated` semantics in one place. This is the main structural divergence from the ticket's literal entry recipe.

## Review follow-ups addressed in this PR

**First-pass review:**
- **KnownError back-compat preserved**: `SubscriptionAlreadyRefunded` / `OneTimePurchaseAlreadyRefunded` are once again thrown by the legacy-`refundedAt` backstop, and `TestModePurchaseNonRefundable` is thrown when an admin sends `amount_usd > 0` against a test-mode purchase. Callers catching by error code keep working through the rework.
- **Idempotency-key comment corrected**: now accurately describes the `(tenancyId, sourceTxnId, amount, priorRefunded)` key and its self-healing behaviour on the Stripe-success → DB-fail retry path (see Documented limitations above for the remaining hole).
- **Renewal-invoice e2e coverage added**: new test sets up a live-mode subscription via Stripe webhooks (`subscription_create` + `subscription_cycle` invoices), refunds the renewal invoice via `invoice_id`, and asserts the resulting `refund_transaction_id` starts with `refund:sub-renewal:` and is linked back via `adjusted_by` on the *renewal* row (not the start row). Plus negative cases: cross-subscription `invoice_id` → 404, `invoice_id` on a one-time purchase → SchemaError.

**Second-pass review:**
- **Idempotent sub-cancel error-code string fix**: the Stripe code for re-cancelling an already-canceled sub is `subscription_already_canceled`, not `subscription_canceled` — the previous catch would have re-thrown.
- **End-only sub refund replay rejected**: when `amount=0, revoke=false, end=true` and the sub is already `cancelAtPeriodEnd` or `endedAt`, throw SchemaError. Otherwise `readPriorRefundSummary` doesn't see end-only events and the call would be a forever-no-op accumulating empty refund rows.
- **`revoke_product=true` with renewal `invoice_id` rejected**: the product grant lives on the sub-start txn, not on renewal txns — a renewal-scoped revocation would write a back-reference to a non-existent entry. Forces admin to revoke against the start invoice (or the default no-`invoice_id` call).
- **Refund row `id` matches the linkage**: the listing route now returns the full refund txnId as `id` for `type: "refund"` rows so it matches `adjusted_by.transaction_id` — the dashboard can join source rows to their refund rows.
- **+2 e2e tests** for the above (end-only replay rejection, revoke+renewal rejection).

**Third-pass review:**
- **Dashboard refund dialog seeds state on open**: previously the reset block lived in `ActionDialog`'s `onOpenChange`, which doesn't fire on the open transition for a controlled dialog. As a result the dialog opened with the initial `useState` defaults (`amountUsd = '0'`), and an admin submitting unchanged on a paid purchase would revoke/end at $0 instead of refunding the charged amount. The seed now runs in the menu `onClick` before `setIsDialogOpen(true)`.
- **`SUBSCRIPTION_START_PRODUCT_GRANT_ENTRY_INDEX` corrected from 1 → 0**: the constant is persisted as `adjustedEntryIndex` on product-revocation entries and copied through verbatim by `mapLedgerEntry`. That mapper drops the hidden `active-subscription-start` entry, so the public-API layout puts the product grant at index 0. The prior value of `1` pointed at the money-transfer entry (or out of range on test-mode subs) through the public listing.
- **`amountTotal` cap gated behind a USD pre-flight**: `SubscriptionInvoice` doesn't persist invoice currency, and the previous code took `invoice.amountTotal` as USD cents directly. Now `getTotalUsdStripeUnits` (which throws on non-USD pricing) is always called first; `amountTotal` is only preferred as the actual cap after that pre-flight succeeds.

## Test plan
- [x] `pnpm typecheck` — 28/28 pass
- [x] `pnpm lint` — 28/28 pass
- [x] `pnpm test run apps/e2e/tests/backend/endpoints/api/v1/internal/transactions-refund.test.ts` — **19/19 pass** (was 14/14 on the original PR; +3 for `invoice_id` path: renewal refund happy path, unrelated `invoice_id` rejection, `invoice_id` on OTP rejection; +2 for second-pass: end-only replay rejection, revoke+renewal rejection)
- [x] curl smoke against `/api/latest/internal/payments/transactions/refund` — unknown purchase → 404, no-op → 400, negative → 400, sub-revoke-without-end → 400
- [x] **Dashboard UI end-to-end re-run pending** — the original agent-browser pass ran before the third-pass dialog-seed fix, so any "money + revoke" submissions may have actually sent `amount_usd = "0"`. Re-test before un-drafting: open the refund dialog from the menu, confirm the amount field pre-fills with the charged amount, exercise validation (negative / exceeds-cap / no-op), and submit both an end-subscription-only sub refund and a money+revoke OTP refund; verify bulldozer rows and Prisma `cancelAtPeriodEnd` updates.

## Summary by CodeRabbit

* **New Features**
  * Ledger-driven refund flow with stable refund IDs, invoice-aware refunds, OTP/product-revocation support, tri-state end_action (now / at-period-end / none), and API responses that include refund_transaction_id.

* **Bug Fixes / Improvements**
  * Deterministic Stripe idempotency, stronger replay protection, refundable-amount caps, test-mode constraints, and transactions listing updated to surface refunds.

* **Tests**
  * Expanded unit and E2E coverage for new request shape, invoice paths, money-unit conversion, and edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/hexclave/stack-auth/pull/1429)


apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
```diff
@@ -39,163 +41,202 @@ export function buildStripeRefundParams(args: {
...
+ */
+function makeStripeIdempotencyKey(args: {
+  tenancyId: string,
+  sourceTxnId: string,
+  amountStripeUnits: number,
+  priorRefundedStripeUnits: number,
+}): string {
+  const fingerprint = `${args.tenancyId}:${args.sourceTxnId}:${args.amountStripeUnits}:${args.priorRefundedStripeUnits}`;
+  return `refund:${createHash("sha256").update(fingerprint).digest("hex").slice(0, 32)}`;
 }
```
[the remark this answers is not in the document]
BilalG1: Known gap — documented inline at lines 330–341. Stable-token fix deferred to a follow-up.

nams1570:
@greptileai please rereview comprehensively

apps/backend/src/lib/payments/schema/phase-1/transactions.ts
```diff
@@ -27,6 +27,20 @@ import type { SeedEventsStoredTables } from "./stored-tables";
...
+// Position of the product-grant entry as exposed by the public transactions
+// API. Refund product-revocation rows persist `adjustedEntryIndex` purely
+// as a back-reference read by SDK consumers — the value is copied through
+// `mapLedgerEntry` verbatim. That mapper drops the hidden
+// `active-subscription-start` entry, so the public-API layout is:
+//   subscription-start: [product_grant, money_transfer?, ...]
+//   one-time-purchase:  [product_grant, money_transfer?, ...]
+// Both product grants land at index 0 publicly. If the public mapping
+// changes (e.g. `active-subscription-start` becomes visible), these need
+// to move in lockstep and any persisted refund rows reconciled.
```
nams1570, line 43: I think we want to see active-subscription-start and active-subscription-end on the txn log on the frontend. Doesnt need to be done on this PR though

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
```diff
@@ -205,230 +246,492 @@ export const POST = createSmartRouteHandler({
...
+    // outstandingGrants state.
+    if (!isTestMode && subscription.stripeSubscriptionId) {
+      const stripe = await getStripeForAccount({ tenancy });
+      // Idempotent cancel: the Stripe sub may already be canceled (natural
+      // end before this refund). `resource_missing` is what Stripe returns
+      // when the sub no longer exists; `subscription_already_canceled` is
+      // the documented code for re-cancel on an existing-but-canceled sub.
+      // Neither is an error from our perspective.
+      try {
+        await stripe.subscriptions.cancel(subscription.stripeSubscriptionId);
```
nams1570, line 750: I feel like ending subscription should just be handled by the should subscription end flag, not by the revoke product flag. Maybe get rid of this

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
```diff
@@ -205,230 +246,492 @@ export const POST = createSmartRouteHandler({
...
+  } else if (options.endSubscription) {
+    // End at period end. Items follow natural lifecycle when sub-end fires
+    // at period boundary.
+    if (!isTestMode && subscription.stripeSubscriptionId) {
+      const stripe = await getStripeForAccount({ tenancy });
+      await stripe.subscriptions.update(subscription.stripeSubscriptionId, {
+        cancel_at_period_end: true,
       });
-      await bulldozerWriteManualTransaction(prisma, manualRefund.rowId, manualRefund.rowData);
     }
```
nams1570: It would be nice to instead have endSubscription have two options- end sub at period end or end sub immediately. For the latter, think of scenario where end user buys product A in product line, realizes it's not what they want, gets a refund, and then tries to buy product b in product line. If cancelAtPeriodEnd = true is the only way to end a sub via refund, end user is blocked fro purchasing b

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
```diff
@@ -205,230 +246,492 @@ export const POST = createSmartRouteHandler({
...
+    updatedSub = await prisma.subscription.update({
+      where: { tenancyId_id: { tenancyId: tenancy.id, id: subscription.id } },
+      data: {
+        cancelAtPeriodEnd: true,
+        canceledAt: subscription.canceledAt ?? now,
+        endedAt: subscription.endedAt ?? subscription.currentPeriodEnd,
+      },
```
nams1570, line 803: potential bug: for stripe backed subscriptions, shouldn't `endedAt` only be set by the webhook handler? remember that stripe's `canceled` status is akin to our ended status.

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
```diff
@@ -205,230 +246,492 @@ export const POST = createSmartRouteHandler({
...
+      data: {
+        // Don't touch `cancelAtPeriodEnd` — it's meaningless once `endedAt`
+        // is in the past, and writing `true` alongside an immediate `endedAt`
+        // creates inconsistent state for any reader that consults the flag
+        // without joining `endedAt`.
+        status: "canceled",
+        canceledAt: subscription.canceledAt ?? now,
+        endedAt: subscription.endedAt ?? now,
+      },
+    });
```
nams1570, line 773: see other comment about endedAt

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
```diff
@@ -39,163 +41,202 @@ export function buildStripeRefundParams(args: {
...
   const usdPrice = selectedPrice?.USD;
   if (typeof usdPrice !== "string") {
-    throw new KnownErrors.SchemaError("Refund amounts can only be specified for USD-priced purchases.");
+    throw new KnownErrors.SchemaError("Refunds are only supported for USD-priced purchases.");
   }
   if (!Number.isFinite(options.quantity) || Math.trunc(options.quantity) !== options.quantity) {
     throw new StackAssertionError("Purchase quantity is not an integer", { quantity: options.quantity });
   }
   return moneyAmountToStripeUnits(usdPrice as MoneyAmount, USD_CURRENCY) * options.quantity;
 }
```
nams1570, line 83: Potential bug: Have a question about the money units. It seems like conversions to stripe units are underweighted if the right amount of decimal places aren't there?  
Here's something I ran locally, and here are the results

```
"1"     -> 1       (should be 100  for $1.00)
"1.5"   -> 15      (should be 150  for $1.50)
"0.5"   -> 5       (should be 50   for $0.50)
"50"    -> 50      (should be 5000 for $50.00)
"50.00" -> 5000    OK
"0.01"  -> 1       OK
"100"   -> 100     (should be 10000 for $100.00)
```

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
nams1570: Unless you feel like there's a two generals risk here?

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
```diff
@@ -205,230 +246,492 @@ export const POST = createSmartRouteHandler({
...
-        paymentIntentId: purchase.stripePaymentIntentId,
-        amountStripeUnits: refundAmountStripeUnits,
+        amountStripeUnits: options.amountStripeUnits,
         metadata: {
-          tenancyId: auth.tenancy.id,
-          purchaseId: purchase.id,
+          tenancyId: tenancy.id,
+          subscriptionId: subscription.id,
+          ...(invoice ? { invoiceId: invoice.id } : {}),
         },
```
nams1570, line 718: Discussion: do you think the refundTxnId would be helpful metadata here? that way on the stripe dashboard you could link a specific stripe refund to a specific txn right? This may not be viable so feel free to reject this

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
```diff
@@ -205,230 +246,492 @@ export const POST = createSmartRouteHandler({
...
+    customerId: subscription.customerId,
+    sourceTxnId,
+  });
+  const remainingStripeUnits = Math.max(0, totalStripeUnits - prior.refundedStripeUnits);
+  if (options.amountStripeUnits > remainingStripeUnits) {
+    throw new KnownErrors.SchemaError(`Refund amount cannot exceed the remaining refundable amount ($${stripeUnitsToMoneyAmount(remainingStripeUnits)}).`);
+  }
+  if (options.revokeProduct && prior.productRevoked) {
+    throw new KnownErrors.SchemaError("This subscription's product has already been revoked.");
+  }
```
nams1570: Not sure where else to put this, but is it possible to race on product revocation refunds? If two refund requests sent via sdk at same time ask for product revocation, can there be duplicate product revocation entries in bulldozer?

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
nams1570, line 1: **Please see**: bug - the way product grants work, you get the items with the original sub-start or otp txn, but you also may have items that repeat. The repeating item grants occur in a separate txn, the `item-grant-repeat` txns. You need to make sure that these are also expired if there's a product-revocation. They should have a sourceTxnId thing so that should help you.

nams1570 (changes requested):
Found some bugs

apps/backend/src/lib/payments/schema/phase-1/transactions.ts
BilalG1: agreed this is probably out of scope

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
BilalG1: yeah this is documented in in refund/route line 314

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
BilalG1: Confirmed bug — moneyAmountToStripeUnits did parseInt(value.replace('.', '')), so unpadded inputs like "10" produced 10 cents. Fixed in 165af2e95 by right-padding to currency.decimals first. Added vitest coverage.

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
BilalG1: I think it might be helpful later and not really a downside, added

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
BilalG1: Real bug for OTPs (subs already covered via subscription-end's itemQuantityChangesToExpire). Fixed in 165af2e95: OTP revoke now emits item-quantity-expire entries

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
madster456: I need to come in and test, but shouldn't we revoke product as well?

Just thinking of the case when a subscription is refunded the product is revoked, right? Depending on if its revoked at time of cancellation, or end of billing.

BilalG1: Done in ec9c16ef9 — revoke_product is gone. The new API has a single end_action: "now" | "at-period-end" | undefined

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
BilalG1: end_action is now tri-state: "now" ends immediately (frees the product line for a B-line purchase per your scenario), "at-period-end" is the previous behavior, undefined is no-op. OTPs reject "at-period-end"

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
BilalG1: we need endedAt locally so the timefold can fire sub-end (which expires outstanding item grants) and ensure-free-plan can free up the product line for a re-purchase. both would otherwise block until the stripe webhook lands.

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
BilalG1: responded above

nams1570:
@greptileai please do a full rereview

nams1570:
@greptileai can you do a full rereview

nams1570 (changes requested):
One bug - refund issues prod revocation, active-subscription-end and also sets endedAt. Bulldozer checks to see if sub has endedAt set and if it does, it creates a sub-end txn which has its own prod revocation entry. What this means is a refund that ends the subscription will also cause a sub-end txn to be created, and the two product revocations will mean the owned products of the end-user will be tallied wrong.

nams1570 (changes requested):
Currently, it seems like refund for a subscription will contain the prod-revocation but not the item-quantity-expire entries. It relies on the subscription-end event to clean up the items. This is inconsistent with how otps are refunded. Could you switch subscription refunds to walk and grab the relevant item-quantity-expire entries from the item-grant-repeat(s) and sub-start entries from the relevant subscription?

nams1570:
@greptileai do a full rereview

apps/backend/src/app/api/latest/internal/payments/transactions/refund/route.tsx
```diff
@@ -447,13 +1017,143 @@ import.meta.vitest?.describe("buildStripeRefundParams", (test) => {
...
+    });
+    // Stripe types `metadata` as `MetadataParam | "" | undefined`; we know
+    // we passed an object, so narrow before reading.
+    const metadata = params.metadata;
+    if (typeof metadata !== "object" || metadata === null) {
+      throw new Error("expected metadata object");
+    }
+    expect(metadata.refundTxnId).toBe("refund:sub-start:abc:uuid");
+  });
+});
```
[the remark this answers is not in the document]
BilalG1: currentPeriodEnd is a non-nullable column, check schema.prisma

nams1570 (approved):
Looks solid except for a slight scaling concern with the transaction walking code for item expiries. But we can deal with that in a later refactor

nams1570 (approved):
Double check the failing actions, besides that I see one thing:
1) test mode sub whose current_period is hit after we refund and end sub at period end will never regrant free plan

We can also just address this later.

PR #1436, opened 2026-05-15
[Docs] Clarify env vars and custom page URL overrides in AI setup prompts

mantrakp04:
## Summary
- Stress in the SDK setup prompt that `STACK_PROJECT_ID` (plus `STACK_SECRET_SERVER_KEY` server-side) are the **complete** env-var set — there is no separate publishable / client key, and any third slot in `.env.local` is wrong.
- Add per-framework prefix guidance (`NEXT_PUBLIC_…`, `VITE_…`) and explicitly note the server key must never be prefixed/exposed.
- In the custom-page prompt, explain that overriding one handler URL does **not** override the others: every target falls back to `urls.default`, so OAuth / magic-link / sign-out / verification flows will visibly redirect through `<projectId>.built-with-stack-auth.com` unless each is customized.
- Add a table of every handler URL target (`signIn`, `signUp`, `oauthCallback`, `signOut`, `magicLinkCallback`, `forgotPassword`, …) and the SDK call each custom page must invoke.
- Call out the Trusted Domains whitelist requirement (and the localhost-callbacks dev toggle) to avoid `REDIRECT_URL_NOT_WHITELISTED`.
- Regenerated `docs-mintlify/guides/getting-started/setup.mdx` and `home-prompt-island.jsx` from the updated prompt source.

## Test plan
- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] Spot-check the rendered Mintlify setup page reflects the new env-var + URL guidance.

## Summary by CodeRabbit

* **Documentation**
  * Clarified environment variable setup instructions for cloud projects with framework-specific prefixing requirements
  * Enhanced security guidance emphasizing server-side secret key protection
  * Improved OAuth callback configuration guidance for hosted domain redirects

[Review Change Stack](https://app.coderabbit.ai/change-stack/hexclave/stack-auth/pull/1436?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/stack-shared/src/ai/prompts.ts
```diff
@@ -553,10 +559,16 @@ export function getSdkSetupPrompt(mainType: "ai-prompt" | "nextjs" | "react" | "
...
-            STACK_PROJECT_ID=<your-project-id>  # if desired, prefix with your framework's convention for client-exposed variables (e.g. NEXT_PUBLIC_, VITE_, etc.)
+            # Prefix STACK_PROJECT_ID with your framework's convention for client-exposed
+            # variables (e.g. NEXT_PUBLIC_STACK_PROJECT_ID for Next.js, VITE_STACK_PROJECT_ID
+            # for Vite). STACK_SECRET_SERVER_KEY must NEVER be exposed to the client and
+            # must NOT be prefixed.
+            STACK_PROJECT_ID=<your-project-id>
             STACK_SECRET_SERVER_KEY=<your-secret-server-key>
             \`\`\`
 
+            These two variables are the **complete** set the SDK reads in the cloud-project setup. Do not add any additional Stack Auth env vars (in particular, there is **no** separate publishable / client key — the project ID alone is sufficient on the client). The dashboard "Project Keys" page exposes exactly these two values; if a third slot is present in any \`.env.local\` you write, it is wrong.
```
N2D4:
```suggestion
            ${isAiPrompt ? deindent`
              If the requirePublishableClientKey config option is enabled as described above, a publishable client key will also be necessary. Otherwise, these two are the only environment variables required to use Stack Auth on a server.
            ` : ""}
```

PR #1475, opened 2026-05-22
feat(hexclave): PR 1 — wire compatibility layer (invisible)

BilalG1:
## Summary

**Stacked on #1468** (`docs/hexclave-rename-plan` — the plan doc). Diff vs that base = the actual PR 1 code.

This is **PR 1 of the Hexclave rebrand: the invisible compatibility layer**. Everything is additive. Old SDKs, old wire identifiers, and old env var names keep working unchanged. The backend dual-accepts and dual-emits; new SDK code emits `x-hexclave-*` headers and the `hexclave_` Bearer prefix; cookies dual-write; env vars dual-read across every category. **No user-visible rebranding lands here** — that's PR 2.

See [`RENAME-TO-HEXCLAVE.md`](./RENAME-TO-HEXCLAVE.md) → *"PR 1 implementation guide"* for the full per-work-area spec, file pointers, and chosen approach.

## What's implemented (all 14 PR-1 work-areas)

- **SDK export aliases** — `Hexclave*` aliases for the user-facing `Stack*` exports added in `packages/template`; codegen propagates them to `@stackframe/{js,stack,react,tanstack-start}`. React-only aliases correctly excluded from `@stackframe/js`. (`e60550a2`)
- **JWT issuer dual-accept** — `decodeAccessToken` accepts both `api.stack-auth.com` and `api.hexclave.com` issuers. Signing unchanged. (`fc781def`)
- **Request-header dual-accept** — backend + dashboard proxies normalize `x-hexclave-*` → `x-stack-*` at the existing empty proxy hook (so `smart-request.tsx` and every route schema keep working unchanged); CORS allowlists extended via a derive-once helper. (`2a056eac`)
- **MCP `ask_hexclave`** — registered alongside `ask_stack_auth` via a shared helper; `ask_stack_auth` behavior byte-identical. (`30ffd604`)
- **Dev-tool** — DOM ids + header emit switched. `window.HexclaveDevTool` exposed alongside `window.StackDevTool`. (`32131ea7`)
- **The big consolidated commit** (`7fed864a`):
  - **Env vars** — central `getEnvVariable` prefix-transform (HEXCLAVE first, STACK fallback); dashboard + template client env files dual-read; `turbo.json` globalEnv; `NEXT_PUBLIC_STACK_PORT_PREFIX` renamed outright across ~82 files including docker.
  - **Cookies** — dual-write/dual-read auth (`stack-access`/`-refresh-*` and custom-domain variants), OAuth-state (`stack-oauth-{inner,outer}-*`), and low-risk cookies (`stack-is-https`, `stack-last-seen-changelog-version`). Bypass sites patched (backend OAuth callback, dashboard remote-dev auth route, impersonation snippets, snapshot serializer).
  - **Bearer prefix** — SDK token parser accepts both `stackauth_` and `hexclave_`; emits `hexclave_`. Discovery correction: this is purely SDK-internal — the backend never parses it.
  - **Response headers** — backend dual-emits `x-hexclave-{request-id,actual-status,known-error}`; SDKs dual-read (new first, stack fallback).
  - **SDK request-header emit switch** — `client/server/admin-interface.ts` + dashboard `api-headers.ts` + `internal-project-headers.ts` + `feedback-form.tsx` switched to `x-hexclave-*`. Plus `stack_response_mode` query param.
  - **Storage keys** — dev-tool / cli-auth / oauth-button / docs keys renamed (straight); `stack:session-replay:v1` dual-read so in-progress recordings survive SDK upgrades; `stack_mfa_attempt_code` dual-read.
  - **Query params** — cross-domain params dual-emit/dual-accept via shared helpers; backend `oauth/authorize` accepts `hexclave_response_mode` and `stack_response_mode`; `stack-init-id` renamed.
  - **`Symbol.for`** — app-internals symbol gets a parallel `Symbol.for("Hexclave--app-internals")` getter on each attach site (no read-site churn — old symbol still attached). 3 file-private symbols renamed outright.
  - **Config discovery** — prefer `hexclave.config.ts`, fall back to `stack.config.ts` at every discovery site (CLI / dashboard / backend / local-emulator); `init` writes the new filename; CLI credentials path migrates.
  - **Internal renames** — `StackAssertionError`, `StackClient/Server/AdminInterface` renamed outright (no alias, per the "internal-only → rename" rule). ~264 files touched.
- **Review-pass fixes** (`21217fbe`) — three real bugs found by parallel review agents and fixed:
  - `snapshot-serializer.ts` was interpolating the whole `keyedCookieNamePrefixes` array (`${arr}`) — adding a second prefix would have corrupted **every** OAuth-cookie snapshot, not just new ones.
  - **Docker port-prefix producer/consumer mismatch** — `entrypoint.sh`/`run-emulator.sh`/cloud-init `user-data` were still producing `NEXT_PUBLIC_STACK_PORT_PREFIX` while the dashboard sentinel + consumers had been renamed; silent self-host regression (custom port prefix would be ignored).
  - **Missing `hexclave-oauth-inner-*` dual-write** in the OAuth authorize route — callback's fallback masked it but the dual-write was specified by the plan.
  - Plus: `mcp.test.ts` tool-list assertions updated to include `ask_hexclave`; two dashboard header-emit sites switched to `x-hexclave-*` for consistency.
- **E2E snapshot serializer follow-up** (`4b16cc5d`) — `x-hexclave-request-id` added to the hidden-headers list (mirroring `x-stack-request-id` treatment), and 2 sample inline snapshots regenerated in `projects.test.ts` to include the new dual-emitted headers.

## Verification

- **`pnpm typecheck`** — clean (the fresh-worktree `@/.source` / Prisma codegen gap in `stack-docs` is pre-existing and unrelated).
- **`pnpm lint`** — 29/29 packages green.
- **`pnpm exec turbo run build --filter=./packages/*`** — 13/13 packages build (including `@stackframe/stack-cli` once the dashboard standalone is present).
- **Live E2E** against a running backend on `cl/hexclave-pr1`:
  - `pnpm test run apps/e2e/tests/backend/endpoints/api/v1/internal/mcp.test.ts` — **6/6 pass** (verifies the new `ask_hexclave` tool — the hand-written inline snapshot matched actual MCP server output).
  - `pnpm test run apps/e2e/tests/backend/endpoints/api/v1/internal/projects.test.ts` — **11/11 pass** (verifies wire dual-accept + dual-emit end-to-end; the snapshot serializer fix was found and applied during this check).

A four-agent parallel **review pass** also audited the full diff for logic/runtime bugs across the work-areas (wire headers + JWT, cookies + bearer + symbols, env vars, query params + config + MCP + aliases). All in-slice review verdicts were ✓ except the three bugs listed above, which are now fixed.

## Known follow-ups (out of scope for this PR)

- **E2E snapshots across the rest of the suite** — backend now dual-emits `x-hexclave-{known-error,actual-status}` alongside `x-stack-*`, which legitimately appears in inline snapshots throughout `apps/e2e`. Two were regenerated here as a sample; the rest should regen with `vitest -u` in CI.
- **Docker shell env vars beyond `PORT_PREFIX`** — `entrypoint.sh` still reads `STACK_*` env vars directly (the JS-side `getEnvVariable` transform doesn't help the shell). JS consumers dual-read so it works in practice; full shell-level dual-read is a deeper self-host follow-up.
- **`@stackframe/stack-cli` build ordering** — pre-existing; needs `build:rde-standalone` first. Not affected by this PR.

## Test plan

- [ ] CI runs full e2e suite (with `vitest -u` to absorb dual-emit snapshot deltas, then committed back)
- [ ] Spot-check: an old SDK build (emitting only `x-stack-*`) still authenticates against the new backend
- [ ] Spot-check: a new SDK (emitting `x-hexclave-*` / `Bearer hexclave_*`) still authenticates against an old backend during deploy ordering
- [ ] Manual: `npx @stackframe/stack-cli@latest init` (new onboarding entrypoint) generates `hexclave.config.ts`
- [ ] Manual: existing `stack.config.ts`-only project still resolves (no migration required)


apps/mcp/src/mcp-handler.ts
```diff
@@ -104,72 +104,79 @@ export function createStackMcpHandler(config: { streamableHttpEndpoint: string }
...
+            const responseConversationId = body.conversationId ?? conversationId ?? "";
+
+            return {
+              content: [{ type: "text", text: `${text.length > 0 ? text : "(empty response)"}\n\n[conversationId: ${responseConversationId} - pass this value as the conversationId parameter in your next ${toolName} call to continue this conversation]` }],
+            };
+          },
+        );
+      };
+      registerAskTool("ask_stack_auth", "Stack Auth");
+      registerAskTool("ask_hexclave", "Hexclave");
```
N2D4: we don't need backwards compat for the mcp

docker/local-emulator/qemu/run-emulator.sh
```diff
@@ -10,7 +10,9 @@ RUN_DIR="${EMULATOR_RUN_DIR:-$HOME/.stack/emulator/run}"
 
 VM_RAM="${EMULATOR_RAM:-4096}"
 VM_CPUS="${EMULATOR_CPUS:-4}"
-PORT_PREFIX="${PORT_PREFIX:-${NEXT_PUBLIC_STACK_PORT_PREFIX:-81}}"
+# Hexclave rebrand: port-prefix var renamed outright to NEXT_PUBLIC_HEXCLAVE_PORT_PREFIX;
+# accept the legacy NEXT_PUBLIC_STACK_PORT_PREFIX as a fallback input.
+PORT_PREFIX="${PORT_PREFIX:-${NEXT_PUBLIC_HEXCLAVE_PORT_PREFIX:-${NEXT_PUBLIC_STACK_PORT_PREFIX:-81}}}"
```
N2D4: we do this in a bunch of places, we shouldn't, it's really ugly. let's just get rid of NEXT_PUBLIC_STACK_PORT_PREFIX

docs/.env.development
N2D4, line 1: no action needed on this pr but in future prs let's keep the docs (old ones, not mintlify) as-is, we will be deprecating them alongside the name change

packages/template/src/components-page/stack-handler-client.tsx
```diff
@@ -264,7 +264,9 @@ export function StackHandlerClient(props: BaseHandlerProps & Partial<RouteProps>
 
   const redirectIfNotHandler = (name: keyof HandlerUrls) => {
     const url = stackApp.urls[name];
-    const isCrossDomainLocalOauthCallback = name === "oauthCallback" && searchParams.stack_cross_domain_auth === "1";
+    // Hexclave rebrand: accept the cross-domain auth marker under either name.
+    const isCrossDomainLocalOauthCallback = name === "oauthCallback"
+      && (searchParams.hexclave_cross_domain_auth === "1" || searchParams.stack_cross_domain_auth === "1");
```
N2D4: for cross domain params we don't need to keep the old ones around, they are shortlived so if you update the sdk it'll be alright

packages/template/src/lib/stack-app/apps/implementations/admin-app-impl.ts
```diff
@@ -591,6 +591,11 @@ export class _StackAdminAppImplIncomplete<HasTokenStore extends boolean, Project
     };
   }
 
+  // Hexclave rebrand: expose the same internals under the parallel symbol.
+  get [hexclaveAppInternalsSymbol]() {
+    return this[stackAppInternalsSymbol];
+  }
+
```
N2D4: not necessary, let's just keep using the old symbol (and rename the symbol variable, while keeping the same Symbol.for key)

packages/template/src/lib/stack-app/apps/implementations/client-app-impl.ts
```diff
@@ -727,21 +741,48 @@ export class _StackClientAppImplIncomplete<HasTokenStore extends boolean, Projec
   protected get _refreshTokenCookieName() {
     return `stack-refresh-${this.projectId}`;
   }
-  private _getRefreshTokenDefaultCookieNameForSecure(secure: boolean): string {
-    return `${secure ? "__Host-" : ""}${this._refreshTokenCookieName}--default`;
+  // Hexclave rebrand: the new base name for refresh-token cookies. Cookies are dual-written under both
+  // `stack-refresh-<projectId>` (via `_refreshTokenCookieName`) and `hexclave-refresh-<projectId>`.
+  protected get _hexclaveRefreshTokenCookieName() {
+    return `hexclave-refresh-${this.projectId}`;
```
N2D4: no need to dual-write the cookies

N2D4: we just need to read from both cookies

packages/template/src/lib/stack-app/apps/implementations/redirect-page-urls.ts
```diff
@@ -1,22 +1,53 @@
...
+// Hexclave rebrand: cross-domain auth query params travel SDK↔SDK across an auth
+// handoff. A handoff between two SDK majors (one per domain) must still resolve, so
+// we dual-emit both the legacy `stack_*` and the new `hexclave_*` param sets when
+// writing, and accept either when reading (preferring the new name).
 export const crossDomainAuthQueryParams = {
+  marker: "hexclave_cross_domain_auth",
+  state: "hexclave_cross_domain_state",
+  codeChallenge: "hexclave_cross_domain_code_challenge",
+  afterCallbackRedirectUrl: "hexclave_cross_domain_after_callback_redirect_url",
+} as const;
```
N2D4: we don't need to keep compat for old crossdomain auth query param names

packages/template/src/lib/stack-app/index.ts
```diff
@@ -11,6 +11,21 @@ export type {
...
+} from "./apps";
+export type {
+  StackAdminAppConstructor as HexclaveAdminAppConstructor,
+  StackAdminAppConstructorOptions as HexclaveAdminAppConstructorOptions,
+  StackClientAppConstructor as HexclaveClientAppConstructor,
+  StackClientAppConstructorOptions as HexclaveClientAppConstructorOptions,
+  StackClientAppJson as HexclaveClientAppJson,
+  StackServerAppConstructor as HexclaveServerAppConstructor,
+  StackServerAppConstructorOptions as HexclaveServerAppConstructorOptions
+} from "./apps";
```
N2D4: can we flip this? the actual name should be HexclaveAdminApp etc., and then there should be a (deprecated) alias called StackAdminApp

N2D4 (approved):

PR #1521, opened 2026-05-29
feat(cli): auto-update RDE dashboard via npx re-exec on `stack dev`

BilalG1:
## What & why

Re-running `stack dev` / `hexclave dev` now picks up the **latest published dashboard without reinstalling the CLI**.

In the RDE, the dashboard is a Next.js standalone build **bundled into the `@hexclave/cli` npm tarball** — so a dashboard change only reaches a user when they get a newer CLI *version*. This PR closes that gap for the recommended `stack dev` flow.

## How it works

1. **npx self-re-exec** — at the top of the `dev` action, the CLI checks npm for a newer `@hexclave/cli`. If found, it re-execs `npx --yes -p @hexclave/cli@<latest> stack dev <your args>` (with a loop guard) and exits with the child's code. The running code — and the dashboard bundled in that tarball — is now the latest; the user's installed devDependency is untouched. npx caches per version, so steady-state runs are fast.

2. **Dashboard version handshake** (the necessary second half) — `stack dev` keeps a **detached background dashboard** alive across runs and reuses it by default, which would otherwise silently defeat the update. The now-latest process compares the running dashboard's version (persisted in dev-env state) against its own and **kills + restarts** the stale one (SIGTERM → wait → SIGKILL) so the new dashboard actually binds `:26700`. Equal/older/unknown versions are reused exactly as before.

## Safety / opt-outs

- Skipped for the re-exec'd child (`STACK_CLI_SKIP_AUTO_UPDATE`, loop guard), when the user opts out (`STACK_CLI_NO_AUTO_UPDATE` / `--no-auto-update`), and in CI (`CI`).
- Registry lookup is TTL-cached in dev-env state with a short timeout and is **offline-safe** — any failure (no network, no npx) falls through to the installed CLI.
- `isVersionNewer` never downgrades and returns false for unparseable versions.

## Changes

- **`packages/stack-cli/src/lib/self-update.ts`** (new) — `maybeReexecToLatest()`, `resolveLatestVersion()`, `isVersionNewer()`, `buildNpxInvocation()`.
- **`packages/stack-cli/src/commands/dev.ts`** — re-exec wiring, `killLocalDashboard()`, version handshake, `--no-auto-update` flag, version stamp on the recorded dashboard process.
- **`packages/stack-cli/src/lib/dev-env-state.ts`** — `localDashboard.version` + `cliUpdateCheck` cache helpers.
- Tests: new `self-update.test.ts` + additions to `dev-env-state.test.ts`.

## Verification

- `pnpm --filter @hexclave/cli run lint` ✅
- `pnpm --filter @hexclave/cli run typecheck` ✅
- `pnpm --filter @hexclave/cli run test` ✅ (132 passed)

## Prerequisite

Relies on `@hexclave/cli` being published to npm with the `latest` dist-tag tracking releases — otherwise the check is a no-op (which is safe).

---
## Summary by cubic
`hexclave dev` now re-execs via `npx` to run the latest `@hexclave/cli`, so the bundled RDE dashboard stays current without reinstalling. It reuses the running dashboard and only restarts it when the current CLI is strictly newer.

- **New Features**
  - Auto-update: always re-execs `npx --yes --min-release-age=0 -p @hexclave/cli@latest hexclave dev ...`; runs in CI; opt out with `--no-auto-update` or `STACK_CLI_NO_AUTO_UPDATE=1`.
  - Per-port dashboard version handshake: records the CLI version per port and restarts only when strictly newer; otherwise reuses it (respects `NEXT_PUBLIC_HEXCLAVE_LOCAL_DASHBOARD_PORT`).

- **Bug Fixes**
  - Safer restarts: after SIGTERM, wait for the port to free instead of pid probes; bail on ESRCH/EPERM; only SIGKILL if the port still answers.
  - Robust execution: ship a single `hexclave` bin (fixes `pnpx`/`pnpm dlx`), forward SIGINT/SIGTERM to children, validate per-port dashboard state, update help/messages to `hexclave`, and make Windows re-exec reliable (`npx.cmd` with shell and argv quoting).

<sup>Written for commit 80c9b30a5c701e7a34da723aafc4326150f91250. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hexclave/hexclave/pull/1521?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * CLI can auto-check and re-exec to a pinned newer release (opt-out: --no-auto-update).
  * Local dashboard startup is version-aware and only restarts when the CLI is strictly newer.
  * Improved child-process signal forwarding for cleaner shutdowns.

* **Tests**
  * Expanded unit tests covering dev workflow, self-update, package metadata, persistence, and dashboard lifecycle.

* **Bug Fixes**
  * Updated user-facing CLI messaging to use "hexclave" command names.

* **Chores**
  * Removed legacy docs workspace entry.


BilalG1:
## Review round (commit 2)

Spawned three review passes (bug-hunt, simplify/reuse, test-coverage) and applied the findings.

**Correctness**
- **H1 (real bug):** `killLocalDashboard` returned as soon as `/health` stopped answering, but the listening socket isn't released until the process actually exits — so the replacement dashboard could hit `EADDRINUSE` and hang for 60s. Now waits for the process to be gone (`processExists`) **and** the port unreachable, with a bounded force-kill phase.
- **M2 (pid reuse):** bail out without waiting/SIGKILL when the recorded pid is gone (`ESRCH`) or owned by another process (`EPERM`), avoiding both a hang and signalling an unrelated process.

**Quality / dedup**
- Extracted `lib/own-package.ts` — single source for reading the CLI's own `package.json` + bin resolution. Was duplicated **3×** (`index.ts`, `sentry.ts`, `self-update.ts`).
- Extracted `lib/child-process.ts` `forwardSignals` — the SIGINT/SIGTERM forwarding boilerplate, now shared by `dev.ts`, `emulator.ts`, `self-update.ts` (also 3 copies).
- Extracted pure, testable decision functions: `decideReexec()` and `shouldRestartDashboard()`/`processExists()`.
- Fixed a stale comment claiming jsdom is the test env (it's `node`).

**Tests:** +35 cases (132 → 167). `decideReexec` branches, `resolveBinName`/`parseOwnPackage`, `isVersionNewer` edges (v-prefix, `x.y`, both-prerelease, large nums), `resolveLatestVersion` (TTL boundary, malformed/non-string body, fetch-URL contract, `npm_config_registry`), `buildNpxInvocation` (Windows, spaced/dashed args), `shouldRestartDashboard` truth table, `killLocalDashboard` early-returns, dev-env-state back-compat + no-clobber.

**Known residual (not fixed, by design):** concurrent `stack dev` runs on the same machine share an unlocked state file and could both decide to restart (M3). Pre-existing to this feature; would need a file lock. Flagging rather than expanding scope.

Lint ✅ · typecheck ✅ · 167 tests ✅

packages/stack-cli/src/lib/self-update.ts
```diff
@@ -0,0 +1,274 @@
...
+  return normalized !== "" && normalized !== "0" && normalized !== "false";
+}
+
+// Auto-update is skipped when we're the re-exec'd child, when the user opted
+// out, or in CI (where re-running an arbitrary newer version would be
+// non-deterministic).
+export function shouldAutoUpdate(env: NodeJS.ProcessEnv): boolean {
+  if (isEnvFlagEnabled(env[SKIP_AUTO_UPDATE_ENV])) return false;
+  if (isEnvFlagEnabled(env[DISABLE_AUTO_UPDATE_ENV])) return false;
+  if (isEnvFlagEnabled(env.CI)) return false;
```
N2D4: should we really skip CI? this will cause the behavior between CI and real environments to diverge. i think we should still auto-update even in the CI

packages/cli/src/lib/self-update.ts
```diff
@@ -0,0 +1,274 @@
...
+    command,
+    shell: isWindows,
+    args: [
+      "--yes",
+      // Override any global npm "cooldown" for this call only — we always want
+      // the just-published latest, and npx of a pinned version newer than the
+      // cooldown window otherwise fails with ETARGET (which would kill
+      // `hexclave dev`). npm's config is `min-release-age` (days, npm >=11.10.0);
+      // older npm silently ignores the unknown flag.
+      "--min-release-age=0",
```
N2D4, line 61: let's have a min-release age of 1 week here, but exclude `@stackframe/*` specifically; this still protects people from supply chain attacks, while keeping our packages up-to-date

packages/stack-cli/src/lib/self-update.ts
```diff
@@ -0,0 +1,274 @@
...
+  const b = parseVersionCore(current);
+  if (a == null || b == null) return false;
+  for (let i = 0; i < 3; i++) {
+    if (a.core[i] !== b.core[i]) {
+      return a.core[i] > b.core[i];
+    }
+  }
+  // Same x.y.z: a final release outranks a prerelease of the same core.
+  return !a.hasPrerelease && b.hasPrerelease;
+}
```
N2D4: why do we do these complicated version checks instead of just always running the latest version with `npx @hexclave/cli@latest`?`and then to prevent recursion and also in the cli script in package.json we just set the environment variable which disables the behavior and instead runs the package directly. that logic should be much simpler

N2D4: that way the logic would be very simple. "if envvar is not set, run latest version with npx and envvar set. otherwise, run actual cli"

N2D4 (approved):

N2D4 (changes requested):
some comments

packages/cli/src/lib/self-update.ts
BilalG1: npm has no per-package exclude for min-release-age

pnpm-workspace.yaml
```diff
@@ -2,7 +2,9 @@ packages:
   - packages/*
   - apps/*
   - examples/*
-  - docs
+  # `docs/` is the legacy fumadocs site, no longer maintained — replaced by
+  # `docs-mintlify/`. Kept on disk for migration reference; excluded from the
+  # workspace so it doesn't gate install / typecheck / lint.
```
N2D4: let's remove it only after we finish the migration where we redirect from old docs to new docs

N2D4 (approved):

PR #1530, opened 2026-06-01
fix(dashboard): disable real Stripe setup in RDE onboarding

aadesh18:
The onboarding wizard's `payments_setup` step was reaching real Stripe in **remote development environment** mode via both paths:

- **Connect** (secondary action) calls `setupPayments()`, which creates a real Stripe Connect account on the backend.
- **Do Later** (primary action) *also* silently calls `setupPayments()` for US-based projects — meant to pre-create the account so it's ready later.

Both are inappropriate for RDE — the project is a throwaway dev project and should never own a real Stripe Connect account.

## Screenshot

Onboarding payments-setup step in RDE — Connect disabled, amber RDE notice rendered

What to look at:
- **Connect** button (right) renders but is **disabled** — the DOM `disabled` attribute is `true`.
- Amber notice under "Powered by Stripe": *"Payments setup is not available in remote development environments."*
- **Do Later** (left) stays enabled — clicking it advances onboarding to "welcome" without invoking the implicit `setupPayments()` call.

Before this PR, both buttons would have been clickable in RDE and clicking either would have created a real Stripe Connect account against the throwaway dev project.

## What changed

`apps/dashboard/src/app/(main)/(protected)/(outside-dashboard)/new-project/page-client-parts/project-onboarding-wizard.tsx`:

- **Connect button**: disabled when `isRemoteDevelopmentEnvironment` is true.
- **Card body**: amber notice *"Payments setup is not available in remote development environments."* replaces the country-availability message.
- **Do Later handler** (`deferPaymentsSetup`): skips `setupPayments()` entirely when in RDE — onboarding still advances to "welcome", just without the implicit Stripe-account-creation side effect.

This mirrors the gating pattern already used by:
- `apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/payments/payouts/page-client.tsx` (gates `<ConnectPayouts />` on `project.isDevelopmentEnvironment`),
- `apps/dashboard/src/components/payments/stripe-connect-provider.tsx` (returns a null Connect instance for dev projects).

## Scope

- **RDE only** — local-emulator mode is intentionally untouched (its dashboard flow has different constraints; can be folded in later if needed).
- **Frontend only.** Backend `setupPayments` / `account-session` / `account-info` routes still execute against real Stripe for dev-env projects — deferred to a follow-up PR for defense-in-depth.

## Tests

Three new cases in `project-onboarding-wizard.test.tsx` — each flips the `NEXT_PUBLIC_STACK_IS_REMOTE_DEVELOPMENT_ENVIRONMENT` env mock to "true" and asserts:

1. The **Connect** button has the `disabled` attribute.
2. The text *"Payments setup is not available in remote development environments."* renders.
3. Clicking **Do Later** advances to "welcome" but `setupPayments` is **not** called — this is the security-relevant assertion that prevents the original bug.

The existing 8 tests still pass (the env mock is now a hoisted `vi.fn` reset to `"false"` in `afterEach`, so default behavior is unchanged).

## Summary by CodeRabbit

* **Bug Fixes**
  * Payments setup is blocked in remote development environments to prevent unintended calls.

* **User Experience**
  * A notice explains payments setup is unavailable in remote development environments.
  * The "Connect" button is disabled in remote development mode.
  * Deferring payments still advances onboarding but will not trigger payments setup in remote development; a dedicated amber warning is shown.

* **Tests**
  * Added tests covering disabled Connect, the notice, and deferred/do-later behavior.


N2D4 (changes requested):
we shouldn't have an entire step just to show the user an error message. we should just not show that onboarding step

PR #1537, opened 2026-06-02
Replace writeConfigObject with AI-aware updateConfigObject

mantrakp04:
## Summary

Replaces `writeConfigObject` (destructive overwrite) with `updateConfigObject` — an async, AI-aware updater that preserves user-authored config structure (imports, external file references, helpers).

**Dual-path approach:**
- **Fast path** (deterministic, no AI): plain static literal configs → `override()` + in-memory validation + atomic write
- **Agent path** (custom structure): configs with `import x from "./file.txt" with { type: "text" }` etc. → Claude agent edits the external files in place, then validates

**Safety guarantees:**
- Snapshot/restore: config + all relative imports are captured before the agent runs; rolled back on any failure
- In-memory validation on fast path (never write unvalidated bytes)
- Semantic check when config is evaluable; no-op detection + structural check when it isn't
- Path traversal guard on imports (rejects `../` escapes)
- Agent isolation: `settingSources: []`, `strictMcpConfig: true`, `CLAUDE_CODE_DISABLE_AUTO_MEMORY`, no Bash tool
- `scheduleSync` only fires after a successful update
- Bounded 120s timeout on agent runs (configurable via env var)

CI failures are preexisting on `dev` (`ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` from overrides move without lockfile regen); this branch has zero lockfile changes vs dev.

Link to Devin session: https://app.devin.ai/sessions/cc7409a357bc472ea19fbed065f1229f
Requested by: @mantrakp04

## Summary by CodeRabbit

* **New Features**
  * Introduced partial configuration update functionality with validation and automatic rollback on failures.
  * Enhanced configuration management with support for more complex file structures and external references.

* **Chores**
  * Added Claude Agent SDK dependency for configuration update operations.

## Documentation

Docs for this feature were added in this branch:

- **New page** `docs-mintlify/guides/going-further/local-development.mdx` — covers `stack dev`, the development-environment flow, and how dashboard edits are written back to the local config file (structure-preserving fast path vs. assistant path, external `import … with { type: "text" }` templates, validation + rollback). Added to `docs.json` nav; also fixes the previously-broken `/guides/going-further/local-development` links from `index.mdx` and `self-host.mdx`.
- **`docs-mintlify/guides/going-further/cli.mdx`** — added a `stack dev` ("Run a development environment") section.
- **Skill-site AI prompts** — filled in the `config-docs` and `dashboard-instructions` placeholders under `packages/stack-shared/src/ai/unified-prompts/skill-site-prompt-parts/`, and added a structure-preserving note to the setup prompt.
- **`CHANGELOG.md`** — user-facing entry.


mantrakp04:
@greptile-ai review

mantrakp04:
@greptile-ai review

mantrakp04:
@greptile-ai review

N2D4 (changes requested):
why is there a fast path and an agent path? i generally don't like fallbacks, so i think we should avoid doing that. i think we can always use the agent path.  
  
generally also the PR turned out larger than i expected. what do you think the reason is? is there a way to simplify it you think? trying to update my priors here  
  
do we need an extra package for the local-config-updater? can we have it in stack-shared instead?

mantrakp04:
> why is there a fast path and an agent path? i generally don't like fallbacks, so i think we should avoid doing that. i think we can always use the agent path.
> 
> generally also the PR turned out larger than i expected. what do you think the reason is? is there a way to simplify it you think? trying to update my priors here
> 
> do we need an extra package for the local-config-updater? can we have it in stack-shared instead?

there is no fallback here
it uses logic/code to write the file in 2 cases:
- if --overwrite tag is passed, this one doesn't use the ai
- creating a new config file for the first time
only the updates are passed thru the ai

and the reason for local-config-updater and not stack shared is cuz claude code was getting bundled with the sdks

packages/shared-backend/package.json
N2D4, line 1: does this need to be its own package? can't we have it in stack-shared? feels superfluous to add an extra package just for this

N2D4 (changes requested):

packages/shared-backend/package.json
N2D4, line 1: rename this as discussed

N2D4 (approved):

PR #1584, opened 2026-06-11
fix: handle ENOENT race and misleading progress log in CLI dashboard restart

Developing-Gamer:
## Summary

Fixes a crash during CLI version-upgrade restarts (`1.0.10 → 1.0.11`) where `replaceDashboardRuntimeSentinels` throws ENOENT because the dying old dashboard process can delete files between `readdirSync` and `readFileSync`.

Three targeted fixes:

1. **ENOENT guard** in `replaceDashboardRuntimeSentinels` — catches and skips files/dirs that vanish (TOCTOU race), using a runtime-checked `hasErrnoCode` helper instead of unsafe `as` casts
2. **Grace period** in `killLocalDashboard` — after port is released, waits 500ms if the pid is still alive so its final I/O settles before the new runtime dir is created
3. **Progress log** — `stop(showDone?: boolean)` replaces the misleading `stop(finalMessage?: string)` API where the string was silently discarded

Link to Devin session: https://app.devin.ai/sessions/456fbdf12ebc477f84fd2bebb40c4aca
Requested by: @Developing-Gamer

## Release Notes

* **Bug Fixes**
  * Updated development dashboard progress output so the “... done!” message appears only when the operation truly completes successfully.
  * Hardened bundled dashboard runtime sentinel replacement to better tolerate restart/upgrade race conditions during directory traversal and file updates.
  * Improved local dashboard shutdown by allowing an extra brief wait for pending I/O when the original process is still present after the port becomes unreachable.
  * Refined dashboard startup reachability handling for more consistent completion behavior.

### DIFF
--- a/packages/cli/src/commands/dev.ts
+++ b/packages/cli/src/commands/dev.ts
@@ -240,7 +240,14 @@ function hasErrnoCode(error: unknown, code: string): boolean {
 }
 
 function replaceDashboardRuntimeSentinels(root: string, env: NodeJS.ProcessEnv): void {
-  for (const entry of readdirSync(root, { withFileTypes: true })) {
+  let entries: ReturnType<typeof readdirSync>;
+  try {
+    entries = readdirSync(root, { withFileTypes: true });
+  } catch (error) {
+    if (hasErrnoCode(error, "ENOENT")) return;
+    throw error;
+  }
+  for (const entry of entries) {
     const path = join(root, entry.name);
     if (entry.isDirectory()) {
       try {
--- a/packages/cli/src/commands/dev.ts
+++ b/packages/cli/src/commands/dev.ts
@@ -235,6 +235,10 @@ function replaceSentinels(content: string, env: NodeJS.ProcessEnv): string {
   });
 }
 
+function hasErrnoCode(error: unknown, code: string): boolean {
+  return error instanceof Error && "code" in error && error.code === code;
+}
+
 function replaceDashboardRuntimeSentinels(root: string, env: NodeJS.ProcessEnv): void {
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/dev.ts
```diff
@@ -235,14 +239,29 @@ function replaceDashboardRuntimeSentinels(root: string, env: NodeJS.ProcessEnv):
   for (const entry of readdirSync(root, { withFileTypes: true })) {
     const path = join(root, entry.name);
     if (entry.isDirectory()) {
-      replaceDashboardRuntimeSentinels(path, env);
+      try {
+        replaceDashboardRuntimeSentinels(path, env);
+      } catch (error) {
+        if ((error as NodeJS.ErrnoException).code === "ENOENT") continue;
```
N2D4: this is not a safe cast to mak

packages/cli/src/commands/dev.ts
```diff
@@ -235,14 +239,29 @@ function replaceDashboardRuntimeSentinels(root: string, env: NodeJS.ProcessEnv):
...
-    const buffer = readFileSync(path);
+    // During a version-upgrade restart, the old dashboard process may still be
+    // shutting down and can delete or move files via cached absolute paths. Guard
+    // against the resulting TOCTOU race (readdirSync sees the entry, but the
+    // file is gone by the time we read it).
+    let buffer: Buffer;
+    try {
+      buffer = readFileSync(path);
+    } catch (error) {
+      if ((error as NodeJS.ErrnoException).code === "ENOENT") continue;
```
N2D4: as above

N2D4 (approved):

Developing-Gamer:
bruh

PR #1587, opened 2026-06-11
docs: teach agents to sign out via user.signOut(), not app.urls.signOut links

BilalG1:
## Problem

Dogfooding caught this: an agent-built app (hdf-12) following skill.hexclave.com ended up with broken sign-out. It rendered `<a href={app.urls.signOut}>` — the exact pattern our own Supabase setup example taught — while using the recommended hosted auth pages (`urls: { default: { type: "hosted" } }`).

With hosted pages, `app.urls.signOut` points to the hosted domain's `/handler/sign-out`. A raw link navigation skips the return-URL and session-handoff params that the SDK's redirect helpers add at click time, so:

- the hosted page revokes the session server-side but **cannot clear the app-domain cookies**; the cached access token is validated statelessly, so the user appears to stay signed in until it expires, and
- with no `after_auth_return_to`, the user is **stranded on the hosted domain** (hosted `afterSignOut` is `/handler/sign-in`).

The SDK already special-cases this — `redirectToSignOut()` signs out locally in hosted mode — but the setup docs never said how to sign out, and the only example showed the raw-link pattern.

## Changes

- **ai-setup-prompt.ts**: add a short "Signing users in and out" step (use `useUser({ or: "redirect" })`, `app.redirectToSignIn()`/`app.redirectToSignUp()`, `await user.signOut()`; never render `app.urls.signIn`/`app.urls.signOut` as plain hrefs) and fix the Supabase example to use the helpers.
- **reminders.ts**: one matching agent-reminder bullet.
- Fix the hand-written mirrors: `docs-mintlify/guides/integrations/supabase/overview.mdx` and the SaaS tutorial's sign-out link.
- Regenerated `setup.mdx`, `llms-full.txt`, and prompt snippets.

Kept intentionally tight (+60/−16 incl. regenerated files) since the setup step is emitted once per framework variant.

## Follow-up (not in this PR)

SDK/platform-level hardening so raw links to the hosted sign-out page degrade gracefully (cross-domain sign-out handoff / trusted-domain fallback when `after_auth_return_to` is missing).

---
## Summary by cubic
Teach users and agents to sign out with `user.signOut()` and use redirect helpers for auth flows. Deprecates `app.urls.signOut` to prevent broken sign-out with hosted auth pages.

- **Bug Fixes**
  - Added a “Signing users in and out” step to the SDK setup: `useUser({ or: "redirect" })`, `app.redirectToSignIn()`/`app.redirectToSignUp()`, `await user.signOut()`; warns against raw `app.urls.signIn`/`signOut` links.
  - Updated examples (Supabase integration and SaaS tutorial) to use buttons calling helpers; removed `next/link` sign-in/out links.
  - Added an agent reminder and refreshed generated docs/prompts (`setup.mdx`, `llms*.txt`, prompt snippets).
  - Marked `app.urls.signOut` as deprecated in SDK types; treated as non-navigable in `ResolvedHandlerUrls`. Use `await user.signOut()`.

<sup>Written for commit e1cf9892805b6c0b746f790018f7f278a889619f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hexclave/hexclave/pull/1587?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Updated auth setup guidance across framework and integration guides to use `app.redirectToSignIn()` / `app.redirectToSignUp()` for sign-in/sign-up and `await user.signOut()` for sign-out, with a warning against rendering raw sign-in/sign-out URLs as plain links.
  * Refreshed Supabase and other examples to use button controls (no `Link` navigation) for auth actions.
  * Updated generated prompts and reminder content to match the new frontend auth flow.
* **Deprecations**
  * Marked direct sign-out URL usage as deprecated, steering implementations toward `user.signOut()`.


N2D4 (changes requested):
how about we just remove `app.urls.signOut`? i don't think it makes much sense actually because it's not really a URL that can be redirected to, so having it in the first place and then telling AI "don't use it" seems a bit superfluous. technically a breaking change — maybe we can keep it and deprecate it (now that the docs don't mention it anymore it should prolly be fine)

N2D4 (approved):

N2D4 (changes requested):
(see above)

PR #1591, opened 2026-06-12
Support OAuth provider config in development environments

mantrakp04:
## Summary
- Development environments only expose one writable config surface (the branch-level `hexclave.config.ts`). Config updates are now accepted iff they can be expressed as branch config (via a new `updateDevelopmentEnvironmentConfig`) and rejected otherwise — instead of blanket-rejecting non-`pushable` updates. `pushable` is ignored in dev.
- Shared OAuth providers are written as pure branch-level config (`isShared` defaults to `true`, all credential fields left unset) so they stay editable in development environments.
- UI surfaces clear messaging and disabled states for providers that require environment-level custom credentials (read-only in dev): an info alert in the settings form, a footnote in the pill toggle, and a disabled Save button.
- Centralizes the `isRemoteDevelopmentEnvironment()` helper in `lib/env.tsx` (removing duplicate inline definitions in `dashboard-user.ts` and `config-update.tsx`).
- Extracts a shared `useProviderConfigActions` hook and a `SHARED_KEYS_TOOLTIP` constant to remove duplication on the auth-methods page.
- Fixes lingering "Stack"-branded copy to "Hexclave".

## Test plan
- [ ] In a development environment, enabling/disabling a **shared** OAuth provider persists to `hexclave.config.ts`.
- [ ] In a development environment, a provider requiring custom credentials shows the "Not available in development environments" alert and a disabled Save button.
- [ ] In production, custom and shared OAuth provider configuration continues to work as before (pushable + non-pushable paths).
- [ ] `pnpm lint` and `pnpm typecheck` pass.

---
## Summary by cubic
Make OAuth provider config schema-driven across layers: branch owns the provider roster and on/off state; environment stores credentials only. This enables toggles in development, fixes onboarding, preserves custom OIDC, and pulls in the latest `dev` (docs regenerated; no functional changes).

- **New Features**
  - Split defined by `OAUTH_PROVIDER_BRANCH_ENABLE_FIELDS` in `@hexclave/shared`; branch schema validates enable fields, environment schema validates credentials.
  - Environment writes are normalized by `migrateConfigOverride("environment", ...)`: whole-provider objects become credential leaf keys and branch-owned fields are dropped, so env cannot clobber the branch roster.
  - Dashboard saves span both layers via new `provider-config` helpers; in dev, env creds are read-only, the toggle hides “Own credentials,” and Save is disabled when custom creds are required.
  - Onboarding writes shared providers to the branch only.

- **Migration**
  - One-time migration moves provider enable fields from environment overrides to branch config, keeps credentials in env, and preserves unrelated config.
  - Handles nested/leaf/object shapes, clears stale branch null markers, and is idempotent.

<sup>Written for commit 0fdd97c63c11895693248beeb0acd5d59a694e2f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hexclave/hexclave/pull/1591?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * OAuth provider credential settings now respect environment writability, showing read-only/pinned or production-only guidance when applicable.
* **Bug Fixes**
  * OAuth provider add/edit/remove actions now apply consistent branch-vs-environment updates, preventing missing credentials when enablement changes.
  * Environment-level override writes that include provider roster/enable fields are rejected.
* **Maintenance**
  * Existing OAuth provider configuration is migrated to the two-layer split to preserve enabled states and credentials.
* **Tests**
  * Added/updated migration and OAuth endpoint coverage to validate idempotency and correct rendering behavior.


apps/dashboard/src/lib/config-update.tsx
```diff
@@ -489,6 +489,38 @@ function GithubPushBody({
...
+): Promise<boolean> {
+  // Imported lazily so the schema module stays out of the client bundle of the
+  // hosted dashboard, which never runs this path.
+  const { branchConfigSchema, getConfigOverrideErrors } = await import("@hexclave/shared/dist/config/schema");
+  const branchValidity = await getConfigOverrideErrors(branchConfigSchema, configUpdate);
+  if (branchValidity.status === "error") {
+    // Same contract as the read-only case in useUpdateConfig below: blocking alert, handled failure.
+    alert("These settings are read-only in a development environment. Update them in your production deployment instead.");
+    return false;
+  }
```
N2D4: you are ignoring all errors here and always assuming that it is a read-only error. why do we make this change here?

apps/dashboard/src/lib/config-update.tsx
```diff
@@ -548,6 +580,10 @@ export type UpdateConfigOptions = {
  *
  * For non-pushable configs, updates the environment config directly.
  *
+ * In a development environment, `pushable` is ignored: the update is written to
+ * the local config file if it's expressible as branch config, and rejected
+ * otherwise (see `updateDevelopmentEnvironmentConfig`).
+ *
```
N2D4: i don't think we should do this. if you're trying to make non-pushable changes, it should reject those in the development environment (!pushable = env config override, which doesn't exist in RDE)

N2D4 (changes requested):
this is a very hacky solution to the problem. instead, we should do a proper implementation by separating the oauth provider updates into two different updates: first, set the OAuth provider to be enabled (pushable/branch config), then set the custom oauth keys (non-pushable/environment config)

nams1570:
@greptileai this is a complete rework of this PR. Can you do a deep rereview? goal is to make sure that you can add new providers if theyre shared keys on RDE. We accomplish this by having branch config be where provider enable/disable info is, and having env config be where secrets live. Also some UI changes.

nams1570:
@greptileai this is a complete rework of this PR. Can you do a deep rereview? goal is to make sure that you can add new providers if theyre shared keys on RDE. We accomplish this by having branch config be where provider enable/disable info is, and having env config be where secrets live. Also some UI changes.
Made some migration changes so check for backwards compatibility and risk of downtime

nams1570:
@greptileai this is a complete rework of this PR. Can you do a deep rereview? goal is to make sure that you can add new providers if theyre shared keys on RDE. We accomplish this by having branch config be where provider enable/disable info is, and having env config be where secrets live. Also some UI changes.
Made some migration changes so check for backwards compatibility and risk of downtime

nams1570:
@greptileai do a rereview? changed the ordering on the two phase approach to applying provider changes. Also check the migration file please

apps/backend/src/app/api/latest/internal/config/override/[level]/route.tsx
```diff
@@ -196,9 +197,42 @@ function findIncludeByDefaultPath(value: unknown, path: string[] = []): string |
...
+// `auth.oauth.providers.<id>` (even credentials-only without `type`), and (2) an enable-field
+// leaf (type/allowSignIn/allowConnectedAccounts), which belongs to the branch. Object-valued
+// credential leaves like `...appleBundles` are fine.
+type EnvProviderViolation = { path: string, kind: "container" | "enable-field" };
+function findEnvProviderOverrideViolation(value: unknown, path: string[] = []): EnvProviderViolation | null {
+  const isProviderPrefix = path[0] === "auth" && path[1] === "oauth" && path[2] === "providers";
+  const isObject = value != null && typeof value === "object" && !Array.isArray(value);
+
+  // A whole provider object: the providers map (depth 3) or one provider (depth 4).
+  if (isProviderPrefix && isObject && (path.length === 3 || path.length === 4)) {
```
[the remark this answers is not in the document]
Comment (nams1570) [reply apps/backend/src/app/api/latest/internal/config/override/[level]/route.tsx]: fixed in followup commit

nams1570:
@greptileai please do a rereview.

apps/backend/src/app/api/latest/internal/config/override/[level]/route.tsx
Comment (N2D4) [inline apps/backend/src/app/api/latest/internal/config/override/[level]/route.tsx:1]: note that the merge from dev may be a little complicated as we have custom OIDC providers now, so you'll need to make sure it's supported

apps/backend/src/app/api/latest/internal/config/override/[level]/route.tsx
```diff
@@ -221,6 +260,16 @@ async function parseAndValidateConfig(
...
+  if (level === "environment") {
+    const disallowedPath = findDisallowedEnvOAuthPath(parsedConfig);
+    if (disallowedPath !== null) {
+      throw new StatusError(
+        StatusError.BadRequest,
+        `The environment config can only set OAuth credentials as individual leaf keys (e.g. auth.oauth.providers.google.clientId). The value at "${disallowedPath}" sets the provider roster or its enabled state, which must be configured on the branch config instead.`,
+      );
+    }
+  }
+
```
Comment (N2D4) [inline apps/backend/src/app/api/latest/internal/config/override/[level]/route.tsx]: this is a hack. this is not where this kind of logic should live. what is this for?

Comment (N2D4) [reply apps/backend/src/app/api/latest/internal/config/override/[level]/route.tsx]: when would a user run into this and what should their action be?

packages/shared/src/config/oauth-providers.ts
```diff
@@ -0,0 +1,141 @@
...
+  })).toEqual({
+    branchEnable: { type: "microsoft", allowSignIn: false, allowConnectedAccounts: false },
+    envCredentialLeafKeys: {
+      "auth.oauth.providers.microsoft.isShared": false,
+      "auth.oauth.providers.microsoft.clientId": "cid",
+      "auth.oauth.providers.microsoft.clientSecret": "secret",
+      "auth.oauth.providers.microsoft.microsoftTenantId": "tenant",
+    },
+  });
+});
```
N2D4: why do we have all these special cases for oauth providers? can't we just update the branchConfigSchema to include the stuff that is configurable on a branhc level, and environmentConfigSchema to include the stuff that is not? we have other config options that are environment-only too

N2D4: i generally don't understand why we need a "split" function besides for the migration (which is in SQL anyways). imagine if we wrote this code as if it had always put auth method providers onto the branch config. in that case, we wouldn't have a notion of "splitting". but that's always the code quality we want to strive for, every bugfix should be fixed in such a way such that the code looks as if we had never made the mistake in the first place

N2D4 (changes requested):
comments

apps/backend/src/app/api/latest/internal/config/override/[level]/route.tsx
Comment (BilalG1) [reply apps/backend/src/app/api/latest/internal/config/override/[level]/route.tsx:1]: Done. custom_oidc is now preserved

apps/backend/src/app/api/latest/internal/config/override/[level]/route.tsx
Comment (BilalG1) [reply apps/backend/src/app/api/latest/internal/config/override/[level]/route.tsx]: removed, the migrateConfigOverride function handles this for environment config now

packages/shared/src/config/oauth-providers.ts
BilalG1: removed, we partition in the schemas instead

N2D4 (commented):
@nams1570 correct me if i'm wrong, this will be obsoleted by the new envconfig/project settings page?

PR #1632, opened 2026-06-19
perf(platform-analytics): cut ClickHouse query peak memory

BilalG1:
## What

Performance pass on the internal **platform-analytics** route. All 17 ClickHouse queries fire in a single `Promise.all` on the shared `stackframe` admin user, which is subject to a **9 GB per-user** memory cap — so the worst case is the *sum* of per-query peaks, not the max. Benchmarked at 10k projects / 1M users / 50M events (power-law, top project ≈100k users), the sum of peaks was ~6.7 GiB. This PR brings it down to ~3.8 GiB.

## Changes

**ClickHouse — `sipHash64(user_id)` as the distinct key** (exact, verified byte-identical):

| query | peak mem | Δ |
|---|---|---|
| `dauSeries` | 949 → 373 MiB | −61% |
| `mauProjects` | 715 → 313 MiB | −56% |
| `activeByProject` | 635 → 374 MiB | −41% |
| `sparkByProject` | 1165 → 809 MiB | −31% |

A 64-bit hash has negligible collision probability over 1M users; the benchmark confirmed identical output. (Same trick already used in the internal-metrics MAU query.)

**ClickHouse — sample the activity split** (`new`/`retained`/`reactivated`):
The split was the single heaviest query (~1.3 GiB) — its cost is a window function over ~25.8M `(user, day)` rows plus an all-history scan, which `sipHash` alone barely helped (−7%). It now uses **consistent 1-in-4 user sampling** (same `cityHash64(user_id) % 4` bucket applied to both subqueries so each sampled user's full activity sequence is preserved; counts scaled ×4):

- **317 MiB (−78%)** peak memory, **~0.4% mean error** (max 1.4% on the smallest day) vs the exact result.

This is an **approximation** — the dashboard "Growth quality" chart now notes it (`subtitle: "… · sampled estimate (~0.4%)"`). `ACTIVITY_SPLIT_SAMPLE` is a single constant in the route; set it to `1` to go back to exact.

## What I tried that did NOT make the cut (documented in the harnesses)

- `country` — peak memory is dominated by the per-user `argMax(country, event_at)` payload, not the key, so hashing does nothing. Left exact/unchanged.
- PG `authMethods` / `email` — with the production composite PK indexes the original plans are already best; correlated-subquery / anti-join rewrites were far worse. No PG query changes in this PR.

## Benchmark harnesses (added)

- `apps/backend/scripts/benchmark-platform-analytics.ts` — full-route baseline (per-query time/memory/rows).
- `apps/backend/scripts/optimize-platform-analytics.ts` — sipHash & PG variant comparison with byte-equality checks.
- `apps/backend/scripts/optimize-split.ts` — exact vs sampled split variants with accuracy measurement.

They seed isolated `bench_pa` databases (server-side, auto-cleaned) and read `system.query_log` / `EXPLAIN (ANALYZE, BUFFERS)`. Run e.g.:
`pnpm --filter @hexclave/backend run with-env:dev tsx scripts/optimize-split.ts`

## Testing

- Backend `typecheck` passes. (Dashboard has pre-existing typecheck errors on the base branch in unrelated files — auth-methods, team-analytics, user-emails, RDE config — not touched here.)
- All exact rewrites verified byte-identical to the originals by the harnesses; the sampled split measured at ~0.4% mean error.

Numbers are local warm-cache (relative shape, not production latency).

---
## Summary by cubic
Cuts worst-case ClickHouse memory for the internal platform analytics route by switching to hashed distinct keys and sampling the heaviest query. On a 10k projects / 1M users / 50M events benchmark, the sum of per-query peaks drops from ~6.7 GiB to ~3.8 GiB with exact results (or ~0.4% error on the sampled chart).

- **Performance**
  - Use sipHash64(user_id) as the distinct key in uniqExact/uniqExactIf for DAU series, MAU/projects, active-by-project, and sparkline. Exact results (verified). Peak memory down 31–61% per query.
  - Sample the new/retained/reactivated split at 1-in-4 users (consistent `cityHash64` bucket across subqueries, counts ×4). Peak memory ~−78% (~1.3 GiB → ~0.3 GiB) with ~0.4% mean error. Toggle via `ACTIVITY_SPLIT_SAMPLE` (set to 4; set to 1 for exact). Dashboard subtitle now notes “sampled estimate (~0.4%).”
  - Added local harnesses to seed isolated data and measure time/memory/equality: `apps/backend/scripts/internal-analytics/benchmark-platform-analytics.ts`, `optimize-platform-analytics.ts`, `optimize-split.ts`.

<sup>Written for commit 60ccf1a06f45cddf19f46b1b8889b2bba78625fe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hexclave/hexclave/pull/1632?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Updates

* **Improvements**
  * Enhanced platform analytics calculations for more consistent and efficient user counting across key performance indicators (DAU, MAU, per-project metrics).
  * Updated the Growth Quality chart to indicate that user counts represent sampled estimates with approximately 0.4% margin of error for improved performance.

### DIFF
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -103,7 +103,7 @@
     "jiti": "^2.6.1",
     "jose": "^6.1.3",
     "json-diff": "^1.0.6",
-    "next": "16.2.7",
+    "next": "16.2.9",
     "nodemailer": "^6.9.10",
     "oidc-provider": "^8.5.1",
     "openid-client": "5.6.4",
--- a/apps/backend/scripts/benchmark-platform-analytics.ts
+++ b/apps/backend/scripts/benchmark-platform-analytics.ts
@@ -1,0 +1,542 @@
+/**
+ * Local-only benchmark harness for the queries in
+ * apps/backend/src/app/api/latest/internal/platform-analytics/route.tsx.
+ *
+ * Seeds an ISOLATED ClickHouse database (`bench_pa`) and Postgres schema
+ * (`bench_pa`) with synthetic data at a configurable scale, then runs every CH
+ * and PG query from the route verbatim (table refs rewritten to the scratch
+ * db/schema) and records duration / memory / rows-read from system.query_log
+ * (CH) and EXPLAIN (ANALYZE, BUFFERS) (PG).
+ *
+ * Defaults model the scale this route is expected to face:
+ *   - 10,000 projects, 1,000,000 users, power-law (top project ~10% of users)
+ *   - 50,000,000 ClickHouse events (~50/user), single branch "main"
+ *
+ * Run:
+ *   pnpm --filter @hexclave/backend run with-env:dev tsx scripts/benchmark-platform-analytics.ts
+ * Env knobs:
+ *   PA_PROJECTS (10000) PA_USERS (1000000) PA_EVENTS (50000000)
+ *   PA_SKIP_SEED=1   reuse an already-seeded bench_pa
+ *   PA_KEEP=1        do not drop bench_pa at the end
+ *   PA_OUT           output json path (default /tmp/platform-analytics-bench.json)
+ *
+ * NOTE: this harness captures the ORIGINAL (pre-optimization) query shapes so it
+ * can be used as a baseline. See optimize-platform-analytics.ts and
+ * optimize-split.ts for the variant comparisons that justify the route changes.
+ */
+import { getClickhouseAdminClient, getClickhouseAdminClientForMetrics } from "@/lib/clickhouse";
+import { globalPrismaClient } from "@/prisma-client";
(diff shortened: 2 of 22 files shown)


mantrakp04 (approved):
lfg! can we move the move the .../scripts/... -> .../scripts/internal-analytics/...

PR #1665, opened 2026-06-24
Add periodic tick loop to bulldozer-js server

BilalG1:
## Summary

Adds a periodic tick loop to the bulldozer-js server that runs every second.

- Introduces `runTickLoop()` which calls `bulldozerDb.withSnapshotReplicated(...)` to run `snapshot.tick(...)` on a 1s interval.
- Uses a monotonic `lastTickMillis` clamp (`Math.max(Date.now(), lastTickMillis)`) so the tick timestamp never moves backwards.
- Errors are caught and logged without breaking the loop.
- Loop is started after the server begins listening.

## Test plan

- [ ] Verify the server starts and the tick loop runs without errors.

---
## Summary by cubic
Adds a 1-second tick loop to the `bulldozer-js` server and reports loop errors to Sentry. Uses a monotonic clamp so tick time never moves backwards and keeps running on failure.

- **New Features**
  - Runs a periodic loop with `runAsynchronously` + `wait`; calls `withSnapshotReplicated(() => snapshot.tick(...))` every second.
  - Clamps `lastTickMillis` via `Math.max(Date.now(), lastTickMillis)` to prevent time from going backwards.
  - Sends loop errors with `captureError("bulldozer-js-tick-loop", ...)`, wired by `initSentry()` using `@sentry/node` (no-op without a DSN).

<sup>Written for commit 20b5c0da15ae13b060ae69a6254c3c569293d13c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hexclave/hexclave/pull/1665?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/bulldozer-js/src/index.ts
+++ b/apps/bulldozer-js/src/index.ts
@@ -570,9 +570,8 @@ const app = new Elysia({ adapter: node() })
 
 console.log(`Bulldozer JS server listening on http://localhost:${app.server?.port ?? port}`);
 
-// Periodically advance the bulldozer clock so timefold-queued rows (subscription renewals, expiries, …)
-// are processed. `lastTickMillis` is clamped monotonically so a backwards wall-clock jump can't rewind it,
-// and no-op ticks are cheap because `withSnapshotReplicated` skips the write when nothing changed.
+// Periodically tick the bulldozer clock to process timefold-queued rows; clamp monotonically so a
+// backwards wall-clock jump can't rewind it.
 runAsynchronously(async () => {
   let lastTickMillis = 0;
   while (true) {
--- a/apps/bulldozer-js/src/index.ts
+++ b/apps/bulldozer-js/src/index.ts
@@ -477,6 +477,17 @@ function ok() {
   return { success: true };
 }
 
+let lastTickMillis = 0;
+async function runTickLoop(): Promise<void> {
+  try {
+    lastTickMillis = Math.max(Date.now(), lastTickMillis);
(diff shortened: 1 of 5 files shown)


apps/bulldozer-js/src/index.ts
```diff
@@ -477,6 +477,17 @@ function ok() {
...
+async function runTickLoop(): Promise<void> {
+  try {
+    lastTickMillis = Math.max(Date.now(), lastTickMillis);
+    await bulldozerDb.withSnapshotReplicated(async (snapshot) => await snapshot.tick(new Date(lastTickMillis)));
+  } catch (error) {
+    console.error("Bulldozer JS tick loop error:", error);
+  }
+  setTimeout(() => { void runTickLoop(); }, 1000);
+}
+
```
N2D4:
```suggestion
runAsynchronously(async () => {
  while (true) {
    try {
      await bulldozerDb.withSnapshotReplicated(async snapshot => await snapshot.tick(new Date()))
    } catch (error) {
      captureError('Bulldozer tick failed. Continuing.', error)
    }
    await wait(1000);
  }
});
```

also make sure that captureError captures errors on sentry — this will probably need extra setup (see in polyfills.ts of the backend to see how that does it)

N2D4 (approved):

PR #1686, opened 2026-06-29
Add back buttons to account settings and purchase pages

Developing-Gamer:
## Summary

Adds "Back" navigation to the hosted purchase and account-settings pages so users can return to the exact page they came from.

**Account settings** (`hosted-components` + SDK):
- `redirectToAccountSettings()` now appends a `back_url` query param with the exact current URL when the target is cross-origin (doesn't overwrite an existing one). `document.referrer` is not used — it only carries the origin cross-origin under the default `strict-origin-when-cross-origin` policy.
- The hosted page reads `back_url`, requires http(s), and only renders it after `isTrustedRedirectUrl(backUrl)` (the SDK's existing trusted-domain check) confirms it — otherwise falls back to `/`, so the hosted origin can't act as an open redirect.
- New internal SDK API: `hexclaveAppInternalsSymbol.isTrustedRedirectUrl(url): Promise<boolean>`; spec updated accordingly.
- `SidebarLayout` accepts `backUrl` and threads it to desktop/mobile layouts; mobile detail view labels the link "Back to site".

**Purchase page** (`apps/dashboard`):
- Validates the `return_url` query param (http/https only) and shows the back link only after `validate-code` succeeds — the server rejects untrusted `return_url`s, so a loaded response implies the URL is safe.

Includes a regression test in `client-app-impl.cross-domain.test.ts` asserting the exact URL (path + query) is preserved in `back_url` on cross-origin account-settings redirects.

Link to Devin session: https://app.devin.ai/sessions/ac8aa8c8d1de48a6a9738d628921f3c8
Requested by: @Developing-Gamer


apps/dashboard/src/lib/hosted-back-url.ts
```diff
@@ -0,0 +1,30 @@
...
+function useExternalBackUrl(): string | null {
+  const referrer = useSyncExternalStore(subscribeToReferrer, getReferrerSnapshot, () => "");
+  return useMemo(() => {
+    if (!referrer || !URL.canParse(referrer)) return null;
+    const parsed = new URL(referrer);
+    if (parsed.protocol !== "https:" && parsed.protocol !== "http:") return null;
+    if (parsed.origin === window.location.origin) return null;
+    return referrer;
+  }, [referrer]);
+}
```
N2D4: shouldn't we check for trusted domains somewhere?

apps/hosted-components/src/hosted-components/account-settings/index.tsx
```diff
@@ -27,6 +27,34 @@ const Icon = ({ name }: { name: keyof typeof iconMap }) => {
 
 const emptyTeams: Team[] = [];
 
+function getReferrerSnapshot(): string {
+  if (typeof document === "undefined") return "";
+  return document.referrer;
+}
```
N2D4: i don't think this will work across different origins (browsers usually only send the origin when doing cross-origin requests, not the full path). i think the cleaner option is a query param that we maintain on navigations (not session storage or ocokie or anything like that)

N2D4 (changes requested):

apps/dashboard/src/app/(main)/purchase/[code]/page-client.tsx
```diff
@@ -247,7 +266,14 @@ export default function PageClient({ code }: { code: string }) {
       <div className="relative flex min-h-screen w-full flex-col lg:flex-row">
         {/* Left Panel: Product & Pricing Selection */}
         <div className="flex flex-1 flex-col border-b border-border/40 bg-white dark:bg-zinc-950 lg:w-1/2 lg:border-b-0 lg:border-r">
-          <div className="mx-auto w-full max-w-md px-6 pb-12 pt-16 lg:pt-20">
+          <div className="mx-auto w-full max-w-md px-6 pb-12 pt-10 lg:pt-12">
+            {/* Only link back once validate-code has succeeded — the server rejects return_urls that
+                are not on the project's trusted domains, so a loaded `data` implies returnUrl is safe. */}
```
Comment (N2D4) [inline apps/dashboard/src/app/(main)/purchase/[code]/page-client.tsx:271]: better to make this explicit instead of implicit like this

Comment (N2D4) [reply apps/dashboard/src/app/(main)/purchase/[code]/page-client.tsx:271]: just check with the trusted domains, much easier and cleaner

N2D4 (approved):

PR #1708, opened 2026-07-01
Remove product-level free trial option; fix custom pricing dropdown overflow

Vedanta-Gawande:
## Summary

This PR removes product-level free trial configuration and makes free trials a price-level setting only.

The important behavior change is not just that the dashboard option is gone. The config/API model now has one clear place for trials:

```ts
payments.products.<productId>.prices.<priceId>.freeTrial
```

instead of also allowing:

```ts
payments.products.<productId>.freeTrial
```

This avoids the old duplicate model where a trial could be configured both on the product and inside the pricing dialog. The pricing dialog is the clearer place for this because a free trial affects a purchase offer, and prices are the purchase offers attached to a product.

This PR also fixes the custom interval dropdown overflow in the pricing UI.

---

## Product-Level Free Trial Removal

Before this PR, products could carry a top-level `freeTrial`, while individual prices could also carry their own `freeTrial`.

That made the behavior harder to reason about:

- product-level trial: applies broadly to the product
- price-level trial: applies to one specific price
- both present: the more specific price-level value needs to win

After this PR, active config and public API schemas only expose price-level free trials.

The dashboard changes remove the product-level free trial option from:

- new product flow
- edit product flow
- product detail view
- product dialog
- product table / product-line prompt surfaces

Price-level free trials remain supported in the pricing dialog.

---

## Config Migration

The delicate part of this PR is backwards compatibility for old config.

If an existing config has:

```ts
payments.products.pro = {
  freeTrial: [7, "day"],
  prices: {
    monthly: { USD: "10" },
    annual: { USD: "100", freeTrial: [14, "day"] },
  },
}
```

the migration rewrites it to:

```ts
payments.products.pro = {
  prices: {
    monthly: { USD: "10", freeTrial: [7, "day"] },
    annual: { USD: "100", freeTrial: [14, "day"] },
  },
}
```

The rule is:

- if a product-level trial exists and a price has no trial, copy the product-level trial to that price
- if a price already has a trial, keep the price-specific value
- after copying, remove the product-level trial
- if the product-level trial has no prices visible in that same override, remove it because there is no override-local price to preserve it on

That last point is intentional. Config migrations mutate override layers, not the final rendered config. The rendered config is read-only output from all layers combined. Writing into the rendered shape would create a value that does not belong to any real source layer, so this migration only writes to prices that are visible in the override object being migrated.

The migration handles the config formats Hexclave already supports:

- nested config objects
- dot-notation override keys
- mixed nested + dot-notation overrides
- dotted product IDs
- product IDs that contain the segment `prices`
- dotted price IDs
- explicit `freeTrial: undefined`, which is treated as missing instead of as a real price-specific override

There is also a validation test that proves legacy product-level config fails the current branch schema before migration, then passes after migration normalizes it to price-level trials.

---

## Inline Product API Compatibility

This PR also updates inline products.

New public inline product schemas no longer expose top-level:

```ts
product_inline.free_trial
```

New callers should use:

```ts
product_inline.prices.<priceId>.free_trial
```

For backwards compatibility, the backend still accepts legacy top-level `product_inline.free_trial` on the relevant purchase/grant routes through a hidden compatibility schema. Before the product is used or returned, that legacy value is normalized onto each inline price that does not already have its own `free_trial`.

The rule mirrors the config migration:

- legacy top-level `product_inline.free_trial` fills missing price-level trials
- explicit price-level `free_trial` wins
- responses do not expose top-level `free_trial`

This keeps existing API callers working while preventing the old field from being advertised in OpenAPI/docs for new integrations.

---

### Delete product-level trials without migration

This would have been the simplest code change, but it could silently change behavior for existing configs. A product that previously had a trial would stop applying that trial after the field was stripped.

### Keep accepting product-level trials everywhere

This would maximize compatibility, but it would keep the old duplicate model alive. New users could still create product-level trials manually, and the dashboard/docs/API would disagree about where trials belong.

### Normalize only at runtime

Runtime-only normalization would avoid persistent config rewrites, but old values would keep flowing through active config validation and tooling. That makes the schema less clear and increases the chance that product-level trials get reintroduced.

### Chosen approach

This PR uses a hybrid compatibility approach:

- public schemas, docs, OpenAPI, and dashboard UI stop exposing product-level trials
- saved config overrides migrate product-level trials onto visible prices before validation
- legacy inline API input is still accepted internally and normalized to price-level trials
- stored historical snapshots can still be read

That gives new users one clear model while still handling old config/API shapes where the code has enough source-layer context to preserve the old intent.

---

## Custom Pricing Dropdown Fix

This PR also fixes a layout bug in `RepeatingInput`.

The custom interval row has:

```text
[number input] [unit dropdown]
```

The number input component internally wraps itself in a full-width element. Inside a flex row, that could consume too much horizontal space and push the unit dropdown out of the popover.

This PR constrains the number input lane with:

```tsx
<div className="w-20 shrink-0">
```

and lets the unit dropdown use the remaining space with `min-w-0 flex-1`.

The behavior of the interval picker is unchanged. This is a layout fix.

Link to Devin session: https://app.devin.ai/sessions/ab98a5e521df4b70bcb3b1b89568891d
Requested by: @Vedanta-Gawande

---
## Summary by cubic
Removed product-level free trials; trials are now set per price. Adds a migration and legacy inline support to keep existing configs working, and fixes the custom pricing dropdown overflow.

- **Refactors**
  - Enforced price-level trials only: product and inline product schemas reject product-level fields; `ProductSnapshot.freeTrial` marked deprecated; seed updated.
  - Config migration moves `payments.products.*.freeTrial` onto each visible price, preserves existing price-level values, handles mixed nested/dot notation and tricky product/price IDs (including product IDs named "prices"), then strips it; also removes branch-level `dbSync.externalDatabases`.
  - API compatibility: latest and `v2beta5` routes accept legacy inline `free_trial`, migrate it to price-level via a wrapper schema and normalizer; added E2E tests for direct grants and purchase-code validation.
  - Dashboard: removed free-trial UI from create/edit/view and the products table.
  - Docs/OpenAPI: removed product-level trial fields; clarified send-email is to users only.

- **Bug Fixes**
  - Constrained the custom pricing input in `repeating-input.tsx` to prevent the unit dropdown from overflowing.

<sup>Written for commit b055b94af625bdc648ee104fd4baf05e14338c3d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hexclave/hexclave/pull/1708?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed product-level free trial settings from the product create/edit flows and product previews, so product pricing setup is now more consistent.
  * Existing saved configurations with legacy free trial values are cleaned up automatically while preserving price-level trial settings.

* **Documentation**
  * Updated product configuration docs to reflect the current settings available.
  * Clarified email sending API docs to cover sending to specific users, all users, or arbitrary email addresses.


packages/shared/src/config/schema.ts
```diff
@@ -607,11 +804,132 @@ import.meta.vitest?.test("migrateConfigOverride removes legacy branch-level dbSy
...
   return mapProperty(obj, pathCond, () => undefined);
 }
 
+function isProductLevelFreeTrialPath(path: string[]): boolean {
+  // p.slice(3, -2) checks for "prices" only in positions where it structurally indicates a price sub-object
+  // (i.e., followed by at least a price-ID segment before the trailing "freeTrial"). This correctly handles
+  // dotted product IDs containing "prices" as a segment (e.g., "pro.prices") — those land at p.length-2
+  // which is excluded by the slice, so the product-level freeTrial is still considered product-level.
+  return path.length >= 4 && path[0] === "payments" && path[1] === "products" && path[path.length - 1] === "freeTrial" && !path.slice(3, -2).includes("prices");
+}
```
nams1570, line 817: **Bug: **a price whose id is **`freeTrial`** gets deleted on config migrate

`isProductLevelFreeTrialPath` treats a path as product-level if it ends in `freeTrial` and `path.slice(3, -2)` does not contain `"prices"`.

For a price with id `freeTrial`, the path is: `payments.products.<productId>.prices.freeTrial`

Here `slice(3, -2)` is empty, so the path is treated as product-level. The final `removeProperty(...)` then deletes the whole price object.

This does not require a legacy product-level trial. `freeTrial` is a valid price id under `userSpecifiedIdSchema`, so this config is legal today.

apps/backend/src/lib/payments.tsx
```diff
@@ -110,6 +115,62 @@ export async function ensureProductIdOrInlineProduct(
   }
 }
 
+export function normalizeInlineProductLegacyProductLevelFreeTrial(
```
nams1570, line 118: Potential Bug: omitted **`prices`** + legacy **`free_trial`** causes a 500

`inlineProductSchema` does not mark `prices` as `.defined()`, so a body with product-level `free_trial` and no `prices` key passes request validation. `normalizeInlineProductLegacyProductLevelFreeTrial` then calls `typedEntries` on `undefined` and throws `TypeError`, which surfaces as a 500.

apps/backend/src/app/api/latest/payments/purchases/create-purchase-url/route.ts
```diff
@@ -1,12 +1,12 @@
...
 import { validateRedirectUrl } from "@/lib/redirect-urls";
 import { getHexclaveStripe, getStripeForAccount } from "@/lib/stripe";
 import { getPrismaClientForTenancy, globalPrismaClient } from "@/prisma-client";
 import { createSmartRouteHandler } from "@/route-handlers/smart-route-handler";
 import { KnownErrors } from "@hexclave/shared/dist/known-errors";
-import { adaptSchema, clientOrHigherAuthTypeSchema, inlineProductSchema, urlSchema, yupNumber, yupObject, yupString } from "@hexclave/shared/dist/schema-fields";
+import { adaptSchema, clientOrHigherAuthTypeSchema, urlSchema, yupNumber, yupObject, yupString } from "@hexclave/shared/dist/schema-fields";
 import { getEnvVariable } from "@hexclave/shared/dist/utils/env";
 import { throwErr } from "@hexclave/shared/dist/utils/errors";
 import { purchaseUrlVerificationCodeHandler } from "../verification-code-handler";
```
nams1570, line 12: <p class='graphite__hidden'><i>[Re: line +20]</i></p>
potential bug: with this route and the other one, wouldn't empty prices + free trial on product fail on the normalize step? that would become a 500 right? wouldnt we want it to be a 400 for malformed input instead?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/hexclave/hexclave/1708?utm_source=unchanged-line-comment">Graphite</a>.</p>

nams1570 (changes requested):
Found some bugs. Code quality wise it's mostly fine, though I think you could have deduped some of the schema code.

PR #1731, opened 2026-07-03
feat(dashboard): manual item quantity adjustments from customer payments view

BilalG1:
Adds a way for dashboard users to create a **manual item quantity change** — positive or negative — for a single customer and a single item, directly from the customer payments view.

**Base:** `dev` → **Head:** `feat/manual-item-quantity-change` · 4 files, +219/−10

## Demo

Full flow on a user detail page: open the Payments tab, hit the ± button on an item balance row, apply **+5** with a description, then **−3** — balance and transaction history update live.

Manual item quantity change flow

<sub>WebM version: [trimmed.webm](https://gist.githubusercontent.com/BilalG1/f55a8c9055cd0626320fb3153371fa97/raw/trimmed.webm)</sub>

## Screenshots

| Adjust dialog (positive) | Adjust dialog (negative) |
| --- | --- |
| Positive change with live preview | Negative change |

| After applying changes | Entry point on item balance rows |
| --- | --- |
| Balance + transaction history updated | Item balances card |

## What's new

- **`ItemQuantityChangeDialog`** (`apps/dashboard/src/components/payments/item-quantity-change-dialog.tsx`)
  - Signed integer quantity input (`10` adds, `-5` subtracts), validated as a non-zero whole number, with a live `New balance: X → Y` preview.
  - Optional description, stored on the transaction.
  - Submits via `adminApp.createItemQuantityChange` (which passes `allow_negative: true`, so balances may go negative — consistent with the existing admin API semantics).
- **Entry point** — a ± icon button on every row of the *Item balances* card in `customer-payments-section.tsx`. Since that card is shared, the feature works on the **user detail page**, **team detail page**, and the payments **Customers page**, i.e. all three customer types (`user` / `team` / `custom`).
- **SDK cache fix** (`packages/template`, SDKs regenerated) — `createItemQuantityChange` previously left the item and transaction caches stale, so a dashboard caller would keep seeing the old quantity. It now refreshes the affected item cache (new protected `_refreshItemCache` helper on the server app impl, mirroring what `item.increaseQuantity` already did) and invalidates the transactions cache. This is what makes the balance and transaction history update immediately after applying a change.

## Notes for reviewers

- The dialog intentionally does not expose `expiresAt` (the API supports it) — kept the surface minimal for the first pass.
- Transaction history already knew how to render `manual-item-quantity-change` transactions (`api_credits (+25)` detail rows), so no changes were needed there.
- The per-row dialog receives `currentQuantity` from the same `useItem` hook that renders the row, so the preview can never disagree with the displayed balance.

## Test plan

- [x] `pnpm typecheck` and `pnpm lint` green
- [x] Manual e2e against the seeded Demo Project (production builds of dashboard + backend): created a user-scoped `api_credits` item, applied `+25`, `−10`, `+5`, `−3` on a user — balance updated live (25 → 15 → 20 → 17), each change produced an *Item quantity change* transaction, success/error toasts behave, invalid input (`0`, non-integers) disables/blocks Apply
- [x] Verified the adjust button renders for user-scoped items on the user page (team/customers pages use the identical shared component)

## Summary by CodeRabbit

* **New Features**
  * Added an item quantity adjustment dialog for customer payments, with live preview and “Apply change” support for manual whole-number deltas (non-zero), plus success/error toasts.
  * Payment item rows now include quick-action controls to open the adjustment dialog directly.
* **Bug Fixes**
  * Quantity updates now immediately refresh the affected item details and keep related transaction data in sync after changes.
  * Improved validation to reject invalid quantity inputs and provide clearer inline feedback during submission.

### DIFF
--- a/apps/dashboard/src/components/payments/customer-payments-section.tsx
+++ b/apps/dashboard/src/components/payments/customer-payments-section.tsx
@@ -3,6 +3,7 @@
 import {
   DesignBadge,
   type DesignBadgeColor,
+  DesignButton,
   DesignCard,
 } from "@/components/design-components";
 import { cn, Skeleton } from "@/components/ui";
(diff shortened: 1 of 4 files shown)


mantrakp04 (approved):
Gs

PR #1768, opened 2026-07-14
Fix recurring production Sentry errors

BilalG1:
Fixes four recurring production Sentry issues (triaged from the last 24h of `vercel-production`):

## X OAuth: 403 client-not-enrolled → 400 (STACK-BACKEND-1JE)
When a customer's X developer App isn't attached to an X API Project, X deterministically rejects `GET /2/users/me` with 403 `client-not-enrolled` even though the token exchange succeeds. This previously surfaced as a `HexclaveAssertionError` (opaque 500 + Sentry alert) on every sign-in attempt for the misconfigured project. It's now a 400 `StatusError` telling the developer to attach their App to a Project, mirroring the existing `invalid_client` handling in `providers/base.tsx`. Other non-OK statuses still fail loud.

## Stripe: ignore `refund.updated` (STACK-BACKEND-1D9)
Stripe emits `refund.updated` asynchronously when the card network fills in `destination_details` on refunds we created — and already ledgered — synchronously in the internal refund route, so it carries no actionable information. Added it to `ignoredEvents` (same approach as #1667/#1461). `refund.failed` deliberately stays fail-loud, since a refund failing after it's ledgered is actionable.

## external-db-sync: fix enqueue deadlocks (STACK-BACKEND-16P)
`enqueueExternalDbSyncBatch` batch-inserts deduplication keys with `ON CONFLICT DO NOTHING`; two concurrent batches inserting overlapping keys in different orders could deadlock (Postgres 40P01, surfaced as an empty-titled `PrismaClientKnownRequestError`). Worse than the noise: the flag-clearing UPDATE and the enqueue run in separate transactions, so a deadlocked enqueue silently lost that batch's sync trigger until the next change for the tenancy. Fixed by deduplicating and inserting keys in canonical (sorted) order, so the circular wait between two batches can't form. Deliberately minimal: deadlocks against other `OutgoingRequest` writers (e.g. `recoverStaleOutgoingRequests`) are theoretically still possible but not evidenced in the Sentry events; if 40P01 reappears, a retry can be added then.

## Password reset: 404 instead of 500 for deleted users (STACK-BACKEND-1JD)
Password reset codes embed the `user_id` at issuance time, so a code can be redeemed after the user was deleted (observed in prod from a scripted client replaying a reset code). The `usersCrudHandlers.adminUpdate` call didn't declare `UserNotFound` via `allowedErrorTypes`, so the CRUD handler wrapped it in `CrudHandlerInvocationError` and the endpoint returned an internal 500. Now declares it and lets the client-safe `USER_NOT_FOUND` KnownError propagate as a 404, matching the ~10 sibling call sites that already do this.

## Testing
- New e2e test: resetting the password of a deleted user returns 404 `USER_NOT_FOUND` (`auth/password/reset.test.ts`) — verified passing locally against the full stack
- Backend + e2e typecheck and lint pass
- No new tests for the other three: the ignore-list entry is covered by `satisfies Stripe.Event.Type[]` and the existing unknown-type webhook tests; the X fix would need the X API base URL to be injectable plus a mock user-info endpoint (possible follow-up); the deadlock is a Postgres timing race that can't be deterministically reproduced through the e2e API surface

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved password reset handling when the account no longer exists.
  * Added clearer guidance for X API configuration errors.
  * Prevented unsupported Stripe refund update events from triggering processing.
  * Ensured external database synchronization requests are deduplicated and consistently ordered.

* **Tests**
  * Added coverage for password reset attempts involving deleted accounts.

### DIFF
--- a/apps/backend/src/lib/external-db-sync-queue.ts
+++ b/apps/backend/src/lib/external-db-sync-queue.ts
@@ -1,4 +1,3 @@
-import { Prisma } from "@/generated/prisma/client";
 import { globalPrismaClient } from "@/prisma-client";
 import { HexclaveAssertionError } from "@hexclave/shared/dist/utils/errors";
 
@@ -16,10 +15,6 @@ export async function enqueueExternalDbSync(tenancyId: string): Promise<void> {
   await enqueueExternalDbSyncBatch([tenancyId]);
 }
 
-function isDeadlockError(error: unknown): boolean {
-  return error instanceof Prisma.PrismaClientKnownRequestError && error.code === "P2010" && error.meta?.code === "40P01";
-}
-
 // Queues sync requests for multiple tenants in a single query.
 // Only inserts for tenants that don't already have a pending request.
 export async function enqueueExternalDbSyncBatch(tenancyIds: string[]): Promise<void> {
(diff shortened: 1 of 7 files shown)


apps/backend/src/oauth/providers/x.tsx
```diff
@@ -34,6 +34,9 @@ export class XProvider extends OAuthBaseProvider {
     );
     if (!fetchRes.ok) {
       const text = await fetchRes.text();
+      if (fetchRes.status === 403 && text.includes("client-not-enrolled")) {
+        throw new StatusError(400, "X rejected the user info request because the X developer App is not attached to an X API Project. Attach the App to a Project in the X developer portal (https://developer.x.com/en/docs/projects/overview), then try signing in again.");
+      }
```
nams1570, line 39: Discussion: do you think we should return 400 status errors for any `fetchRes.status` = 400 case? Or would you rather they fall back to 500s and we can deal with them individually?

apps/backend/src/lib/external-db-sync-queue.ts
```diff
@@ -37,7 +39,8 @@ export async function enqueueExternalDbSyncBatch(tenancyIds: string[]): Promise<
       ),
       NULL,
       'sentinel-sync-key-' || t.tenancy_id
-    FROM unnest(${tenancyIds}::uuid[]) AS t(tenancy_id)
+    FROM unnest(${sortedTenancyIds}::uuid[]) AS t(tenancy_id)
+    ORDER BY t.tenancy_id
```
nams1570, line 43: Nice catch, good change.

nams1570: Is it possible to add test coverage for this? the function is exported right? cant we try with diff ordered tenancies?

nams1570 (approved):
Approved, nice work, though see if you can add test coverage

apps/backend/src/oauth/providers/x.tsx
BilalG1: we should throw 500, in most cases besides this one a 4xx response on this request is our fault, not our user's

apps/backend/src/lib/external-db-sync-queue.ts
BilalG1: yep good idea

PR #1771, opened 2026-07-15
Retry GitHub user-info 401s to absorb token propagation lag

BilalG1:
Fixes the recurring production Sentry issue STACK-BACKEND-15J (~10 events/day, 282 lifetime): end-users signing in with GitHub intermittently hit a 500 at the last step of the OAuth flow.

## What happens
The token exchange with GitHub succeeds, but the immediate follow-up `GET https://api.github.com/user` sometimes returns 401 "Bad credentials" — for a token GitHub issued milliseconds earlier (observed with `ghu_` GitHub App user-to-server tokens on customer projects). Since any non-OK status threw a `HexclaveAssertionError`, the user got an opaque 500 and Sentry paged.

## Why GitHub does this
This is a known GitHub behavior: token issuance and the REST API run on separate distributed infrastructure, and newly-minted tokens take ~1–3 seconds to propagate to all API servers. Community reports match our symptom exactly (intermittent post-issuance 401s from `/user` that succeed on retry):
- https://github.com/orgs/community/discussions/162975
- https://github.com/orgs/community/discussions/118908

## The fix
Retry the user-info request on 401 with 1s and 2s delays, so the final attempt lands ~3s after the first — covering the reported propagation window while staying well under the serverless request watchdog. Most affected sign-ins should now succeed transparently instead of erroring at all.

If all attempts fail, we deliberately still throw the `HexclaveAssertionError` (500) rather than mapping to a 4xx — now including the attempt count. Rationale: a 401 that survives 3 seconds of retries is not explainable by propagation lag alone; the other candidate cause is a racing duplicate callback whose second code exchange invalidates the first token. We can't distinguish the two server-side, so we keep failing loud. Post-deploy, STACK-BACKEND-15J's volume is the experiment: if it drops to ~zero, propagation lag was the whole story; if events persist (now tagged with `attempt: 3`), that's evidence for the invalidation race and the surviving 401s can then confidently be mapped to a "please try signing in again" 4xx.

Other non-OK statuses (403, 5xx, …) are deliberately not retried — they're deterministic, and retrying would just add 3s of latency to failure modes a retry can't help.

## Testing
- Four in-source unit tests (`import.meta.vitest`, following the precedent in `src/oauth/index.tsx`) with a stubbed global `fetch` and fake timers:
  - success on first attempt → no retry
  - 401 → 401 → 200 → succeeds, with exact 1s/2s delay boundaries asserted
  - persistent 401 → throws with `attempt 3` in the message after exactly 3 calls
  - 403 → throws immediately after 1 call (no retry tax on non-401 failures)
- No e2e test: `api.github.com` is hardcoded and the e2e harness has no mock for GitHub's user-info endpoint; making the base URL injectable is a possible follow-up.
- Backend typecheck and lint pass.

---
## Summary by cubic
Retry the GitHub user-info request on 401 to absorb token propagation lag and stop intermittent 500s at the end of the GitHub OAuth flow. Non-401 failures behave as before.

- **Bug Fixes**
  - Retries 401s from `https://api.github.com/user` at +1s and +2s; cancels intermediate 401 response bodies between retries; if all attempts fail, throws a `HexclaveAssertionError` with the attempt count in error extras (not the message) to keep Sentry grouping stable.
  - Skips retries for non-401 statuses to avoid unnecessary latency.
  - Unit tests cover: no-retry success, 401→retry→success, exhausted 401 retries, and immediate failure on non-401.

<sup>Written for commit e8e340c6a48cebbfc21d8b74982d15a64e5d0d9c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hexclave/hexclave/pull/1771?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GitHub sign-in reliability by retrying GitHub profile fetches when authorization fails with temporary 401 errors.
  * Added more detailed failure reporting when profile information can’t be retrieved after retries.
  * Ensured retries only occur for 401 responses, not for other error statuses.
* **Tests**
  * Added coverage for no-retry on first success, correct retry timing, and correct error behavior after retry exhaustion.

### DIFF
--- a/apps/backend/src/oauth/providers/github.tsx
+++ b/apps/backend/src/oauth/providers/github.tsx
@@ -1,7 +1,51 @@
 import { HexclaveAssertionError, StatusError } from "@hexclave/shared/dist/utils/errors";
 import { getJwtInfo } from "@hexclave/shared/dist/utils/jwt";
+import { wait } from "@hexclave/shared/dist/utils/promises";
 import { OAuthUserInfo, validateUserInfo } from "../utils";
 import { OAuthBaseProvider, TokenSet } from "./base";
+
+// GitHub occasionally rejects a freshly-issued access token with 401 "Bad credentials" because
+// token issuance and the REST API run on separate infrastructure, and new tokens reportedly take
+// 1-3 seconds to propagate (https://github.com/orgs/community/discussions/162975). The delays
+// below are chosen so the final attempt lands ~3s after the first, covering that window while
+// staying well under the serverless request watchdog.
+const USER_INFO_401_RETRY_DELAYS_MS = [1000, 2000];
+
+// Returns `any` because fetch's json() is `any` and the shape is validated by validateUserInfo
+// at the call site; typing GitHub's response here would be pretend-precision.
+async function fetchRawGithubUserInfo(tokenSet: TokenSet): Promise<any> {
+  for (let attempt = 1; ; attempt++) {
+    const rawUserInfoRes = await fetch("https://api.github.com/user", {
+      headers: {
+        Authorization: `Bearer ${tokenSet.accessToken}`,
+        "X-GitHub-Api-Version": "2022-11-28",
+      },
+    });
+    if (rawUserInfoRes.ok) {
+      return await rawUserInfoRes.json();
+    }
+    // Only 401 is retried: it's the propagation-lag symptom, while other statuses are
+    // deterministic and retrying them would just add latency. A 401 that survives all retries
+    // deliberately stays a HexclaveAssertionError (500) rather than a 4xx: the other candidate
+    // cause is a racing duplicate callback whose second code exchange invalidates the first
+    // token, and we can't distinguish the two from here. If Sentry keeps reporting this error
+    // (with the attempt count below) after the retries shipped, that's evidence for the
+    // invalidation race, and the surviving 401s can then confidently be mapped to a
+    // "please try signing in again" 4xx.
+    if (rawUserInfoRes.status !== 401 || attempt > USER_INFO_401_RETRY_DELAYS_MS.length) {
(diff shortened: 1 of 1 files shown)


apps/backend/src/oauth/providers/github.tsx
```diff
@@ -40,23 +75,7 @@ export class GithubProvider extends OAuthBaseProvider {
...
-        hasAccessToken: !!tokenSet.accessToken,
-        hasRefreshToken: !!tokenSet.refreshToken,
-        accessTokenExpiredAt: tokenSet.accessTokenExpiredAt,
-        jwtInfo: await getJwtInfo({ jwt: tokenSet.accessToken }),
-      });
-    }
-    const rawUserInfo = await rawUserInfoRes.json();
+    const rawUserInfo = await fetchRawGithubUserInfo(tokenSet);
 
     const emailsRes = await fetch("https://api.github.com/user/emails", {
```
nams1570, line 80: So this is hitting the same host as the /users endpoint, wouldn't this also suffer from the replication lag?

nams1570 (commented):
I'm a little wary since we don't have official github documentation of this, but it seems reasonably supported. Though the second linked PR does not make the claim your desc says it does. 

https://github.com/octokit/auth-app.js/issues/589

This one seems to indicate the same issue.

apps/backend/src/oauth/providers/github.tsx
BilalG1: it's possible but leaving for now since we haven't had issues with this endpoint on sentry. will add same retry mechanism if issue shows up

nams1570 (approved):

PR #1773, opened 2026-07-16
Treat canceled-at-period-end subscriptions as still in effect

BilalG1:
## Bug

Repro: buy a team plan (test mode), then cancel it from the account settings payments tab.

- The plan immediately shows up as **"Team — One-time purchase"** in Active plans: no end date, no indication it's winding down, and the cancel button disappears into nothing.
- Entitlements (e.g. `dashboard_admins` seats) stay at the paid plan's level until the period ends — which is correct (the cancel dialog promises exactly that) — but nothing in the UI or API explains why.

## Root cause

Canceling a non-Stripe-backed subscription writes `status: canceled` with `endedAt` set to the end of the current period, so the customer stays entitled until then. But the read paths classified subscriptions with `isActiveSubscription` (`active`/`trialing` only), so during the paid-through window the owned product had no recognized subscription and fell through to the one-time-purchase branch. One predicate was answering two different questions: "will this renew / can I act on it?" and "is this currently entitling the customer?"

## Fix

- New `isSubscriptionInEffect(sub, nowMillis)` in `lib/payments.tsx`: status-agnostic, `endedAt`-based — the same semantics Bulldozer's grant timefold and `ensure-free-plan.ts` already use (the latter now reuses the shared predicate).
- Read sites use it: product list classification, plan-usage resolution, switch route's sub-vs-OTP occupancy check. Write sites (cancel, switch-from) keep `isActiveSubscription`, so winding-down subs can't be re-canceled or switched from.
- Product list: `is_cancelable` is now computed (`active && !cancel_at_period_end`) instead of hardcoded `true`.
- Cancel route: double-cancel returns "already canceled, ends at the end of the current billing period" instead of falsely claiming the product is a one-time purchase.
- Stripe-backed cancel now uses `subscriptions.update({ cancel_at_period_end: true })` instead of `subscriptions.cancel()` (which ends immediately) — matching the confirmation dialog's "takes effect at the end of the billing period" and the local-sub semantics. The webhook sync handles the period-end transition and free-plan regrant as before.
- Account settings panels (template + dashboard) render **"Ends on \<date\>"** for winding-down subscriptions; the cancel button disappears via `is_cancelable`.

## Tests

- Updated three inline snapshots in `products.test.ts` that had captured the buggy post-cancel behavior (`"type": "one_time"`, `"subscription": null`).
- New test: post-cancel listing keeps `type: "subscription"` with `cancel_at_period_end: true` / `is_cancelable: false`, and a second cancel returns the dedicated 400.
- `products`, `switch-plans`, `billing`, `purchase-session` e2e suites green; typecheck + lint green on backend, template, dashboard, e2e.
- Verified in the dashboard: the repro team's plan now shows "Team — Ends on Aug 16, 2026" with no cancel button, and Billing & Usage keeps reporting the paid plan during the wind-down window.

---
## Summary by cubic
Keeps cancel-at-period-end subscriptions in effect through the paid-through period. Previously they read as one‑time and usage fell to free; now API/UI keep them as subscriptions, entitlements persist until the boundary, and cancel/switch/purchase flows respect the wind‑down state.

- Products, usage, and reads
  - Use a new status‑agnostic predicate: isSubscriptionInEffect. Plan usage resolves from the in‑effect plan sub; ensure‑free‑plan uses the same predicate.
  - Product listing picks a representative sub with subscriptionDisplayRank (cancelable > active > in‑effect) and computes is_cancelable via isSubscriptionCancelable.
  - Switch logic treats in‑effect subs as sub‑backed; a canceled‑but‑in‑effect local sub is switchable via isSubscriptionSwitchable. Competing‑sub checks use in‑effect, not active‑only.
  - UI shows “Ends on …” when cancel_at_period_end; stackable partial cancels no longer hide the cancel button.

- Cancel, refund, Stripe, and purchases
  - Cancel at period end for Stripe subs; write canceledAt/endedAt eagerly and refresh currentPeriodStart/End from Stripe’s response via getStripeSubscriptionPeriod. Local cancels roll periods forward with getCurrentBillingPeriod so “Ends on …” matches the entitlement boundary.
  - Cancel without subscription_id cancels all cancelable subscriptions for the product. Batch cancel skips rows Stripe already ended; if none are canceled, returns 400.
  - Switch on a Stripe sub clears cancel_at_period_end and resets local canceledAt/endedAt; webhook sync reconciles final state.
  - Purchase session rejects repurchasing a winding‑down Stripe sub; wind‑down replacements proceed only for non‑Stripe subs.
  - Webhook sync derives endedAt/canceledAt for both terminal and non‑terminal updates (pending‑cancel schedules endedAt; reversal clears it). Helpers moved to lib/stripe, plus isStripeSubscriptionAlreadyTerminalError.

- Rollout
  - Apply the SQL migration to backfill endedAt on terminal rows.
  - Run scripts/bulldozer-payments-init.ts after deploy to re‑emit affected rows so grants expire correctly.

<sup>Written for commit c7dc856697c195acf28354b918a65ef981d3e8b0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hexclave/hexclave/pull/1773?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Subscription handling now consistently uses “in-effect” state through the end of the billing period, including plan usage resolution.
  * Product listings and payments panels now present correct end-of-period messaging and cancelability for cancel-at-period-end subscriptions.

* **Bug Fixes**
  * Stripe-backed cancellations and switches now use cancel-at-period-end semantics, derive end timestamps from Stripe, and keep wind-down fields aligned after switching.
  * Purchase validation and replacement/conflict detection now correctly treat winding-down subscriptions as replaceable.

* **Chores / Tests**
  * Added unit, e2e, and migration backfill coverage for terminal `endedAt` behavior and updated expectations for cancel-at-period-end states.


apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/route.ts
```diff
@@ -65,12 +66,22 @@ export const GET = createSmartRouteHandler({
...
-      Object.values(subMap).filter(s => isActiveSubscription(s)).map(s => [s.productId ?? "__null__", s] as const)
-    );
+    // so we normalize subscription productIds to match. In-effect, not
+    // active-only: a winding-down sub still backs its product and must not
+    // fall through to the one_time branch below; productId ties pick the
+    // highest subscriptionDisplayRank.
+    const nowMillis = Date.now();
+    const inEffectSubByProductId = new Map<string, SubscriptionRow>();
+    for (const s of Object.values(subMap)) {
+      if (!isSubscriptionInEffect(s, nowMillis)) continue;
```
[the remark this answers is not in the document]
Comment (BilalG1) [reply apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/route.ts:78]: added a migration to fix this

apps/backend/src/lib/stripe.tsx
```diff
@@ -333,6 +355,19 @@ export async function syncStripeSubscriptions(stripe: Stripe, stripeAccountId: s
...
+        // webhooks, or a purchase-session sub whose first local write is a
+        // late webhook). Without this, a terminal row is created with
+        // endedAt = null, which isSubscriptionInEffect reads as "entitled
+        // forever". Note: rows created before this derivation existed can
+        // still hold that bad shape (terminal status, endedAt = null) and
+        // only self-heal when another webhook for the same customer sweeps
+        // them — a backfill migration (status IN terminal AND endedAt IS
+        // NULL) is tracked as a follow-up, since it also needs the bulldozer
+        // rows re-emitted.
+        ...getEndedAtForSync(subscription, sanitizedDates.end),
```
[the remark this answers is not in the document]
BilalG1: added migration

apps/backend/src/lib/stripe.tsx
```diff
@@ -236,9 +236,13 @@ const getTenancyFromStripeAccountIdOrThrow = async (stripe: Stripe, stripeAccoun
 
 const TERMINAL_STRIPE_STATUSES = ["canceled", "incomplete_expired", "unpaid"] as const;
 
-function getEndedAtForSync(subscription: Stripe.Subscription, sanitizedEnd: Date): { endedAt: Date } | {} {
+export function getEndedAtForSync(subscription: Stripe.Subscription, sanitizedEnd: Date): { endedAt: Date | null } {
```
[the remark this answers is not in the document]
BilalG1: pre-existing problem, can fix if it becomes an issue

nams1570:
@greptileai please rereview

apps/backend/src/lib/payments.tsx
```diff
@@ -116,11 +116,51 @@ export async function ensureProductIdOrInlineProduct(
 // getCustomerPurchaseContext, OwnedProduct type, getOwnedProductsForCustomerLegacy
 // were removed. All reads now go through customer-data.ts backed by Bulldozer.
 
+/**
+ * The subscription will renew and can be acted on (canceled, switched). For
+ * "does this still entitle the customer" use `isSubscriptionInEffect` — a
+ * canceled sub keeps entitling until its future `endedAt`.
```
nams1570: this doc is a little confusing. the first line isnt super clear to me, and on stripe when you set a sub to cancel at period end its status is still active. From stripe's pov, a canceled sub maps to our ended sub.

apps/backend/src/lib/payments.tsx
```diff
@@ -116,11 +116,51 @@ export async function ensureProductIdOrInlineProduct(
...
+ */
 export function isActiveSubscription(subscription: { status: string }): boolean {
   const s = subscription.status;
   return s === "active" || s === SubscriptionStatus.active || s === "trialing" || s === SubscriptionStatus.trialing;
```
nams1570: not related to ur changes, but isnt this redundant?

apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/switch/route.ts
```diff
@@ -116,15 +116,17 @@ export const POST = createSmartRouteHandler({
...
       );
       const hasOtpInProductLine = Object.entries(ownedProducts).some(
         ([productId, p]) => productId !== body.from_product_id
           && p.productLineId === fromProduct.productLineId
           && p.quantity > 0
-          && !activeSubProductIds.has(productId)
+          && !subBackedProductIds.has(productId)
       );
       if (hasOtpInProductLine) {
         throw new StatusError(400, "Customer already has a one-time purchase in this product line");
```
Comment (nams1570) [inline apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/switch/route.ts:135]: <p class='graphite__hidden'><i>[Re: line +172]</i></p>
Bug:  
Notwithstanding the fact that we should prolly rewrite this route, i think this should use isSubscriptionInEffect instead of isActiveSubscription no?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/hexclave/hexclave/1773?utm_source=unchanged-line-comment">Graphite</a>.</p>

apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/[product_id]/route.ts
Comment (nams1570) [inline apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/[product_id]/route.ts:1]: Bug: Cancel correctly sets `endedAt` from Stripe’s update response but doesn’t update `currentPeriodEnd`. UI “Ends on” uses `currentPeriodEnd`, so a stale local period (e.g. missed renewal sync) shows the wrong end date while entitlements stay correct. Please also write period start/end from the Stripe response here (and on the refund at-period-end path). I managed to reproduce this locally too

nams1570 (changes requested):

apps/backend/src/lib/payments.tsx
BilalG1: reworded it to be simpler

apps/backend/src/lib/payments.tsx
BilalG1: yeah don't think there is any point in that. Small fix so I'm just including in this PR

apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/[product_id]/route.ts
Comment (BilalG1) [reply apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/[product_id]/route.ts:1]: good catch, fixed and added a test for that

apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/switch/route.ts
Comment (BilalG1) [reply apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/switch/route.ts:135]: yeah should use isSubscriptionInEffect here, fixed

apps/backend/src/lib/payments.tsx
nams1570, line 1: `grantProductToCustomer` sets `currentPeriodEnd = now + price.interval` once and nothing ever moves it. cancel copies that into `endedAt`.

ex: grant monthly on jan 1 → period end jan 31. user cancels on mar 15. we still schedule the end for **jan 31**, which is already past, so they lose the product immediately. or they never cancel: the row stays `active` with “ends on jan 31” sitting in the past forever. there’s no stripe/renewal to roll the period forward.

is that what we want for test mode, or should cancel do something like `max(storedEnd, now)` / `now + interval`?

apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/switch/route.ts
Comment (nams1570) [inline apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/switch/route.ts:1]: potential bug? stripe pending-cancel stays `status=active`, so `from=pro → to=plus` works and un-cancels. test-mode cancel writes `status=canceled` with a future `endedAt`, so the same call 400s even though GET still shows the product.

ex: test-mode user on basic, hits cancel, still sees basic until “period end”, tries to switch basic → pro, gets “cannot be switched.” free → pro is already blocked because we use `isSubscriptionInEffect` there. this lookup didn’t.

if the rule is “you can switch the sub you still have,” this should be in-effect, not active.

apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/[product_id]/route.ts
Comment (nams1570) [inline apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/[product_id]/route.ts:1]: we should prolly document better that cancel w/o a sub id cancels all subs corresponding to that product

Comment (nams1570) [inline apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/[product_id]/route.ts:1]: Also I think in refund route we have a guard in case sub was already canceled, which we dont here. if you try to cancel a sub that was already canceled it becomes a 500.

Comment (BilalG1) [reply apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/[product_id]/route.ts:1]: added

Comment (BilalG1) [reply apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/[product_id]/route.ts:1]: fixed, added try catch to our stripe call

apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/switch/route.ts
Comment (BilalG1) [reply apps/backend/src/app/api/latest/payments/products/[customer_type]/[customer_id]/switch/route.ts:1]: fixed and added test

apps/backend/src/lib/payments.tsx
BilalG1: I think test mode should be as close as possible to real stripe sub, so cancel shouldn't make user immediately lose product if it is past first period end. Updating so test mode subs cancel at end of their current billing period

nams1570 (approved):

PR #1867, opened 2026-08-04
feat(observability): issues expansion, alert surface and attachments

mantrakp04:
Expands the issues backend from materialization into the full product surface.

### Scope
- ownership (users/teams, ownership rules), activity feed, comments, subscriptions, bookmarks, priority, and saved search views
- the public issues API, the public search projection/query, and the internal merge/unmerge routes
- the **issue alert surface**: rule DSL, evaluator, cooldown claims, delivery ledger, and the workflow integration that delivers them
- the attachment ledger and the payload scrubber

### Why alerts are in this PR rather than their own
Issue materialization dispatches alert signals (`issue-store` → `issue-alerts/ingestion`), while the alert routes read the public issue API. The two import each other, so they cannot be ordered into separate PRs without a production refactor. They ship together.

Attachments and the scrubber are here for the same reason: the issue detail endpoint renders attachments and scrubs payloads when projecting an occurrence, so they must sit at or below this layer.

### ClickHouse
Two columns on `analytics_internal.logs`:
- `issue_grouping_provenance` — ordered primary/secondary hash decisions, stored as a `String` rather than a dynamic JSON path so fingerprint tokens cannot expand ClickHouse's shared dynamic-column namespace
- `error_envelope` — the bounded canonical envelope as one JSON string, for the same reason

`$error` rows written through the batch path get a populated envelope here. Non-error log rows carry `'{}'`. The Sentry ingest path that also populates it lands in the next PR.

### Migration
`20260731000000_add_issues`. Every table is new, so indexes and FK validation are O(1). The composite unique key on `Tenancy` that these scope FKs reference is created concurrently by `20260726000000_add_releases`, which now orders before this migration.

New coverage: `issue-product-metadata`, `issue-saved-search-views`, `issue-grouping-provenance`, `issue-alert`, and `error-attachments` schema/constraint tests.


mantrakp04:
Superseded by the consolidated [source maps, releases, and issues PR](https://github.com/hexclave/hexclave/pull/1973). The original PR and its review history are retained here.

PR #1868, opened 2026-08-04
feat(dashboard): issues UI — event graph, saved views, alerts and perf

mantrakp04:
Rounds out the observability app in the dashboard.

### Scope
- issue list with saved search views, bulk actions, and the occurrence graph
- issue detail: event sections, release context, attachment rendering
- the **alert rules pages** (rule form, delivery history)
- the performance page, and logs page updates

### Why the alert pages are here rather than with the alert backend
They import `issuesListHref` from the issues UI, so they have to sit at or above this layer even though their backend ships two PRs down.


mantrakp04:
Superseded by the replacement [observability UI PR](https://github.com/hexclave/hexclave/pull/1975). The original PR and its review history are retained here.

PR #1932, opened 2026-08-10
feat(deployments): config-as-code services on a Fly.io container runtime

BilalG1:
Rebuilds the (alpha) Deployments app end to end: services are declared in code, deployed per repository, and run as containers on **Marshal**, a new Fly.io-backed runtime in this repo.

This supersedes #1833 and #1852, which were stacked drafts of the same work — everything in them is here, based directly on `dev`.

---

## Deployment sources: a project, deployed from several repositories

A project owns **deployment sources**; a source is deployed one or more times; each deployment ships a set of services.

A source is one deploy file — `hexclave.deploy.ts`, named by its own `id` export — and it is what one `hexclave deploy` ships: **one upload, one build, one rollout**. That is what lets a frontend repo and a backend repo deploy into the same project without touching each other's services.

```ts
// hexclave.deploy.ts
import type { HexclaveDeploymentConfig } from "@hexclave/js";

export const id = "backend";

export const deployment: HexclaveDeploymentConfig = ({ isDev, secret, service, hexclave }) => ({
  services: {
    api: {
      type: "server",
      ports: { 8080: {} },
      dockerfilePath: "apps/api/Dockerfile",
      env: {
        DATABASE_HOST: isDev ? "localhost" : service("db").hostname(),
        DATABASE_PORT: "5432",
        OPENAI_API_KEY: isDev ? null : secret("OPENAI_API_KEY"),
      },
    },
    db: {
      type: "server",
      ports: { 5432: { protocol: "tcp" } },
      persistentVolumes: { pgdata: { path: "/data", sizeGb: 10 } },
      env: { POSTGRES_PASSWORD: secret("POSTGRES_PASSWORD") },
    },
  },
});
```

Service ids stay unique per **project**, not per source, so a reference never has to name a source: `service("api")` means one thing everywhere, whichever repository declares it. A second source claiming an id the first one owns is refused at sync, naming the owner — the alternative is two deploy files silently overwriting each other's definition on every deploy, with neither author able to see why.

`hexclave.config.ts` may still carry a `deployment` export for a project small enough to keep everything in one file; those services belong to a source named after that file.

A sync is the whole truth about its own deploy file, so **removing a service from `services` tears it down** on the next deploy — but keeps its persistent volume and any custom domain, unattached. A config typo must never destroy tenant data or release a certificate.

`hexclave deploy` no longer pushes the project config by default (`--config-push` opts in): each push replaces the whole configuration, and any of several repositories would otherwise clobber the others'.

## Ports

Ports are declared — and stored, and sent over the wire — as an object keyed by port number:

```ts
ports: { 3000: { public: true } }                  // one public HTTP port
ports: { 8080: {}, 5432: { protocol: "tcp" } }     // private, mixed protocols
ports: {}                                          // a worker that only dials out
```

One shape from the deploy file to the JSONB column, so nothing translates between two spellings of one thing — and a duplicate port becomes impossible by construction rather than by a rule. `public` defaults to false, `protocol` to `"http"`.

There is no service-level visibility: a service is public exactly when one of its ports says so. Each port answers on its **own number** on the private network; the service's single HTTP port additionally answers on 80/443 so its platform URL and any custom domain certificate work on the standard ports.

**A public port must be the service's only port.** This is a Fly.io limitation rather than a preference: Fly's `services` array is the proxy's listener set for the whole app with no per-address scoping, so every declared port answers on every IP the app holds. Once a public IP exists, a "private" sibling is on the internet — a public web port next to a private 5432 would publish the database. Private service traffic goes over Flycast, which *is* that proxy, so a private port's listener cannot simply be omitted; the escape hatch (`<app>.internal`, which bypasses the proxy) would cost autostart/autostop, load balancing and a stable address, and is not taken here. The rule is enforced in the config evaluator, the wire schema, the runtime and a CHECK constraint, and documented at each.

An **empty** port set is a worker. It has no URL of any kind and can hold no custom domain, and since only inbound traffic wakes a stopped machine it needs an always-on instance or it will never run. That is left to the author rather than refused.

## Connecting services

```ts
API_URL:  service("api").url(8080)     // that port's URL
DB_HOST:  service("db").hostname()     // private hostname, no port
DB_PORT:  "5432"                       // just write the number
```

`url(port?)` replaces the old `url` / `internalUrl()` pair. It resolves to the service's **public** URL when that port is public (its platform URL, or a verified custom domain) and to its **private** address otherwise, with the scheme from the port's protocol. A bare `url()` needs the target to declare exactly one HTTP port. `hostname()` is the private DNS name — a pure function of the service identity, so it answers before the target has ever deployed, and it points at the stable private IPv6 address of the app.

Which references make a deploy **wait** follows from that: a public `url` needs the target up, a private `url(5432)` does not. Getting this wrong in the safe-looking direction would serialize independent deploys and reject mutually-wired services as circular, so it is one shared function (`connectionRequiresTargetDeployed`) rather than a rule re-typed per layer.

## Service types and scaling

| | `server` | `serverless` |
|---|---|---|
| Instances | exactly one | `minInstances`–`maxInstances` |
| `minInstances` | 1 (default) stays up; 0 **suspends** — resumes with memory intact | 0 (default) scales to zero; every start is cold |
| Persistent volume | yes | no |

Both bounds cap at 10, and the CLI rejects a larger fleet by name and value before anything is packaged. `minInstances` above 0 is a paid capability for **both** types — a `server` at its default is as always-on as a pinned serverless — so a Free project must write `minInstances: 0` and the deploy fails up front naming the services that don't.

## One deploy, one build

`hexclave deploy` uploads the deploy file's directory **once** and the runtime builds every service from it in **one builder machine**, then rolls them out in dependency order. Previously each service was packaged, uploaded and built separately.

The build context is the whole upload, so a monorepo service can `COPY` shared code from above its own directory; `rootDirectory` only scopes where Railpack detection starts, and `dockerfilePath` is relative to the upload root. Each target gets its **own** build-env directory inside the machine, so a value belonging to one service is never offered to another's build.

A build failure fails the whole deployment and applies nothing. One machine builds them all in order, so there is no half-built source to salvage, and shipping the targets that happened to be first is not what the author asked for.

Progress is made on **read**: Marshal has no background worker and also runs on a serverless host, so `GET /deployments/:id` applies at most one more service per call under the source lease and reports stored state when another request already holds it. The backend polls while a deploy is in flight, which is what turns a series of cheap reads into a rollout.

## Deployments replace runs

`DeploymentRun` is gone. With one build per deploy there is nothing a run holds that a per-service outcome on the deployment does not, so the `Deployment` owns the build, its log, the redaction snapshot, and a `services` JSON of what each service did. The dashboard's deployment list is flat across sources with the **deployment source id** on every row — with several repositories deploying into one project, that is the first thing that tells two rows apart. Build logs live on the deployment and say so: they cover every service the deploy built.

## Automatic Hexclave credentials

Every deployed service receives `HEXCLAVE_PROJECT_ID`, `HEXCLAVE_API_URL`, `HEXCLAVE_PUBLISHABLE_CLIENT_KEY` and `HEXCLAVE_SECRET_SERVER_KEY`, plus `NEXT_PUBLIC_`/`VITE_` copies of the public three — a framework that inlines values at build time only reads its own prefix, so an unprefixed variable is invisible to the client bundle. The secret server key deliberately gets no prefixed copy: that name is a request to publish the value to the browser. An API key set is provisioned on demand if the project has none, and an explicitly declared env var of the same name always wins.

## Secrets

A per-project, write-only `ProjectSecret` table, KMS envelope-encrypted through the data vault's server-side flow. Values are set under **Project Settings → Secrets** or via `setProjectSecret`, resolved server-side at deploy time, and redacted from build logs using a per-deployment encrypted snapshot so historical logs stay redacted after a rotation.

`secret(key, default)` defaults never leave the deploy file: they travel with the deploy request and only fill secrets with no stored value. A key with a default is exempt from the preflight that fails a deploy on missing secrets — so a default is a value you are willing to ship. A deploy with an unfillable secret fails up front, naming every key that needs a value.

## Volumes and domains outlive their service

Both describe a resource that survives the service using them, so neither lives on the service row any more:

- A **volume** is owned by the deployment source and merely mounted by a service. The id names the disk within its source; one holder at a time, enforced by a unique index. Disks grow only — a shrink fails the sync rather than the apply.
- A **domain** is owned by the project and merely pointed at a service and one of its ports (validated to be the service's sole HTTP port). It has no foreign key to the service, which is the whole reason it is a table of its own.

Dropping a service leaves both behind, unattached, until someone deletes them explicitly.

## Marshal — the Fly.io runtime

`apps/marshal` is a stateless HTTP service (no database; its only state is an S3-compatible bucket). It takes deployments and service specs and converges Fly apps and machines toward them.

- **Builds** run on an ephemeral BuildKit machine and push to `registry.fly.io`, one machine per deployment. [Railpack](https://railpack.com) auto-detects by default; `dockerfilePath` selects a Dockerfile. Completion arrives over an HMAC-signed webhook — authenticated from the URL and headers *before* the body is parsed, with its own size cap — plus a lazy backstop for dead builders.
- **Networking.** Services are private by default and reach each other over a per-namespace 6PN network. `hostname` is a pure function of the service identity, so only a public `url` is a deploy dependency.
- **Domains** get certificates and DNS instructions, with hostname ownership enforced per tenancy. The backend and the runtime share one hostname pattern, so a hostname the API accepts cannot be one the runtime later refuses.
- **Safety.** Specs are AES-GCM encrypted at rest; Fly mutations are serialized behind a renewable S3-backed lease (per source for a deployment, per service for everything else); incomplete Fly REST responses are fenced rather than acted on; revisions are HMACs so a stored revision isn't an offline oracle for low-entropy secrets. Source archives are inflated off the event loop under a concurrency bound.
- **Mocks.** `MARSHAL_FLY_API_TOKEN=mock_hexclave_fly_key` points it at the `fly-mock` docker service and `MARSHAL_BUILDER=mock` skips real builds — both refused unless `MARSHAL_ALLOW_MOCKS=1`. That's what dev and e2e use.
- **Hosting.** Two entrypoints, like `apps/backend`: `src/server.ts` owns the listener, `src/vercel.ts` exports the same app binding nothing. `src/index.ts` re-exports the tsdown build so the deployed function runs the artifact `pnpm build` produces. It deliberately does not load `.env` files — on a hosted deployment the platform's environment is the only configuration.

## Platform capacity

Fly org limits (app count, IP allocations, API rate) are shared by every project on an instance, and hitting them fails in the middle of somebody's rollout. A deploy that would take the instance past `HEXCLAVE_MAX_DEPLOYED_SERVICES` (1000 by default, overridable because a self-hosted instance runs against its own org) is refused with a 503 saying it is a platform limit rather than the project's, and reported to Sentry so an operator hears about it first. Only services that do not yet hold a Fly app count, so re-deploying something already running keeps working at the ceiling. A non-numeric limit refuses deploys rather than silently disabling the guard.

## Data model

Three migrations. `DeploymentSource` (one per deploy file); `DeploymentService` gains its definition columns, its source, and a type; `DeploymentVolume` and `DeploymentDomain` become tables of their own; `Deployment` owns the build, the outcomes and the redaction snapshot; `DeploymentRun` and the Vercel columns are dropped. `ports` is a single JSONB object, and since it is the only record of whether a service is public, the database states the port rules itself — CHECK constraints over IMMUTABLE helpers, because a CHECK may not contain the subquery that walking JSON needs. Each rule was exercised against a real Postgres to confirm it produces its own *named* violation rather than a raw error.

## Testing

- The deployments e2e suite runs end to end against Marshal + `fly-mock` + `s3mock`, wired into all three e2e workflows. Migrations carry their own tests.
- Fly's behaviour was smoke-tested against real Fly before any of it was built; those throwaway scripts were removed from the branch before review.
- **Verified against real Fly and R2** (then torn down): volume naming and mount, `suspend`/`stop`/`off` autostop, detach keeping the disk, re-attach recovering it, grow in place, shrink refused, and a full upload → remote Dockerfile build → run cycle (87s, image digest matching what the machine runs).
- Unit tests, `pnpm typecheck` and `pnpm lint` are green.

## Rollout prerequisites

- Provision Marshal at a stable HTTPS URL with a dedicated production Fly org/token and S3/R2 bucket. Set `MARSHAL_API_KEY`, a separate `MARSHAL_WEBHOOK_SECRET`, a random 64-hex-character `HEXCLAVE_MARSHAL_DATA_ENCRYPTION_KEY`, `MARSHAL_PUBLIC_URL`, a production-unique `MARSHAL_ENV_ID` (Fly app names are globally unique, so dev/stg/prd must differ), and the `MARSHAL_FLY_*` / `MARSHAL_S3_*` values. Use `MARSHAL_BUILDER=fly` and do **not** set `MARSHAL_ALLOW_MOCKS`.
- Deploying Marshal on Vercel needs project settings that can't live in the repo, recorded in `apps/marshal/README.md`: root directory `apps/marshal` with source files outside it included (this is a pnpm workspace), build command `pnpm run build`, and **Deployment Protection off** (or a bypass) — both the backend and the Fly builder machine call in holding no Vercel credential. `MARSHAL_PUBLIC_URL` must be a stable custom domain, since builder machines call back minutes later.
- Point the backend at it with `HEXCLAVE_MARSHAL_URL` + `HEXCLAVE_MARSHAL_API_KEY`, then smoke-test health, a real build/deploy, and certificate provisioning.
- Consider `HEXCLAVE_MAX_DEPLOYED_SERVICES` against the Fly org's actual app limit before the alpha widens.
- Migrate any existing project to a `hexclave.deploy.ts` with an `id` export: ports as `{ <port>: { public?, protocol? } }`, `internalUrl()`/`internalHost` replaced by `url(port)`/`hostname()`, and `minInstances: 0` written out on any `server` on the Free plan.

## Known gaps

- Volume moves between services are identity-only; the data does not follow. Making it real would need one Fly app per deployment, at the cost of per-service certificates and Flycast hosts.
- A public service is restricted to one port (see **Ports**). Lifting it means moving private addressing off Flycast onto `<app>.internal`, which trades away autostart for private traffic.
- Two public ports, or a public raw-TCP port, are refused. Both would need a dedicated Fly IPv4 per service — a shared IPv4 routes only HTTP/TLS on 80/443 — so they are a billing decision as much as a code change.
- Every deploy rebuilds every service of its source: the build context is the whole upload, so any change invalidates every image. Per-target caching is a follow-up.
- Build-time env can skew from runtime env: a value edited without a code change rolls the machines while the already-built image keeps the old one baked in. The fix belongs in the product surface (report the service as stale, offer a redeploy), not in the runtime.
- A deployment advances only while something polls it. The CLI and the dashboard both do; a deploy nobody watches sits at its last applied service until the next read.
- The claim that Fly serves every declared port on every address the app holds is load-bearing for the one-port rule. It follows from the Machines API having no per-address scoping on `services`, but it has not been re-verified against real Fly since the restriction was added; worth confirming before the alpha widens.


apps/backend/src/lib/deployments/index.tsx
```diff
@@ -1,213 +1,459 @@
...
-// how you delete a live project by mistake.
-
-import { getBranchConfigOverrideSource, overrideBranchConfigOverride } from "@/lib/config";
+// KNOWN GAP — services removed from the config file. The definitions are
+// synced additively: a service that disappears from the `services` export
+// keeps its DeploymentService row (still visible in the dashboard) and its
+// live Marshal service. Deleting infrastructure automatically on a sync would
+// turn a config typo into a torn-down production deployment, so removal is
+// deliberately out of scope for now; an auto-cleanup layer (or an explicit
+// prune command) is planned to cover it.
```
N2D4: this is a problem. instead, in the schema we should have a Deployment row or something like that; it contains all the info on deployments, and a new deployment creates a new Deployment row.

N2D4 (approved):

N2D4 (approved):

N2D4 (approved):

N2D4 (approved):

PR #1942, opened 2026-08-12
Data Warehouse app (alpha)

BilalG1:
Adds `data-warehouse-alpha`: a per-project ClickHouse database plus a dedicated user with read/write access to it, on the same instance that serves analytics. Gated on the team plan or higher, provisioned explicitly from the dashboard.

## Screenshots

Dashboard, dark theme.

| Before provisioning | Free plan |
| --- | --- |
|  |  |
| A single Provision button, in the shape of the pre-Stripe payments empty state. | Without the entitlement the button is disabled and says why on hover, with an upgrade path in the banner. |

| Password, shown once | Provisioned |
| --- | --- |
|  |  |
| Provisioning and rotation both open this dialog. The password is never retrievable afterwards. | Connection details, with rotation in the page header. |

## The isolation problem this is built around

ClickHouse row policies filter *rows*; they do not restrict which *databases* a query may name. `/analytics/query` connects as a single shared `limited_user`, so granting that user access to one project's warehouse would have let every other project read it by naming it explicitly.

So the `default.*` grants and row policies move onto a new `analytics_reader` role, and each provisioned project gets its own ClickHouse user holding that role plus its own database. `limited_user` keeps the role and remains the default connection; the query path switches to the per-project user only when a warehouse exists — which is also what lets a project join analytics data against its own tables.

Those per-project users pin `SQL_project_id`/`SQL_branch_id` as `CONST` user settings instead of receiving them per query. That is stricter than the current model, where a caller that forgot to set them would read across tenants.

## What's in here

- **`apps/backend/src/lib/data-warehouse.tsx`** — provisioning, rotation, grants, per-user settings profile and quota, and explicit revokes of the table engines and sources that can reach outside the instance (`URL`, `S3`, `remote()`, `file()`, …). Every statement is idempotent, so a half-failed provision is repaired by retrying or rotating.
- **Credentials** — the password is shown once and stored KMS-encrypted, because the backend connects as that same user for analytics queries. It is decrypted per request; there is no cache to invalidate on rotation.
- **API** — admin-auth `GET /data-warehouse`, `POST /data-warehouse/provision`, `POST /data-warehouse/rotate-password`, plus matching admin-app SDK methods.
- **Dashboard** — Provision button, show-once credential panel, connection details, rotate action, `clickhouse-client` snippet, and a free-plan upsell.
- **Entitlement** — new `data_warehouse` item, seeded on team and growth.

## Testing

- 12 new e2e tests against real ClickHouse, including the one that matters: project B cannot read project A's warehouse, directly or through `/analytics/query`. Also covers free-plan rejection, the write path, rejection of attempts to override the pinned settings, denial of the exfil engines, and rotation invalidating the old password.
- Two `analytics-query` snapshots updated for the role change (`currentRoles()` is now `["analytics_reader"]`, and grants are held by the role); the other 68 tests in that file are unchanged and pass.
- Repo-wide `pnpm lint` and `pnpm typecheck` green.
- Dashboard flow driven manually against local built servers: provision → password panel → rotate, plus the upsell state with the entitlement removed.

## Deliberately deferred (all called out in code comments)

Storage caps (ClickHouse has no native per-database size limit, so nothing yet stops a project from filling the disk shared with our analytics), deprovisioning, plan-change reconciliation of the ClickHouse-side settings snapshot, self-hosted support, and per-branch databases.

## Deploy notes

- New Postgres migration `20260811000000_add_data_warehouse`.
- The ClickHouse migration re-points existing row policies from `limited_user` to the role with `CREATE ROW POLICY OR REPLACE`; it must run before a backend that expects the role.
- Direct customer connections need the instance reachable over TLS and `STACK_CLICKHOUSE_PUBLIC_HOST` set; without it the dashboard falls back to the host in `STACK_CLICKHOUSE_URL`.

---
## Summary by cubic
Adds the `data-warehouse-alpha` app so each project can provision a private ClickHouse database and user. `/analytics/query` now uses the project’s warehouse user when provisioned (falls back to the shared `limited_user` otherwise) to enforce isolation and enable joins with project tables; the warehouse path binds the same ClickHouse settings as the shared path to preserve spill/join behavior, omitting `max_memory_usage_for_user` because the profile enforces a MAX.

- Introduces an `analytics_reader` role carrying analytics row policies and SELECT grants. Migration order is hardened: create and grant the role first, move policies (temporarily also name `limited_user`), then revoke `limited_user`’s direct SELECT to avoid a cross‑tenant read window.
- Warehouse users pin `SQL_project_id`/`SQL_branch_id` as CONST and enforce per‑user quotas with MIN floors. Explicitly revoke outbound engines and table functions; grant table engines only when the admin can grant them. Config enables `access_control_improvements.table_engines_require_grant`.
- Admin API: `GET /data-warehouse`, `POST /data-warehouse/provision`, `POST /data-warehouse/rotate-password`. Overlapping provision/rotate return 409. Passwords are shown once, stored KMS‑encrypted; rotation invalidates the old one and does not terminate already‑established native sessions. Rotation is entitlement‑gated.
- Dashboard: provision card with plan upsell, show‑once password dialog, connection details, and a rotate action in the header. Entitlement gated by `data_warehouse` on Team/Growth.
- Security: drop the ClickHouse version gate; document residual DESCRIBE‑path SSRF via `mysql()`/`postgresql()` limited to blind port scanning. Local example ports are plaintext; set TLS ports in production. Direct‑credential users rely on network isolation.

**Rollout**
- Prisma: apply `20260811000000_add_data_warehouse` (new `DataWarehouse` table and `DataWarehouseStatus` enum).
- ClickHouse migration: create `analytics_reader`, grant it to `limited_user`, move row policies to the role while also naming `limited_user`, then `REVOKE ALL PRIVILEGES ON *.*` from `limited_user`; set `DEFAULT ROLE ALL` as needed before deploying the backend.
- ClickHouse config: enable `access_control_improvements.table_engines_require_grant` (mount `docker/dependencies/clickhouse-config.d/table-engine-grants.xml`); expose native port 9000 in local/CI.
- Connectivity: set `HEXCLAVE_CLICKHOUSE_PUBLIC_HOST`, `HEXCLAVE_CLICKHOUSE_PUBLIC_HTTPS_PORT`, and `HEXCLAVE_CLICKHOUSE_PUBLIC_NATIVE_PORT` to externally reachable endpoints; use TLS ports in production and ensure reachability.

<sup>Written for commit 7778f44af1b4074a82a1c7534642542066679512. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hexclave/hexclave/pull/1942?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added an alpha Data Warehouse for eligible plans, with isolated project databases and analytics access.
  * Administrators can provision warehouses, view connection details, retry failed setup, and rotate passwords.
  * Credentials are displayed securely once after provisioning or rotation.
  * Added dashboard status, error reporting, and connection details.
  * Analytics queries can use project-specific warehouse credentials.
* **Security**
  * Improved tenant isolation and restricted warehouse access to approved analytics data and operations.
  * Added resource limits and protections against unauthorized external data access.


apps/backend/src/lib/data-warehouse.tsx
```diff
@@ -0,0 +1,691 @@
...
+ * IMDS (which needs an HTTP GET) and cannot read anything back without valid
+ * target credentials. What remains is blind port scanning and blind outbound
+ * connects. An app-layer guard would not help — warehouse users hold direct
+ * ClickHouse credentials and bypass our routes. Self-hosted operators should
+ * firewall ClickHouse egress themselves.
+ */
+const FORBIDDEN_SOURCES = [
+  "FILE", "URL", "REMOTE", "MYSQL", "ODBC", "JDBC", "HDFS", "S3", "HIVE",
+  "MONGO", "REDIS", "SQLITE", "POSTGRES", "AZURE",
+] as const;
```
nams1570: Should we not also block `KAFKA, NATS, RABBITMQ, BIGQUERY, YTSAURUS`? Kafka was blocked as a table engine but not as a table source- any reason why? it's the same risk pattern no?

apps/backend/src/lib/data-warehouse.tsx
```diff
@@ -0,0 +1,691 @@
...
+
+/**
+ * The matching *source* privileges, gating the table functions (`url()`, `s3()`,
+ * `remote()`, `file()`, …). Same reasoning as the engines above.
+ *
+ * Known gap (ClickHouse #99122): these revokes gate execution but not schema
+ * resolution, so `DESCRIBE TABLE mysql(...)`/`postgresql(...)` (and `CREATE TABLE
+ * AS ...`) still dial out to an attacker-supplied host. Verified on Cloud
+ * 26.2.1.558; `url()`/`s3()`/`remote()` are denied even for DESCRIBE, and the same
+ * bypass exists via `/analytics/query` independently of this app.
```
nams1570: See dms for discussion on this

apps/backend/src/lib/data-warehouse.tsx
```diff
@@ -0,0 +1,691 @@
...
+ * Table engines that can reach outside this ClickHouse instance. Recent versions
+ * revoke these from non-admin users by default, but the warehouse user is the only
+ * account that can create tables at all, so we revoke them explicitly rather than
+ * trust a server default. Revoking a privilege that was never granted is a no-op.
+ */
+const FORBIDDEN_TABLE_ENGINES = [
+  "AzureBlobStorage", "Distributed", "File", "HDFS", "Hive", "JDBC", "Kafka",
+  "MongoDB", "MySQL", "NATS", "ODBC", "PostgreSQL", "RabbitMQ", "Redis", "S3",
+  "SQLite", "URL", "IcebergS3", "DeltaLake",
+] as const;
```
nams1570: I think there are a bunch of engines that should be restricted here but arent

apps/backend/src/lib/data-warehouse.tsx
```diff
@@ -0,0 +1,691 @@
...
+/**
+ * Local-only engines a customer warehouse still needs. With
+ * `table_engines_require_grant` enabled, engines must be granted explicitly, so
+ * this allow-list is the counterpart to the deny-list above.
+ */
+const ALLOWED_TABLE_ENGINES = [
+  "MergeTree", "ReplacingMergeTree", "SummingMergeTree", "AggregatingMergeTree",
+  "CollapsingMergeTree", "VersionedCollapsingMergeTree", "GraphiteMergeTree",
+  "Log", "TinyLog", "StripeLog", "Memory", "Set", "Join", "Buffer", "Null",
+] as const;
```
nams1570: Why no SharedMergeTree and ReplicatedMergeTree?

nams1570: Wouldn't the mergetree subclasses be safe? they're all in process and tbh if ppl know they're using a clickhouse cluster they prolly want some control over that

apps/backend/src/lib/data-warehouse.tsx
nams1570: Although broadly, forbid lists for both sources and engines can easily go stale. A stricter allow list is prolly easier to maintain

apps/backend/src/lib/data-warehouse.tsx
```diff
@@ -0,0 +1,691 @@
...
+      for (const engine of ALLOWED_TABLE_ENGINES) {
+        await client.command({ query: `GRANT TABLE ENGINE ON ${engine} TO ${quotedUser}` });
+      }
+    }
+    for (const engine of FORBIDDEN_TABLE_ENGINES) {
+      await client.command({ query: `REVOKE TABLE ENGINE ON ${engine} FROM ${quotedUser}` });
+    }
+    for (const source of FORBIDDEN_SOURCES) {
+      await client.command({ query: `REVOKE ${source} ON *.* FROM ${quotedUser}` });
+    }
```
nams1570: Why not have REVOKE SOURCES ON *.8 or REVOKE TABLE ENGINES ON...?

apps/backend/src/lib/data-warehouse.tsx
```diff
@@ -0,0 +1,691 @@
...
+      `,
+    });
+
+    // Password last: if any grant or quota command above fails, the old password
+    // still works, so a rotation can repair the rest instead of locking everyone out.
+    await client.command({
+      query: `
+        ALTER USER ${quotedUser}
+        IDENTIFIED WITH sha256_password BY {password:String}
+        DEFAULT ROLE ALL
```
nams1570: Why default role all instead of default role analytics_reader? rn everything uses that one role right? isnt it better to scope it

apps/backend/src/lib/data-warehouse.tsx
```diff
@@ -0,0 +1,691 @@
...
+
+/**
+ * Creates (or repairs) the ClickHouse database, user, grants, settings and quota
+ * for a project, and sets the user's password.
+ *
+ * Every statement is idempotent, so this doubles as the retry path for a
+ * provisioning run that died halfway.
+ */
+async function applyWarehouseDdl(options: {
+  tenancy: Tenancy,
```
nams1570: Wouldn't people be able to run `SELECT name FROM system.databases;￼SELECT database, name, engine FROM system.tables;￼SELECT database, table, name, type FROM system.columns;`? That's system wide metadata info no? is that an issue?

apps/backend/src/lib/data-warehouse.tsx
```diff
@@ -0,0 +1,691 @@
...
+    // still works, so a rotation can repair the rest instead of locking everyone out.
+    await client.command({
+      query: `
+        ALTER USER ${quotedUser}
+        IDENTIFIED WITH sha256_password BY {password:String}
+        DEFAULT ROLE ALL
+        SETTINGS
+          SQL_project_id = '${tenancy.project.id}' CONST,
+          SQL_branch_id = '${tenancy.branchId}' CONST,
+          max_execution_time = ${timeoutSeconds} MIN ${MIN_EXECUTION_TIME_SECONDS} MAX ${MAX_EXECUTION_TIME_SECONDS},
```
nams1570: Wait doesn't this mean that people can bypass their quota limit and get the growth plan timeout?

nams1570: They could connect via the dashboard and run something like `SELECT count() FROM default.events￼CROSS JOIN numbers(100000000)￼SETTINGS max_execution_time = 300` no?

apps/backend/src/lib/data-warehouse.tsx
```diff
@@ -0,0 +1,691 @@
...
+  const rows = await result.json<{ can_grant: string }>();
+  const canGrant = Number(rows.data[0]?.can_grant ?? 0) > 0;
+  adminCanGrantTableEnginesCache = canGrant;
+  return canGrant;
+}
+
+async function runClickhouseCleanupSteps(
+  context: string,
+  steps: ReadonlyArray<() => Promise<unknown>>,
+  reportError: (context: string, error: unknown) => void = (errorContext, error) => captureError(errorContext, error),
```
nams1570: Nit: the weird dependency injection thing is odd here for a func that's called in only one place. I appreciate that you're doing this to allow you to unit test it but this is where an e2e test is more appropriate imo

apps/backend/src/lib/data-warehouse.tsx
```diff
@@ -0,0 +1,691 @@
...
+  const quotaName = `\`${userName}_quota\``;
+  const client = getClickhouseAdminClient();
+  // Keep the database and customer data so a retry stays non-destructive, but drop
+  // credentials the failed request could neither return nor store. Every step runs
+  // even if an earlier one fails, so a quota error cannot strand live credentials.
+  return await runClickhouseCleanupSteps("data-warehouse-clean-up-unpersisted-user", [
+    async () => await client.command({ query: `DROP QUOTA IF EXISTS ${quotaName}` }),
+    async () => await client.command({ query: `DROP USER IF EXISTS ${quotedUser}` }),
+    async () => await client.close(),
+  ]);
```
nams1570: How would keeping the db and customer data be non destructive if youre getting rid of the user? I guess you're worried about getting rid of user data? but if you're in the cleanup phase, wouldn't you want to wipe the table too?

nams1570: Ah, unless you feel like we should perenially keep the data just in case

apps/backend/src/lib/data-warehouse.tsx
nams1570: Just realized that AI queries won't be able to access data warehouse as it stands rn. Probably a good idea to expose that no?

apps/backend/src/lib/data-warehouse.tsx
```diff
@@ -0,0 +1,691 @@
...
+    throw new StatusError(500, "Failed to provision the data warehouse. Please try again.");
+  }
+  return { password: result.password, warehouse: result.warehouse };
+}
+
+/**
+ * Issues a new password for an existing warehouse and returns it once. Re-applies
+ * the full DDL, so rotating also repairs drifted grants or a half-failed provision.
+ */
+export async function rotateDataWarehousePassword(tenancy: Tenancy): Promise<{ password: string, warehouse: DataWarehouse }> {
```
nams1570: Here we don't check to see if a row is ready before we rotate no? so it's theoretically possible to call rotate password on a failed row which would 500. Should probably treat that as a 400

nams1570 (changes requested):

PR #1991, opened 2026-08-19
Let a Deploy service run a prebuilt image instead of building one

BilalG1:
## What

A Deploy service can now name an already-built public image instead of building one from your source:

```ts
database: {
  type: "server",
  ports: { 5432: { protocol: "tcp" } },
  image: "postgres:16",
  persistentVolumes: { pgdata: { path: "/data", sizeGb: 10 } },
  env: { POSTGRES_PASSWORD: secret("POSTGRES_PASSWORD") },
},
```

This replaces the shim our own docs currently ask for — a `./database` directory whose entire content is a Dockerfile saying `FROM postgres:16` — and with it the builder machine that rebuilt and re-pushed that image on every deploy. A prebuilt deploy uploads nothing, starts no builder, and finishes in seconds.

## Design decisions

**`image` is mutually exclusive with `rootDirectory`/`dockerfilePath`.** Those say where the code is; this says what to run. The exclusivity is a union in the authoring types, so a service declaring both does not compile, and it is re-checked in the schema, the CLI and the runtime.

**Marshal does not resolve images. The reference reaches the machine config as written, and Fly resolves it when it pulls.** An earlier revision of this PR resolved a tag to a digest at deploy time, which meant a registry client — the one place Marshal fetched a host the *user* chose, and therefore ~500 lines of SSRF hardening (pinned DNS, private-range and NAT64 blocking, per-hop redirect re-validation, bounded bodies). Fly already does this resolution and reports what it resolved to, so that client is gone.

Tags and digests are both accepted, and which one you write is the whole of the difference:

- A **digest** names fixed bytes by definition. Nothing can move under you.
- A **tag** is a pointer, and Fly is what dereferences it. Three consequences follow, and they are the contract:
  - Two machines of one service, or a machine recreated later for a volume change or a crash, can run different bytes under one revision if the publisher moves the tag in between.
  - A redeploy of an *unchanged* tag rolls nothing: the machine config is identical, so the config hash matches and nothing is pulled. Moving forward onto a republished tag means changing the reference.
  - A reference that does not exist is Fly's error at apply time, not a 400 on the deploy request — and on a mixed deployment that lands after the build has already run.

A reference with neither tag nor digest is still refused rather than defaulted to `:latest`. That rule matters *more* without resolution: an unpinned pointer in the machine config is the one reference guaranteed to move under a running service.

**The digest Fly resolved is read back and recorded**, so deployment history still says which bytes ran even when the spec named a tag. Two details, both confirmed against real Fly:

- It is composed onto the author's canonical reference, not onto Fly's own `image_ref.registry`/`.repository` — those describe how Fly *fetched* the image rather than what was asked for, and a Docker Hub image comes back as `docker-hub-mirror.fly.io/library/redis`. Recording that would put Fly's internal mirror in the user's deployment history.
- For a multi-platform image Fly reports the **linux/amd64 manifest**, not the index. So a service pinned to an index digest records a different — more precise — one: the exact bytes the machine ran.

**`upload_id` became optional end to end** — required exactly when a target is built from source, refused otherwise. An all-prebuilt deployment opens directly in `deploying` and reports no build logs. Mixed deployments work: the builder gets only the source targets, and the build's images merge into the pre-seeded ones.

**Private registries are not supported yet**, but the builder decision is expressed as "targets needing a builder" rather than "targets without an image", so mirroring an image through the existing builder stays an additive change. Fly's Machines API has no pull-secret field, so mirroring is what that will have to be.

## Two things this exposed

- `has_build_logs` was `marshalBuildId !== null` — i.e. "the runtime accepted this deployment" rather than "a builder ran" — so a buildless deploy would have offered a log that streams nothing. It now comes from the runtime's own `has_logs`.
- Image normalization lived only in the CLI, so `hexclave deploy` stored `docker.io/library/postgres:16` while a direct API write stored `postgres:16`. It moved into the schema, so every write path agrees.

## One accepted gap

An all-prebuilt deployment has **no replay fence**. A source deploy is fenced by its upload, consumed exactly once; a prebuilt one has no upload, so two identical requests each create a deployment. The applied state still converges — the runtime serializes a source's applies and re-applying one spec is a no-op — so the cost is a duplicate row in deployment history, not a broken deploy. Fencing it needs a client-minted idempotency key threaded through the CLI, the wire, a schema column and a unique index; that was built and then deliberately reverted as too much machinery for an untidy list. The route says so where the upload is consumed.

## Testing

**Verified against real Fly and real Docker Hub**, deployed with the CLI against a local backend — a mixed deploy of `redis:7-alpine` (tag), `redis@sha256:ff02b5…` (digest) and a source-built service:

- The machine config held the tag **verbatim** (`docker.io/library/redis:7-alpine`). Fly pulled it, and redis 7.4.11 was accepting connections **4.45s** later — no builder for either prebuilt service.
- Both prebuilt services recorded `sha256:1db42cc…`, which is exactly the `linux/amd64` manifest inside the index Docker Hub returns for that tag. The tag and the digest spellings converged on the same bytes, which is the resolution being correct.
- The cross-kind connection resolved: curling the built service's public URL returned the prebuilt service's `.flycast` hostname.
- **A redeploy of the unchanged tag left both prebuilt machines untouched** — same `instance_id`, same `updated_at` — while the rebuilt source service rolled. That is the second bullet of the contract above, observed rather than argued.
- A nonexistent tag failed at apply with Fly's `MANIFEST_UNKNOWN` naming the tag, so the moved failure mode is still legible to a user.

All Fly apps created by that smoke were destroyed afterwards.

- Deployments e2e: **48 passed**, 1 skipped, including all-prebuilt, mixed, and upload-refusal cases. The prebuilt case asserts both halves of the split: the machine is given `docker.io/library/postgres:16`, and the outcome records `docker.io/library/postgres@sha256:…`.
- Unit: Marshal 123, CLI 243, shared 23, backend-deployments 21.
- The fly-mock grew an `image_ref` that resolves a reference to a deterministic digest, which is what the readback is exercised against in CI.

## Migrations

Two, both additive and backwards-compatible: `DeploymentService.image` (nullable; NULL = built from source) and `Deployment.hasBuildLogs` (defaults true, which is what every pre-existing row is).

---
## Summary by cubic
Run Deploy services from a prebuilt public image via a new image field instead of always building from source. Prebuilt deploys upload nothing, start no builder, and still record the digest Fly resolved so history names the bytes that ran.

- Enforce image XOR rootDirectory/dockerfilePath across schema, CLI, backend, and runtime; `upload_id` is required only when any service builds from source. Mixed deploys build only source targets; all‑prebuilt deploys open in deploying with `hasBuildLogs=false`.
- Image references: require an explicit tag or digest and canonicalize to a fully‑qualified string (Docker Hub aliases and `library/`, case‑insensitive hosts, ports 1–65535). Validate shape only; Fly resolves at pull. Consequences: machines under one revision can differ if a tag moves; redeploying an unchanged tag rolls nothing; nonexistent refs fail at apply. Record and show the digest Fly resolved (including when nothing was built).
- CLI upload: use node:http with the presigned slot’s deadline; detect stalls and partial writes; succeed only after the full body is sent; use monotonic timers; clear errors; avoid false “slot expired” on clock skew; add a close backstop to avoid hanging sockets.
- CLI/Dashboard links: always print a shareable deployment link; when a service fails, deep‑link directly to that service’s build logs. Build URLs robustly; the dashboard consumes and strips the params once, selects the linked service/panel once, and hides Build logs when `hasBuildLogs=false`.
- Source manifest: persist a listing of packaged files (paths and sizes only), largest‑first, deduped, and bounded by count and path length. Backend parses leniently; the LIST endpoint omits manifests; the single‑deployment GET returns them; the Source tab renders per‑service slices and totals; show the resolved digest in the “nothing built” state.
- Guards and correctness: reject reserved service ids like “__proto__”; validate empty/malformed image digests; normalize older outcomes to include `image=null`; the mock runtime reports deterministic resolved digests.

**Rollout**
- Add `DeploymentService.image` (nullable; NULL = built from source), `Deployment.hasBuildLogs` (NOT NULL DEFAULT true), and `Deployment.sourceManifest` (JSONB, nullable; no backfill).

<sup>Written for commit 51e72ee43b83b7d9808a6ed821f64aadc4f02296. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hexclave/hexclave/pull/1991?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Deploy services using prebuilt container images, including tagged or digest-pinned references.
  - Support mixed image-based and source-built deployments without unnecessary uploads.
  - View the image applied to each service, including resolved digests.
  - Open deployments and services directly from dashboard links.
  - CLI deployments now provide dashboard links and include deployment URLs in JSON output.

- **Bug Fixes**
  - Improved validation and clearer errors for invalid or conflicting image configurations.
  - Build logs now accurately indicate when no build occurred.
  - Source uploads provide stronger deadline, timeout, and error handling.

- **Documentation**
  - Added guidance on image tag resolution and digest pinning.


apps/marshal/src/registry.ts
```diff
@@ -0,0 +1,492 @@
...
+    })).map((entry) => entry.address);
+  if (addresses.length === 0) throw badRequest(`the registry host ${JSON.stringify(bare)} could not be resolved`);
+  for (const address of addresses) {
+    if (!isPublicAddress(address)) {
+      throw badRequest(`the registry host ${JSON.stringify(bare)} resolves to a non-public address`);
+    }
+  }
+  return addresses;
+}
+
```
N2D4: we already have a lot of this logic in other places where we check for SSRF, can we dedupe?

apps/marshal/src/registry.ts
```diff
@@ -0,0 +1,492 @@
...
+/**
+ * One request to a user-influenced host, with the guards that implies.
+ *
+ * Redirects are followed BY HAND rather than by the HTTP client, because each
+ * hop is a new destination that has to pass the private-network check: a
+ * registry that redirects to "http://169.254.169.254/" must not be followed just
+ * because the first hop was fine.
+ */
+async function registryGet(url: string, headers: Record<string, string>, deadline: number): Promise<RegistryResponse> {
```
N2D4: can't we let fly do the downloading instead of downloading and then reuploading to fly?

N2D4 (changes requested):
can we make fly do the downloading of the image instead of downloading and then reuploading? that makes things much easier for us

apps/marshal/src/registry.ts
BilalG1: fixed, we pass image value directly to fly now

N2D4 (approved):

PR #1999, opened 2026-08-21
feat(tv-mode): add project-scoped dashboards and secure display pairing

AddarshKS:
> [!NOTE]
> This PR supersedes #1945. The original PR remains the historical record for earlier automated-review discussions and resolutions.

## Summary

This PR introduces TV Mode: a production-intended, full-screen operational display experience powered by live, project-scoped Hexclave data.

TV Mode gives teams a shared view of audience activity, collected revenue and payment health, and email delivery health. It adds configurable presentation profiles, secure independent displays, prioritized event interruptions, bounded Incident and Critical Incident takeovers, recovery presentations, celebrations, and backend-enforced financial privacy.

Normal playback is entirely live-data driven. Synthetic fixtures are restricted to explicitly gated developer previews and automated tests.

**Base:** `dev` → **Head:** `TV_Mode`  
**Scope:** Backend snapshot assembly, evaluators, persistence, dashboard UI, migrations, independent-display authorization, and regression coverage. Raspberry Pi appliance productization is delivered separately in stacked PR #2024.

## Visual Tour

Captured from the current local development environment using the seeded Demo Project. TV Mode is a greenfield surface relative to `dev`, so these are after-only captures rather than artificial before/after comparisons.

### Presentation Profiles

| Light | Dark |
| --- | --- |
| profiles-overview-after-light | profiles-overview-after-dark |

### Profile Configuration

The editor keeps the playlist and rotation controls prominent, uses one single-column flow at every viewport, and consolidates Celebration, Incident, and Critical Incident timing behind a shared category selector. General Mode remains visible; synthetic Event and State Previews use the existing project-scoped development-feature gate.

| Light | Dark |
| --- | --- |
| profile-config-after-light | profile-config-after-dark |

Configuration flow:

profile-config-scroll-dark

### Independent Displays

| Light | Dark |
| --- | --- |
| displays-after-light | displays-after-dark |

### Live Presentation

Captured from the centralized, development-gated TV fixture so every screen shows representative, deterministic activity—including qualified email outcomes and non-zero live activity. The in-product preview label makes the synthetic source explicit.

| Live Pulse | Audience Momentum |
| --- | --- |
| live-pulse-fixture-wide | audience-momentum-fixture-wide |

| Revenue & Payments | Email Health |
| --- | --- |
| revenue-payments-fixture-wide | email-health-fixture-wide |

### Profiles

Projects can create and manage named presentation profiles with configurable:

- Screen playlists and rotation durations
- Profile descriptions and display names
- Event interruption policy
- Incident, Critical Incident, recovery, celebration, animation, and highlight timing
- Financial visibility
- Reduced-motion behavior

Built-in profiles are immutable virtual templates. Customizing one creates a saved, tenancy-scoped profile rather than modifying the template.

### Independent Displays

The Displays section supports persistent TV devices through the public `/tv` entry point without granting dashboard access.

Administrators can:

- Open or copy the independent TV link
- Pair a display using a short-lived secure code
- Name the display
- Assign or reassign a profile
- Review connection and last-seen state
- Explicitly authorize exact financial visibility
- Unpair a display immediately

Display credentials use a narrow display-specific principal. Pairing includes expiry, replay protection, distributed rate limiting, credential rotation, project isolation, and server-authoritative assignment/privacy checks.

Unpairing now performs one atomic, tenancy-scoped hard deletion of the exact display. PostgreSQL cascades only that display’s credential rows, while the independent security event remains as the audit trail. Active refresh credentials use a rolling 30-day idle lifetime; used-token replay evidence is retained for 24 hours and successful rotations clean at most 100 old rows from the exact display/family.

### Live Pulse

Provides a concise operational overview of current activity and monitored-source readiness. Empty, insufficient, limited, unavailable, failed, and stale/offline states remain distinct; source readiness is never presented as business health.

### Audience Momentum

Shows project-scoped audience data including users, growth, new users, monthly activity, lifecycle, verification, signed-in visitors, and identified-session duration.

Core audience data survives optional Analytics disablement or failure. Missing enrichment is represented explicitly rather than replaced with stale values, false zeros, or fixture data.

### Revenue & Payments

Presents the broadest trustworthy collected-revenue view currently supported by Hexclave:

- Successfully collected subscription invoices
- Successful `PURCHASE_PAGE` one-time purchases
- Subscription collection success
- Active subscriptions and related payment context

Gross collected revenue excludes test-mode purchases, API grants, pending or failed collections, duplicated provider representations, and refunds. It does not approximate net revenue or perform FX conversion.

Payment integration changes add nullable outcome fields, indexes, population and ordering safeguards, migration support, and data mirroring. Existing Stripe-provided invoice-status behavior is preserved; TV Mode consumes the additional outcome facts through its own evaluator rather than redefining Payments invoice status.

### Email Health

Uses actual project email records for finished sends, confirmed deliveries, delivery rate, failure/error/bounce outcomes, comparison windows, freshness, and source availability. No hardcoded delivery percentage or demo fallback exists in normal playback.

## Snapshot and Presentation Architecture

All four screens consume one normalized backend-built `TvSnapshot`. The backend is authoritative for:

- Project and tenancy scope
- Source querying and aggregation
- Freshness and diagnostic state
- Partial-source failure handling
- Financial redaction and PII exclusion
- Evaluator evidence and event lifecycle state
- Presentation selection

The authenticated dashboard route and independent `/tv` route share the same presentation controller and renderer.

Polling replaces a snapshot atomically. It does not restart rotation, replay an assigned phase, extend deadlines, or retain stale source-specific enrichment inside a fresh snapshot. Whole-snapshot failure continues with the last safe snapshot and explicit stale/offline treatment.

Profile reads use the primary database; profile deletion and display operations preserve atomic state changes. Failed pairing approvals refund consumed rate-limit tokens. Evaluator state persists atomically, and stale payment candidates reset rather than retaining old evidence.

### Email Evaluator V2

Uses project-relative delivery behavior, minimum samples, bounded persistence, threshold-jitter handling, conservative fallback, and validated recovery. It distinguishes qualifying degradation, borderline evidence that pauses persistence, and healthy evidence that resets a candidate or confirms recovery without introducing another user-facing lifecycle state.

### Subscription Collection Evaluator

Payment incident detection is intentionally narrower than the broader Revenue & Payments screen. It evaluates authoritative terminal subscription-invoice outcomes using persisted outcome timestamps, currency, and collected amount.

Each invoice is classified by its latest authoritative terminal outcome and windowed by outcome time rather than creation time. Pending, open, draft, voided, and unknown invoices remain outside the success/failure denominator. One-time purchases, refunds, abandoned checkouts, and raw payment intents do not enter this evaluator.

### Incidents, Recovery, and Celebrations

TV Mode supports bounded Incident and Critical Incident takeovers, active Highlights, escalation, frozen recovery phases, resolved Highlights, and milestone celebrations. Evaluators own occurrence truth; the orchestrator owns profile-specific presentation priority.

Assigned takeover, recovery, Highlight, and animation deadlines are frozen. Polling, refreshing, or editing a profile cannot replay or extend an active phase. Incident-to-Critical escalation receives a fresh bounded Critical phase, and celebration suspension/resumption respects the original deadline.

Revenue Milestones remain visibly unavailable because no live evaluator exists.

## Presentation Priority

1. Critical Incident takeover
2. Incident takeover
3. Recovery takeover
4. Celebration takeover
5. Active or resolved Highlights
6. Normal profile rotation

## Security and Privacy

- Every live query is project- and tenancy-scoped.
- Normal playback never falls back to Demo Project or fixtures.
- Cross-project profile, snapshot, pairing, and display access is rejected.
- Built-in profiles remain immutable; saved profiles remain tenancy-owned.
- Exact financial visibility requires explicit profile configuration and display acknowledgement.
- Financial redaction and PII exclusion are enforced by the backend.
- Independent display credentials cannot access ordinary dashboard APIs.
- Display assignment, privacy authorization, credential version, and revocation remain server-authoritative on each poll.
- Unpair deletion is constrained by exact display ID and authenticated tenancy ID.
- Presentation mode remains read-only and isolated from ordinary dashboard navigation.

## Validation

For PR head `27990f877`, GitHub reports successful lint/build, Prisma migration/schema synchronization, fast-fail tests, API E2E tests, fallback E2E tests, and Docker build/run checks.

Earlier development validation also covered shared/backend/dashboard/E2E typechecks, focused TV unit/component/backend tests, disposable migration tests, Prisma schema validation, migration-ledger verification, and manual dashboard/independent-display/presentation QA. These are historical validation results, not a claim that every check was rerun at the current head.

The separate [migration-compatibility workflow](https://github.com/hexclave/hexclave/actions/runs/34665639754) is not green and is not included in the passing results above. Its merge disposition remains with the maintainers.

## Deployment Notes

- Requires PostgreSQL 15 or newer.
- Run the snapshot range-index migration as a dedicated, single-runner deployment step before enabling production snapshot polling. Follow the checked-in [deployment guide](https://github.com/hexclave/hexclave/blob/27990f8775517651186ad9e5bce324ed81be0ccb/apps/backend/prisma/migrations/20260725000000_add_tv_snapshot_range_indexes/DEPLOYMENT.md) for timeout sizing, index verification, retry and rollback.
- Migration handling includes transient-failure retry and interrupted-index verification; payment backfill detects the available outcome columns. These safeguards do not replace the documented deployment preflight.
- Apply the required TV Mode and payment-outcome migrations before serving the new application code. Do not treat operation against the pre-TV schema as a supported rollout sequence.
- Configure the independent display origin and the documented TV snapshot CORS/header contract for the deployment topology.
- Independent displays use narrowly scoped HTTP-only refresh cookies, with secure and same-site behavior selected for the configured deployment.

## Future Roadmap and Deferred Scope

These are follow-up products, not implicit scope in this PR. They should reuse the snapshot, evaluator, orchestration, presentation, profile, and narrow display-auth boundaries established here.

### Near-Term Product Roadmap

- **TV Box / Raspberry Pi productization:** Stacked PR #2024 implements the lightweight `/tv-box` renderer and pilot appliance, including automatic startup, Wi-Fi provisioning, persistent pairing, supervised recovery, diagnostics, reset workflows, and versioned image provisioning. Further GA filesystem/release hardening, broader hardware/security qualification, and signed remote-update/rollback support are deferred to separate future PRs. Those appliance follow-ups are not part of this TV Mode PR.
- **Deployment Support:** Add trustworthy project-scoped deployment status after defining authoritative signals, privacy rules, freshness semantics, and interruption policy.
- **Daily Briefing:** Add scheduled project briefings after defining time-zone behavior, qualification/freshness, missed-briefing semantics, and priority relative to incidents and celebrations.
- **Revenue Milestones:** Enable the existing unavailable presentation only after a live project-scoped evaluator defines authoritative evidence, deterministic milestones, and replay behavior.
- **Project-Level Event and Evaluator Platform:** Generalize trustworthy event truth for TV Mode, Daily Briefing, and future consumers while keeping consumer-specific presentation separate.
- **Custom Dashboard Playlist Entries:** Consider bounded dashboard views only after defining their snapshot/privacy contract and independent-display rendering constraints; do not couple arbitrary dashboard navigation or administrator authentication into TV playback.

### Longer-Term Financial Scope

- One-time-purchase payment incident evaluation
- Refund and net-revenue analytics
- Currency conversion
- Historical payment-outcome backfill
- Additional event types that reuse the evaluator/orchestrator/presentation architecture

## Architectural Invariants

Future changes should preserve:

- One backend-built `TvSnapshot`
- No fixture fallback in normal playback
- Strict project and tenancy isolation
- Evaluators owning event truth
- Orchestration owning presentation priority
- Bounded takeovers and frozen deadlines
- No deadline replay/extension through polling
- Server-enforced privacy
- Narrow independent-display authorization
- Shared rendering between authenticated and independent TV playback


apps/backend/prisma/schema.prisma
```diff
@@ -2238,6 +2476,13 @@ model SubscriptionInvoice {
   status                        String?
   amountTotal                   Int?
   hostedInvoiceUrl              String?
+  paidAt                        DateTime?
+  // Event time of the newest exact invoice outcome applied; NULL means no exact outcome has been applied.
+  paymentOutcomeEventAt         DateTime?
+  markedUncollectibleAt         DateTime?
+  voidedAt                      DateTime?
+  currency                      String?
+  amountPaid                    Int?
```
BilalG1, line 2482: look at /Users/bgodil/source/stack-auth/apps/backend/src/app/api/latest/internal/metrics/route.tsx, function loadPaymentsOverview. We don't need to change payment logic, we should reuse our existing payments analytics instead

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/displays/page-client.tsx
```diff
@@ -0,0 +1,147 @@
+"use client";
+
+import type { TvProfileResource } from "@hexclave/shared/dist/interface/admin-tv-mode";
```
Comment (BilalG1) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/displays/page-client.tsx:3]: for dashboard changes add few screenshots to pr body so it's easier to review

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/profiles/[profileId]/page-client.tsx
```diff
@@ -0,0 +1,908 @@
...
+      <div className="shrink-0">{control}</div>
+    </div>
+  );
+}
+
+export default function PageClient() {
+  const projectId = useProjectId();
+  const adminApp = useAdminApp();
+  const project = adminApp.useProject();
+  const profileId = getProfileIdFromPath(usePathname());
```
Comment (BilalG1) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/profiles/[profileId]/page-client.tsx:272]: this page is way too confusing, need to simplify the ui a lot. I think most important thing is playlist configuration

Also it would be interesting if we could reuse custom dashboards in playlists because that way people can show stuff specific to their own company

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/profiles/[profileId]/page-client.tsx
```diff
@@ -0,0 +1,908 @@
...
+                  control: <Switch checked={false} disabled />,
+                })}
+              </div>
+            </DesignCard>
+
+            <DesignCard title="Privacy" subtitle="Control what a shared room can see" icon={ShieldCheckIcon} gradient="green" glassmorphic>
+              {settingRow({
+                title: "Show Exact Financial Values",
+                description: "Shows exact currency in Revenue & Payments. Operational counts and direction remain visible when off.",
+                control: <Switch checked={draft.showExactFinancialValues} onCheckedChange={(showExactFinancialValues) => setDraft({
```
Comment (BilalG1) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/profiles/[profileId]/page-client.tsx:786]: I don't think we need options like "Show exact financial values". If someone doesn't want this data shown then they should just disable the payments screen in playlist

apps/backend/prisma/schema.prisma
```diff
@@ -864,6 +869,234 @@ model Tenancy {
...
+  parentId        String? @db.Uuid
+  issuedAt        DateTime @default(now())
+  expiresAt       DateTime
+  usedAt          DateTime?
+  revokedAt       DateTime?
+  replacementId  String? @db.Uuid
+
+  @@index([displayId, familyId, revokedAt], name: "TvDisplayCredential_family_idx")
+  @@index([expiresAt], name: "TvDisplayCredential_expiresAt_idx")
+}
```
BilalG1, line 976: we shouldn't create a separate auth flow for tv mode, reuse cli auth routes for now and later we'll extend to use our device auth routes

apps/backend/prisma/schema.prisma
```diff
@@ -864,6 +869,234 @@ model Tenancy {
...
+
+model TvEventOccurrence {
+  id        String @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  tenancyId String @db.Uuid
+  tenancy   Tenancy @relation(fields: [tenancyId], references: [id], onDelete: Cascade)
+
+  eventType         TvEventType
+  presentationClass TvEventPresentationClass
+  lifecycle         TvEventOccurrenceLifecycle
+  deduplicationKey  String @db.VarChar(200)
```
BilalG1, line 1022: A lot of these tables and logic are specific to tv mode but we're going to need the same stuff for users who aren't using tv mode. For example, email delivery degraded is probably an event that we track for every project and tv mode profiles are just one consumer of the event. Think about how we can make it more general

BilalG1 (changes requested):

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/profiles/[profileId]/page-client.tsx
Comment (AddarshKS) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/profiles/[profileId]/page-client.tsx:272]: Agreed that Playlist should be the primary surface, and I'll simplify the profile editor accordingly. Custom dashboards require a separate typed snapshot, rendering, and security contract beyond the four backend-built TV screens, so we’re deferring that additive feature to a focused follow-up rather than expanding this PR.

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/profiles/[profileId]/page-client.tsx
Comment (AddarshKS) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/profiles/[profileId]/page-client.tsx:786]: I suggest we retain this option because disabling Revenue & Payments hides the entire operational screen, including collection health and revenue direction. This option allows backend-enforced financial redaction, that lets shared TVs retain those useful aggregates while withholding exact currency amounts, a useful option for tenants providing a per-profile privacy boundary.

apps/backend/prisma/schema.prisma
AddarshKS: The CLI flow issues a project-user credential with broader project authority. Independent displays require a narrow principle that can only refresh/unpair itself and receive the assigned profile’s TV snapshot, while preserving immediate revocation, reassignment, and server-enforced financial policy. I therefore designed this display-specific pairing credentials rather than granting the TV CLI-level project access. I highly suggest retaining this.

apps/backend/prisma/schema.prisma
AddarshKS: Agreed. Email, payment, analytics, and milestone evaluation should ultimately produce project-level event truth consumed by TV Mode and other products coming in the future (Daily Briefing, for example). Doing that correctly requires scheduling independent of TV polling and generic lifecycle/evidence persistence, so I suggest recording it as an important follow-up for TV Mode rather than applying a superficial table rename or expanding this PR. I would like to know your views about this.

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/displays/page-client.tsx
Comment (AddarshKS) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/displays/page-client.tsx:3]: Done! The latest commit includes updated description with the addition of screenshots and gifs using the pr-visual-writeup skill from the repository.

apps/backend/prisma/schema.prisma
AddarshKS: The current payments fields cannot represent the authoritative outcome timestamps, collected amounts, and currencies required by TV Mode’s revenue reporting and subscription-collection evaluator. I therefore added normalized observational fields and additive Stripe ingestion writes; I haven't changed the existing checkout, collection, refund, subscription, or entitlement behavior. TV Mode consumes these additional facts without replacing or altering the existing payment analytics pathway.

BilalG1: for currency we only support USD right now so isn't necessary. The other fields added I see how they make tv mode revenue reporting more accurate but they also expand scope of this pr and make it less marspox

apps/backend/prisma/schema.prisma
BilalG1: yeah I think that is fine for keeping TV mode separate for now

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/profiles/[profileId]/page-client.tsx
Comment (BilalG1) [reply apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/tv-mode/profiles/[profileId]/page-client.tsx:786]: I agree it could be useful but it is very special-cased. There are all types of data that a team might not want to show and long-term solution is probably closer to custom dashboards than list of toggles

apps/backend/prisma/schema.prisma
BilalG1: I'd still recommend reusing our cli auth and eventually device auth flow once it is in. We usually don't want to reimplement security sensitive stuff like this

AddarshKS: Agreed, I did some back-tracking on the current CLI Auth hexclave offers. I agree that we should reuse established primitives and avoid maintaining parallel authentication implementations unnecessarily.

The distinction here is the identity being authorized: the current CLI flow hands over a project-user refresh token, whereas an independent TV should never become a logged-in project user.

The display needs a narrowly scoped identity bound to its approved tenancy and assigned profile, with only snapshot and self-service credential operations. Administrator unpairing must invalidate that display independently of the approving user’s session, and profile assignment and financial visibility must remain server-enforced.

I have made sure the implementation already reuses our existing JWT and secure-random utilities. TV Mode should retain this display-specific authorization model for this PR. When the generic device-auth flow supports purpose-bound, least-privilege principals, we can reuse that flow without granting TVs broader user access or weakening their existing lifecycle guarantees.

apps/backend/prisma/schema.prisma
AddarshKS: I re-audited this against the existing payment logic and removed the TV-introduced invoice-status override. 
Invoice status is now stored directly from Stripe again, preserving the existing Payments behavior.
I retained the additive payment fields because TV needs collection timestamps, terminal outcomes, actual collected amounts, and currency - information the previous fields did not fully represent. Their population, ordering safeguards, nullable compatibility, indexes, and mirroring remain intact.

TV’s Payments V2 evaluator consumes those facts independently; it does not overwrite the shared invoice status. This preserves TV’s reporting functionality without introducing a new invoice-status policy into Payments. Regression coverage now explicitly checks that separation.

BilalG1 (approved):

PR #2000, opened 2026-08-21
Data Sources: sync a customer's Postgres into their warehouse

BilalG1:
Builds on #1942. Where that PR gives a project its own ClickHouse database, this one fills it: a customer connects their own PostgreSQL, picks tables, and each one syncs into that warehouse in whichever mode suits it.

Stacked on `feat/data-warehouse` — review that first, or read this diff alone.

## Two sync modes, chosen per table

The mode is not a question we put to the customer cold. A capability probe reads what the server can do and what shape each table is in, and the answer falls out of that — overridable per table.

| Mode | What it does | Sees deletes |
| --- | --- | --- |
| **CDC** | Reads the write-ahead log each interval | Yes |
| **Cursor** | Re-reads rows past a watermark on a column they pick | No |

A full-refresh mode was built and then removed before review. Its limit was on the wrong axis — a row count, when the cost is bytes — and measurement showed how far that misses: 100k rows is 11 MB of scalars or 600 MB of documents, and we were recommending a reload of either every five minutes. Sizing it properly needs table-byte probing and a frequency-aware gate, which is more machinery than the mode earns while these two cover the common cases. The snapshot reader it shared with CDC is still here, so re-adding it later is a swap-in.

**Choosing a cursor column is explained, not assumed.** Any `NOT NULL` timestamp or integer qualifies, but they behave differently: a timestamp the application bumps on every write picks up edits, while an `id` or `created_at` only moves on insert, so edits to existing rows are never synced. That is correct for an append-only table and quietly wrong everywhere else — so the picker says what to choose and flags a non-temporal choice with a warning icon rather than refusing a call only the customer can make.

**CDC without a long-lived process.** Replication slots are durable server-side state, so there is no need to hold a streaming connection open. Every sync connects over ordinary SQL, drains the slot with `pg_logical_slot_peek_binary_changes`, and advances it with `pg_replication_slot_advance` — which drops straight into the existing cron runtime instead of needing a new stateful component.

Peek rather than `get_changes` is the load-bearing detail: the destructive read advances the slot as it returns, so a failure between reading and writing to ClickHouse would lose those changes permanently. We advance only after the destination write succeeds, which makes the whole path at-least-once — and at-least-once is exactly what `ReplacingMergeTree` already tolerates.

That means decoding `pgoutput` ourselves (~240 lines, `apps/backend/src/lib/data-sources/pgoutput.ts`). Chosen over `wal2json` because it is built into every Postgres 10+, including managed servers that will not install an extension for us.

## Shared decision logic

`packages/shared/src/data-sources/modes.ts` decides availability and the recommendation, and **both the backend and the table picker import it**. A mode the dashboard offers but the backend rejects is the worst possible outcome of that screen. Choices are re-validated against a fresh probe on save, so a table that lost its primary key since the picker rendered is refused rather than silently downgraded.

Unavailable reasons ride on the dropdown option itself, so the UI needs no tooltip layer:

```
CDC — needs wal_level=logical    (disabled)
Cursor · updated_at
Full refresh — table too large   (disabled)
```

## Destination shape

`ReplacingMergeTree(_hexclave_version, _hexclave_deleted)` keyed on the source primary key, or append-only `MergeTree` without one.

Versions are the commit LSN for CDC and the cursor value for polling. CDC snapshots are written at **version 0** so any WAL change for the same row wins — which is what makes it safe to create the slot *before* snapshotting. The overlap deduplicates rather than leaving a gap, and a gap would be silent data loss.

## Testing

- **20 API-level e2e tests** over all six endpoints, against a real Postgres source and the real ClickHouse warehouse. Assertions read the destination as the *customer's own warehouse user*, so isolation is exercised, not just the happy path.
- Each e2e test gets a throwaway **database**, not a schema: logical slots are per-database, and a slot leaked by a failing test would pin WAL on the app's own database for everything running alongside.
- 39 unit and integration tests: the `pgoutput` decoder against hand-built byte fixtures, the mode rules, version arithmetic, plus a Postgres integration suite gated behind `HEXCLAVE_DATA_SOURCE_TEST_POSTGRES` (skips in CI; the docker command is in the file header).
- Repo-wide `pnpm typecheck` and `pnpm lint` green.
- Driven end to end by hand against a live stack: connect → probe → choose tables → sync → CDC insert/update/delete → mode switch → disconnect.

## Review found real bugs — and so did running it

Four review agents plus an independent `codex` pass. ~25 findings fixed in `4b6be995c`; the ones that mattered most:

- **Silent truncation.** A 50-batch cap applied to CDC snapshots too, so a table over 500k rows had its LSN cursor written after a partial load and the remainder was never read again.
- **Versions were not comparable within a table.** Cursor versions are epoch microseconds; LSNs are many orders of magnitude smaller. Switching modes froze the destination forever. Streams now rebuild rather than merge.
- **Isolation.** Syncs used the ClickHouse *admin* client, bypassing the per-project user #1942 built as the tenancy boundary. Now the boundary is a ClickHouse privilege and quota, not the correctness of an interpolated database name.
- **Cursor candidates were matched against type names `format_type()` never emits**, so every Prisma-backed source reported "no usable column" and could not sync incrementally at all.
- **`REPLICA IDENTITY NOTHING` tables are refused for CDC** — adding one to a publication makes the customer's own `UPDATE`s start failing.

Three bugs came only from *running* it, not from reading:

- The first sync failed outright: warehouse databases are named by project UUID and the identifier quoter rejected hyphens. The destination path had never executed against a production-shaped name — the tests hardcoded `wh_test`.
- CDC streams silently dropped columns added after setup. The WAL carries no DDL, and the destination schema was only reconciled during the snapshot.
- The cursor watermark went through `Date.toISOString()` — millisecond precision — so a `timestamptz` with microseconds produced a watermark *below* the rows just read, and every sync re-read the whole table forever. Caught by an e2e assertion reaching 51 rows instead of 26.

## Deploy notes

- New migration `20260821000000_add_data_sources` (`DataSource`, `DataSourceStream`, four enums).
- New cron in `apps/backend/vercel.json`: `/api/latest/internal/data-source-sync-step`, every minute. It takes a leased claim, so overlapping invocations cannot sync one source twice.
- Requires a provisioned Data Warehouse and the `data_warehouse` entitlement (team plan and above), checked on every outbound path including the scheduler.

## Known gaps

- **No full-refresh mode**, so a small mutable table on a server without logical replication has no mode that sees edits or deletes. This is the main thing to weigh before this ships broadly.
- **Integer cursors have no safety lag.** Sequence gaps can skip rows; the honest fix is CDC, not a larger lag. The 10-second temporal lag is a mitigation, not a guarantee.
- **`bytea` and array values encode differently** on the snapshot path (base64 / JSON) versus the CDC path (Postgres text literals).
- **Publication creation needs table ownership.** If the role lacks it, the sync fails with the exact `CREATE PUBLICATION` statement for a DBA to run — the industry norm, but real friction on the CDC path.
- **No deletes-reconciliation pass** for polling modes. Cursor mode still cannot see deletes; the UI says so, but the primary-key-only tombstone scan is not built.
- `PAUSED` / `SNAPSHOTTING` and `syncIntervalSeconds` are modelled but unwired.

### Configure PostgreSQL streams

Dark-mode table picker

### Synced source status

Dark-mode source details

### Newly discovered tables require explicit opt-in

Existing streams remain selected while a newly discovered table starts unchecked.

Newly discovered table remains unchecked

### DIFF
--- a/apps/backend/src/app/api/latest/internal/data-source-sync-step/route.tsx
+++ b/apps/backend/src/app/api/latest/internal/data-source-sync-step/route.tsx
@@ -1,0 +1,55 @@
+import { runDueDataSourceSyncs } from "@/lib/data-sources";
+import { createSmartRouteHandler } from "@/route-handlers/smart-route-handler";
+import { yupBoolean, yupNumber, yupObject, yupString, yupTuple } from "@hexclave/shared/dist/schema-fields";
+import { getEnvVariable } from "@hexclave/shared/dist/utils/env";
+import { StatusError } from "@hexclave/shared/dist/utils/errors";
+
+export const dynamic = "force-dynamic";
+export const fetchCache = "force-no-store";
+// Snapshots of large tables are the slow case here, so this takes the full
+// function budget rather than the default.
+export const maxDuration = 300;
+
+const FUNCTION_BUDGET_MS = maxDuration * 1000;
+const FUNCTION_SHUTDOWN_SLACK_MS = 20 * 1000;
+const HARD_DEADLINE_MS = FUNCTION_BUDGET_MS - FUNCTION_SHUTDOWN_SLACK_MS;
+
+export const GET = createSmartRouteHandler({
+  metadata: {
+    summary: "Sync due data sources",
+    description: "Internal endpoint invoked by Vercel Cron. Syncs every data source whose interval has elapsed, until the queue drains or the function's time budget runs out.",
+    tags: ["Data Warehouse"],
+    hidden: true,
+  },
+  request: yupObject({
+    auth: yupObject({}).nullable().optional(),
+    method: yupString().oneOf(["GET"]).defined(),
+    headers: yupObject({
+      "authorization": yupTuple([yupString()]).defined(),
+    }).defined(),
+    query: yupObject({
+      max_duration_ms: yupString().optional(),
+    }).defined(),
+  }),
+  response: yupObject({
+    statusCode: yupNumber().oneOf([200]).defined(),
+    bodyType: yupString().oneOf(["json"]).defined(),
+    body: yupObject({ ok: yupBoolean().defined() }).defined(),
+  }),
+  handler: async ({ headers, query }) => {
+    if (headers.authorization[0] !== `Bearer ${getEnvVariable("CRON_SECRET")}`) {
+      throw new StatusError(401, "Unauthorized");
+    }
+    const requested = query.max_duration_ms != null ? Number(query.max_duration_ms) : HARD_DEADLINE_MS;
+    if (!Number.isFinite(requested) || requested < 0) {
+      throw new StatusError(400, "Invalid max_duration_ms");
+    }
+    const deadlineMs = Date.now() + Math.min(requested, HARD_DEADLINE_MS);
+
+    while (Date.now() < deadlineMs) {
+      const { didWork } = await runDueDataSourceSyncs({ deadlineMs });
+      if (!didWork) break;
+    }
+    return { statusCode: 200, bodyType: "json", body: { ok: true } };
+  },
+});
--- a/apps/backend/src/lib/data-sources.tsx
+++ b/apps/backend/src/lib/data-sources.tsx
@@ -1,9 +1,9 @@
-import { getClickhouseAdminClient } from "@/lib/clickhouse";
-import { ensureDataWarehouseEntitlement, getDataWarehouse, getDataWarehouseNames } from "@/lib/data-warehouse";
+import { createClickhouseWarehouseClient } from "@/lib/clickhouse";
+import { ensureDataWarehouseEntitlement, getDataWarehouse, getDataWarehouseNames, getDataWarehouseQueryAuth } from "@/lib/data-warehouse";
 import { getDestinationTableName } from "@/lib/data-sources/clickhouse-destination";
-import { DATA_SOURCE_SSL_MODES, type DataSourceCredentials } from "@/lib/data-sources/postgres";
+import { DATA_SOURCE_SSL_MODES, quotePgIdentifier, withDataSourceClient, type DataSourceCredentials } from "@/lib/data-sources/postgres";
 import { probeDataSource, type DataSourceProbeResult, type ProbedTable } from "@/lib/data-sources/probe";
-import { runStreamSyncs, type StreamSyncPlan } from "@/lib/data-sources/sync";
+import { runStreamSyncs, type StreamSyncPlan, type SyncCursorState } from "@/lib/data-sources/sync";
 import { getTenancy, type Tenancy } from "@/lib/tenancies";
 import { getPrismaClientForTenancy, globalPrismaClient, retryTransaction } from "@/prisma-client";
 import { Prisma, type DataSource, type DataSourceStream } from "@/generated/prisma/client";
@@ -20,6 +20,9 @@ const encryptedPasswordSchema = yupObject({
   edkBase64: yupString().defined(),
   ciphertextBase64: yupString().defined(),
 }).defined();
+
+/** How long a claimed sync may run before the scheduler assumes it died. */
+const SYNC_CLAIM_LEASE_SECONDS = 900;
 
 const MODE_TO_PRISMA = {
   full_refresh: "FULL_REFRESH",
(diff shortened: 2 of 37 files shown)


nams1570 (commented):
A suggestion for the future: it would be very cool to expose a WAL stream for our users to build cdc mechanisms using as well! I imagine we'll be extending this to more data sources in the future, and so the work of actually synchronizing the changes into one format will have to be done by us anyway, and it allows people to build reactive flows based on one stream that we provide to them.  
  
I'm a little nervous about the cursor method. I appreciate that it's just a fallback for the sync, but it doesn't track deletes and we don't have a reconciliation mechanism meaning we'd have stale data no?

packages/shared/src/data-sources/modes.ts
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * The cursor column we preselect. Indexed candidates first — an unindexed cursor
+ * makes every sync a sequential scan — then the conventional names, so the common
+ * case needs no thought from the customer.
+ */
+export function getDefaultCursorColumn(table: DataSourceTableInfo): string | null {
+  const preferredNames = ["updated_at", "updatedat", "modified_at", "last_modified", "created_at", "createdat"];
+  const ranked = [...table.cursorCandidates].sort((a, b) => {
+    if (a.indexed !== b.indexed) return a.indexed ? -1 : 1;
```
nams1570: Aren't primary key tables usually indexed? so wouldn't they end up being preferred over the timestamp tables?

apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/data-warehouse/sources/stream-picker.tsx
```diff
@@ -0,0 +1,284 @@
...
+  const selection: Record<string, Selection | undefined> = {};
+  for (const table of catalog.tables) {
+    const key = `${table.schema_name}.${table.table_name}`;
+    const prior = existingByKey.get(key);
+    selection[key] = prior != null
+      ? { on: true, mode: prior.mode, cursorColumn: prior.cursor_column ?? table.default_cursor_column }
+      // Recommendations are defaults only during initial setup. While editing,
+      // newly discovered tables must require an explicit opt-in.
+      : { on: existing == null && table.recommended_mode != null, mode: table.recommended_mode, cursorColumn: table.default_cursor_column };
+  }
```
Comment (nams1570) [inline apps/dashboard/src/app/(main)/(protected)/projects/[projectId]/data-warehouse/sources/stream-picker.tsx]: Let's say you start a sync and then you delete the primary key column in the postgres table. Wouldn't the checkbox be greyed out here? and this would block dropping the stream no?

apps/backend/src/lib/data-sources/clickhouse-destination.ts
```diff
@@ -0,0 +1,167 @@
...
+    : "MergeTree";
+  const orderBy = primaryKeyColumns.length > 0
+    ? primaryKeyColumns.map(quoteClickhouseIdentifier).join(", ")
+    : quoteClickhouseIdentifier(EXTRACTED_AT_COLUMN);
+
+  return `CREATE TABLE IF NOT EXISTS ${quoteClickhouseIdentifier(databaseName)}.${quoteClickhouseIdentifier(tableName)} (
+  ${definitions.join(",\n  ")}
+) ENGINE = ${engine}
+ORDER BY (${orderBy})`;
+}
```
nams1570: Rebuild only looks at mode and cursor column. If the source primary key changes and the customer saves the same mode, we write the new primaryKeyColumns onto the stream but leave status ACTIVE and keep the existing ClickHouse table.

ensureDestinationTable uses CREATE TABLE IF NOT EXISTS, so ORDER BY stays the old key. ReplacingMergeTree then dedups on a key that no longer matches Postgres.

We should treat a PK change like a mode change: clear the cursor, mark PENDING, and drop/recreate the destination before the next sync.

nams1570: Idea here is that the clickhouse row identity wont match the postgres identity

apps/backend/src/app/api/latest/internal/data-source-sync-step/route.tsx
```diff
@@ -0,0 +1,55 @@
...
+    if (headers.authorization[0] !== `Bearer ${getEnvVariable("CRON_SECRET")}`) {
+      throw new StatusError(401, "Unauthorized");
+    }
+    const requested = query.max_duration_ms != null ? Number(query.max_duration_ms) : HARD_DEADLINE_MS;
+    if (!Number.isFinite(requested) || requested < 0) {
+      throw new StatusError(400, "Invalid max_duration_ms");
+    }
+    const deadlineMs = Date.now() + Math.min(requested, HARD_DEADLINE_MS);
+
+    while (Date.now() < deadlineMs) {
```
nams1570: Why Date.now() instead of performance.now() dont the other crons use performance?

apps/backend/src/lib/data-sources.tsx
```diff
@@ -0,0 +1,544 @@
...
+import { yupObject, yupString, yupValidate } from "@hexclave/shared/dist/schema-fields";
+import { StatusError, captureError } from "@hexclave/shared/dist/utils/errors";
+
+const encryptedPasswordSchema = yupObject({
+  edkBase64: yupString().defined(),
+  ciphertextBase64: yupString().defined(),
+}).defined();
+
+/** How long a claimed sync may run before the scheduler assumes it died. */
+const SYNC_CLAIM_LEASE_SECONDS = 900;
```
nams1570: I could just be misunderstanding this, but dont vercel crons last for like 300s? and each cron can trigger a sync? isnt this a mismatch?

nams1570: Some jobs will take longer, but then ext cron run would try to pick it up no? isnt that an issue?

nams1570: Maybe we should quick ack vercel but also have the system acknowledge in the back that the sync could still be going on

apps/backend/src/lib/data-sources.tsx
```diff
@@ -0,0 +1,544 @@
...
+
+  try {
+    if (removesLastCdcStream) {
+      // Unlike source deletion, a configuration update must fail loudly here.
+      // Otherwise it can succeed while leaving an unconsumed, WAL-retaining slot.
+      await dropCdcInfrastructure(source);
+    }
+
+    const keep = new Set(configs.map(c => `${c.schemaName}.${c.tableName}`));
+    await retryTransaction(prisma, async tx => {
```
nams1570: Could we accomplish this with CTEs instead? I believe retryTxn is deprecated. I appreciate why you want this to be atomic though so if you feel like it's necessary that's fine

apps/backend/src/lib/data-sources.tsx
```diff
@@ -0,0 +1,544 @@
...
+
+  // Drop the replication slot before forgetting it exists: a slot nobody reads
+  // retains write-ahead log on the customer's server until their disk fills.
+  //
+  // Attempted unconditionally rather than only when replicationSlotName is set,
+  // because that column is written after a sync completes — a sync that created
+  // the slot and then timed out leaves one behind with no record of it. The name
+  // is derived from the source id, so it is always recoverable.
+  try {
+    await dropCdcInfrastructure(source);
```
nams1570: why just flag here whereas in the setDataSources thing if it fails we drop it? Here if this fails, isnt it better to just fail the datasource delete too?

apps/backend/src/lib/data-sources.tsx
```diff
@@ -0,0 +1,544 @@
...
+  await withDataSourceClient(credentials, async client => {
+    await client.query(
+      `SELECT pg_drop_replication_slot($1) WHERE EXISTS (SELECT 1 FROM pg_replication_slots WHERE slot_name = $1)`,
+      [slotName],
+    );
+    await client.query(`DROP PUBLICATION IF EXISTS ${quotePgIdentifier(slotName)}`);
+  }, { allowWrites: true });
+}
+
+export async function deleteDataSource(tenancy: Tenancy, dataSourceId: string): Promise<void> {
```
nams1570: Is it possible for one cron instance to claim the data source and start working/ syncing from it while anither one tries to delete it?

nams1570 (changes requested):
I'm broadly concerned about what happens when a stream is dropped or a postgres column that is important (LoAd BeArIng lol) is dropped (say, primary key or any other cursor table), especially when there's something else going on concurrently with that stream... Also txns and scaling worry me

